Issue #13, opened 2018-09-18
Tracking implementation of SHDR chunk


colinator27:
This chunk is pretty complex compared to the last one I documented, and many others... I'm not sure how much of this will make sense, but I guess this issue can track the implementation of the SHDR chunk to this library/tool.

Here's what I got, from the latest GMS2 compiler. I haven't tested too much because it gets complicated very fast.
```
SHDR: ListChunk
This chunk seems pretty strange and complex in some places, and I tried my best to document it (somewhat) clearly.
---
<normal FourCC and chunk length>

List<T> format beginning:
 Int32 - count
 <list of 32-bit pointers to each item>
 
 For each item:
  32-bit string pointer to shader name
  
  Int32 - shader type with a bitwise OR against -0x80000000 (in big endian here, also it's signed and negative), as a number it is -2147483648
   Basically, in this case, this just changes the last byte of the little-endian Int32 to 0x80. For what reason is unknown?
   Little-endian-formatted types that are known, total of seven:
    0x01000080 (1) -> GLSL ES
    0x01000080 (2) -> GLSL
    0x03000080 (3) -> Unknown
    0x04000080 (4) -> HLSL
    0x05000080 to 0x07000080 (5 to 7) -> Unknown
   Missing from this list (the "unknown" types) are PSSL (PlayStation Shader Language), Cg, and Cg_PS3 (for PS3 seemingly)
    Also potentially a difference between HLSL9 and HLSL11
    I was unable to test these
  
  <note: many of the strings and pointers in this chunk will be null>
  32-bit string pointer to GLSL ES vertex shader code
  32-bit string pointer to GLSL ES fragment shader code
  32-bit string pointer to GLSL vertex shader code
  32-bit string pointer to GLSL fragment shader code
  32-bit string pointer to HLSL9 vertex shader code
  32-bit string pointer to HLSL9 fragment shader code
  
  <note: I'm going to give some pointers ID numbers so it's easy to tell where they go in the chunk>
  32-bit pointer to HLSL11 vertex shader string/bytes (ID #1)
  32-bit pointer to HLSL11 pixel shader string/bytes (ID #2)
  
  Int32 - count of vertex shader attributes
  Array (NOT a List<T>) of 32-bit string pointers, using the count specified on the last line, each representing a vertex shader attribute
  
  Int32 - literally just the number 2
  
  32-bit pointer to PSSL vertex shader string/bytes (ID #3)
[...]

PR #25, opened 2018-09-28
Added internal support for SHDR chunk

colinator27:
So, using the information I obtained in #13, I managed to get a working implementation of the SHDR chunk's serialization/unserialization. I've tested it with an example project, and it's able to produce an identical file to the input if you export.

The code still needs implementations with the editor, if that's planned.

By the way, if you're curious about some of the strange parts of the code, I was forced to make workarounds to problems that resulted from some limitations of the program's code (for instance it can be hard to figure out where you are in a list at a given time when unserializing, and where the next object is meant to be, etc.), but it seems to work pretty well now. Originally I was going to make my own List class, but then it got really crazy and I discovered a much simpler approach.

### DIFF
--- a/UndertaleModLib/Models/UndertaleShader.cs
+++ b/UndertaleModLib/Models/UndertaleShader.cs
@@ -1,0 +1,363 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace UndertaleModLib.Models
+{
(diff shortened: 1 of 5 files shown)


krzys-h:
Yeah, things that don't have a length specified anywhere are a problem. Not using the pointers for figuring out data layout at all was actually my original design choice as I was afraid of losing some data after you read a file and write it back - but looking at objects like this, maybe it was not the best idea.

I will merge this soon, I just want to double-check it first

PR #30, opened 2018-10-08
Various additions and improvements (read comments)

colinator27:
I just did a bunch of work on the tool and library.

Here's a copy of what I said on my commit with the changes:
```
Implemented various structures, added some editors, fixed some bugs, cleaned up

- Implemented Timeline data structure internally, added semi-partial editor
- Added partial editor for Shaders
- Added semi-partial editor for Extensions
- Implemented physics shape vertices on GameObjects (could cause errors without them)
- Renamed variables in gameobject event actions to their proper names/meanings (as well as the Awake variable in GameObjects)
- Made string references in the editor more clear as to when they are null or empty
```

There's only one issue with it, and it is that the vertex shader attributes list editor (in a new shader editor) is currently a bit buggy, and I'm not quite sure how to fix it. Let me know what you think.


UndertaleModTool/Editors/UndertaleCodeEditor.xaml.cs
```diff
@@ -372,6 +372,12 @@ private async void DecompileCode(UndertaleCode code)
                                                 possibleObjects.Add(data.Fonts[id]);
                                             if (id < data.Sounds.Count)
                                                 possibleObjects.Add(data.Sounds[id]);
+                                            if (id < data.Shaders.Count)
+                                                possibleObjects.Add(data.Shaders[id]);
+                                            if (id < data.Extensions.Count)
```
krzys-h: Can extensions actually be referenced in scripts by ID? If not, then this is just cluttering up the context menu list.

UndertaleModTool/Editors/UndertaleShaderEditor.xaml
```diff
@@ -0,0 +1,104 @@
...
+        <local:UndertaleStringReference Grid.Row="6" Grid.Column="1" Margin="3" Height="40" ObjectReference="{Binding HLSL9_Vertex}"/>
+
+        <TextBlock Grid.Row="7" Grid.Column="0" Margin="3">HLSL9 fragment source</TextBlock>
+        <local:UndertaleStringReference Grid.Row="7" Grid.Column="1" Margin="3" Height="40" ObjectReference="{Binding HLSL9_Fragment}"/>
+
+        <Border Grid.Row="8" Grid.Column="0" Margin="20" Background="LightPink" BorderBrush="Red" BorderThickness="3">
+            <TextBlock TextWrapping="Wrap" Margin="3" Foreground="Red">NOTE: The vertex shader attributes editor is currently buggy and may not work as expected.</TextBlock>
+        </Border>
+        
+        <!-- TODO: Fix this editor -->
```
krzys-h: I guess I'll know what 'fix' means when I try to run this but this comment could probably be a little bit more specific

krzys-h: Now I see that you wrote what it means in the pull request itself ;)

UndertaleModTool/Editors/UndertaleCodeEditor.xaml.cs
colinator27: I kind of recall that being possible somewhere, but now I'm not so sure. I put it in just to be safe, but if it's unnecessary, I guess it's okay to get rid of it.

colinator27:
Thanks for the feedback, some of these things I didn't even have the time to fully test. I'll try to fix some of these when I get the chance.

To add to that list, though (I really should have explained everything I noticed, I don't know why I didn't now):

5. If you edit a string in the list and select to "only change this one value", for some reason it fails to update the string object in the list. It works when you select "change all occurrences" for obvious reasons, as I'm assuming the string reference remains the same.

By the way, by that global initialization script editor, do you want it to be linked to by the editor somewhere, like in the tree?

krzys-h:
I think it should work similarly to the "general info" tab, because the chunk does not contain any objects itself and just points to entries in the CODE block

krzys-h:
The reason the editor fails to update objects is probably because you need to manually specify `﻿Mode=TwoWay, UpdateSourceTrigger=PropertyChanged` (or `LostFocus` when using a plain TextBox and not UndertaleString/ObjectReference) for bindings inside CellTemplate. This is because DataGrid by default has a read-only CellTemplate and read-write EditableCellTemplate activated by double-clicking, but I decided not to do that to make editing simpler. Look at how other editors (e.g. the paths or sprite textures) do it.

colinator27:
I apologize for the long time it's been taking to work on these changes. I've been pretty busy and I just got started on it now.

colinator27:
I forgot to include this in the commit message, but I also fixed the vertex shader attribute list not being initialized and removed extensions from the "possible objects" list.

colinator27:
Sorry if these messages are messed up right now, I've been trying to delete the excess ones but GitHub is still broken right now...

krzys-h:
Thanks!

I tried to get that shader attributes editor to work and I have no idea how to do it. In other places it works fine when the UndertaleString is wrapped inside another object. I know this is ugly, but how about something like that as a temporary workaround:
```csharp
class VertexShaderAttribute : UndertaleObject
{
    public UndertaleString Name;
}
List<VertexShaderAttribute> VertexShaderAttributes;
```

For the general info editor, the easiest way would be probably to literally replicate how the general info editor works, just with one List<String> instead of the other objects (look for GeneralInfoEditor in MainWindow.xaml.cs)

Is there anything else unfinished except these two things? (and the string editor background, but that's a minor thing that can be fixed later) I would like to merge this soon, there are quite a few improvements in this pull request and it would be nice to release a new version with them.

colinator27:
The UndertaleObject implementation works like a charm... I managed to also take advantage of the existing UndertaleIO code to make it more consistent, like `UndertaleSimpleList`.

You can read the description of the last commit for everything else I just changed. That should be about it, except I still haven't gotten the string editor background to work, of course. I tried working on it before but to no avail...

colinator27:
One minor problem still remains: when moving scaled objects around, their center origins do not update until you stop moving them.

colinator27:
Alright, so that's fixed. I accidentally included padding calculations in the loop that detects the last chunk, but I just found out that the chunk length includes the padding in it.

colinator27:
So now instead of getting an exception message when loading bytecode version 14, now it just loads the file but ignores the sections with code. It also disables exporting because at that point it's useless and would likely fail anyway.

krzys-h:
I kinda need your CanSave changes right now so... time to merge this... *pleasedon'tbreak*

Next time, could you please avoid useless merge commits at the beginning (if you have remotes set up correctly just do `git fetch && git reset --hard upstram/master && git push -f` before you start) and split your changes into smaller pull requests so that they are easier to review?

PR #62, opened 2018-11-13
Name some more variables, and some editors for them, especially GMS2

colinator27:
One note: SWF sprite parsing is still broken. Something's wrong with the parsing code I wrote, but I have no idea what it exactly is. Not a big issue at the moment, however. Previously there was no code to parse it *at all*.


krzys-h:
Checked this really quickly but it doesn't seem to break anything too badly. I think we need a better name for "special sprite" though, it's not really special in the GMS2 world :P

PR #63, opened 2018-11-13
Attempt to fix switch statement default failing to decompile if it branches at the start

colinator27:
Read the comments in the code.

I found evidence that the "default" statement is always put after the other conditions, which would make sense, and I made sure that it doesn't do anything weird by testing my own code.

Basically, the situation was something like this:
```gml
switch(variable)
{
    case 0:
        if (anothervariable == 0)
            somethingelse = 1;
        break;
    default:
        if (anothervariable == 0)
            somethingelse = -1;
        break;
}
```

The switch statement decompiler got all messed up once it got to `default`, because there is an immediate `if` statement there. I guess that it interpreted that as part of the switch cases... and then it interpreted the expression `anothervariable == 0` as one of the case expressions, which in the end resulted in the error `Malformed switch statement: bad condition var (self.anothervariable)`. I think there's some weird branching for `default`, and that may be why the problem didn't occur on other normal cases.

Let me know if there are any issues with this.

### DIFF
--- a/UndertaleModLib/Decompiler/Decompiler.cs
+++ b/UndertaleModLib/Decompiler/Decompiler.cs
@@ -1453,6 +1453,23 @@ namespace UndertaleModLib.Decompiler
                         if (!caseEntries.ContainsKey(block.nextBlockTrue))
                             caseEntries.Add(block.nextBlockTrue, new List<Expression>());
                         caseEntries[block.nextBlockTrue].Add(caseExpr);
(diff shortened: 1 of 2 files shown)


colinator27:
Just squeezed in a fix for sprites.

krzys-h:
Looks good, thanks!

krzys-h:
92c732637099f1c179b01253aca8e9d53f5bc905

PR #76, opened 2018-11-15
Fix/finish EMBI, add TGIN internal support, fix EXTN potential crash

colinator27:
Exactly as the title says. This should fix a whole lot, including an oversight with EXTN I made. Apparently "product ID" only happens on newer games... (probably 1.4.9999 and 2+)

### DIFF
--- a/UndertaleModLib/Models/UndertaleFont.cs
+++ b/UndertaleModLib/Models/UndertaleFont.cs
@@ -131,7 +131,7 @@ namespace UndertaleModLib.Models
             writer.WriteUndertaleObjectPointer(Texture);
             writer.Write(ScaleX);
             writer.Write(ScaleY);
-            if (writer.undertaleData.GeneralInfo?.Major >= 2 || (writer.undertaleData.GeneralInfo?.Major == 1 && writer.undertaleData.GeneralInfo?.Build == 9999))
+            if (writer.undertaleData.GeneralInfo?.Major >= 2 && writer.undertaleData.GeneralInfo?.BytecodeVersion >= 17)
                 writer.Write(AscenderOffset);
             writer.WriteUndertaleObject(Glyphs);
         }
--- a/UndertaleModLib/Models/UndertaleFont.cs
+++ b/UndertaleModLib/Models/UndertaleFont.cs
@@ -21,6 +21,7 @@ namespace UndertaleModLib.Models
         private UndertaleTexturePageItem _Texture;
         private float _ScaleX;
         private float _ScaleY;
+        private int _AscenderOffset;
 
         public UndertaleString Name { get => _Name; set { _Name = value; PropertyChanged?.Invoke(this, new PropertyChangedEventArgs("Name")); } }
         public UndertaleString DisplayName { get => _DisplayName; set { _DisplayName = value; PropertyChanged?.Invoke(this, new PropertyChangedEventArgs("DisplayName")); } }
@@ -35,6 +36,7 @@ namespace UndertaleModLib.Models
         public float ScaleX { get => _ScaleX; set { _ScaleX = value; PropertyChanged?.Invoke(this, new PropertyChangedEventArgs("ScaleX")); } }
         public float ScaleY { get => _ScaleY; set { _ScaleY = value; PropertyChanged?.Invoke(this, new PropertyChangedEventArgs("ScaleY")); } }
         public UndertalePointerList<Glyph> Glyphs { get; private set; } = new UndertalePointerList<Glyph>();
+        public int AscenderOffset { get => _AscenderOffset; set { _AscenderOffset = value; PropertyChanged?.Invoke(this, new PropertyChangedEventArgs("AscenderOffset")); } }
 
         public event PropertyChangedEventHandler PropertyChanged;
 
@@ -45,16 +47,18 @@ namespace UndertaleModLib.Models
             private ushort _SourceY;
             private ushort _SourceWidth;
             private ushort _SourceHeight;
-            private ushort _Shift;
-            private uint _Offset; // TODO: probably signed
+            private short _Shift;
+            private int _Offset;
+            private UndertaleSimpleListShort<GlyphKerning> _Kerning;
 
             public ushort Character { get => _Character; set { _Character = value; PropertyChanged?.Invoke(this, new PropertyChangedEventArgs("Character")); } }
             public ushort SourceX { get => _SourceX; set { _SourceX = value; PropertyChanged?.Invoke(this, new PropertyChangedEventArgs("SourceX")); } }
             public ushort SourceY { get => _SourceY; set { _SourceY = value; PropertyChanged?.Invoke(this, new PropertyChangedEventArgs("SourceY")); } }
             public ushort SourceWidth { get => _SourceWidth; set { _SourceWidth = value; PropertyChanged?.Invoke(this, new PropertyChangedEventArgs("SourceWidth")); } }
             public ushort SourceHeight { get => _SourceHeight; set { _SourceHeight = value; PropertyChanged?.Invoke(this, new PropertyChangedEventArgs("SourceHeight")); } }
-            public ushort Shift { get => _Shift; set { _Shift = value; PropertyChanged?.Invoke(this, new PropertyChangedEventArgs("Shift")); } }
-            public uint Offset { get => _Offset; set { _Offset = value; PropertyChanged?.Invoke(this, new PropertyChangedEventArgs("Offset")); } }
+            public short Shift { get => _Shift; set { _Shift = value; PropertyChanged?.Invoke(this, new PropertyChangedEventArgs("Shift")); } }
+            public int Offset { get => _Offset; set { _Offset = value; PropertyChanged?.Invoke(this, new PropertyChangedEventArgs("Offset")); } }
+            public UndertaleSimpleListShort<GlyphKerning> Kerning { get => _Kerning; set { _Kerning = value; PropertyChanged?.Invoke(this, new PropertyChangedEventArgs("Kerning")); } }
 
             public event PropertyChangedEventHandler PropertyChanged;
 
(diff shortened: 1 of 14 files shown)


UndertaleModLib/Models/UndertaleTextureGroupInfo.cs
```diff
@@ -0,0 +1,146 @@
...
+            Sprites = new List<UndertaleResourceById<UndertaleSprite>>();
+            SpineSprites = new List<UndertaleResourceById<UndertaleSprite>>();
+            Fonts = new List<UndertaleResourceById<UndertaleFont>>();
+            Tilesets = new List<UndertaleResourceById<UndertaleBackground>>();
+        }
+
+        public void Serialize(UndertaleWriter writer)
+        {
+            writer.WriteUndertaleString(GroupName);
+            uint pointer1 = writer.Position; writer.Write(0);
```
krzys-h: Stuff like this was kinda supposed to be abstracted away into UndertaleReader/Writer to not clutter up code with manual pointer magic like this... (I also have to kinda fix it a bit to better handle non-standard pointers e.g. messed up by other tools, but that's another story). You can use WriteUndertaleObjectPointer as long as you make the target implement UndertaleObject (also see my note about lists which should handle that)

UndertaleModLib/Models/UndertaleTextureGroupInfo.cs
```diff
@@ -0,0 +1,146 @@
...
+            uint pointer2 = writer.Position; writer.Write(0);
+            uint pointer3 = writer.Position; writer.Write(0);
+            uint pointer4 = writer.Position; writer.Write(0);
+            uint pointer5 = writer.Position; writer.Write(0);
+
+            SeekWritePointer(writer, pointer1);
```
krzys-h: This is quite repetitive. How about implementing something like UndertaleSimpleList that handles UndertaleResourceById instead of normal objects? UndertaleSimpleResourcesList? Could probably be used in a couple more places as well, this is not the first time we see a sequence like that.

krzys-h:
See 3951acaf8f0efd3bfa79c92e3b2b4f9542c99d46 and 5b036f3fba8c265fc8c2ef62283f4a23e1506153, they should help with a cleaner implementation. You will need to update the code as the UndertaleResourceById syntax has changed.

You should now be able to simply do:
```csharp
SomeList = reader.ReadUndertaleObjectPointer<UndertaleSimpleResourcesList<SomeObjectType, UndertaleChunkXXXX>>();
//...
var returned = reader.ReadUndertaleObject<UndertaleSimpleResourcesList<SomeObjectType, UndertaleChunkXXXX>>();
if (returned != SomeList) throw new Exception("I need to make it check this automatically somehow but for now do it like this (UndertaleRoom works like that too)");
```

colinator27:
Alright, thanks. That should be pretty helpful.

colinator27:
Okay, so hold on a little bit. I may have just discovered a major problem...

colinator27:
Phew, that's fixed now... It should be fine.

krzys-h:
> However there are a few issues that may have to get fixed, for the codebase in general, about how reading of objects are done. Basically, if an object is never referred to as a pointer, then it should probably never be added to the object pool (or have anything to do with it). The reason why is because otherwise if you have an object at the start of another object, they can "overlap", and that causes errors when it tries to convert between the two types. 

Yeah, I know, I hit that limitation a couple of times. Having them all in the object map can be useful though, for example it's used for generating the offset map file that lists every single object read, which can aid in comparing hexdumps like we did for the two Switch versions that were almost identical. I'm still thinking of better ways to solve this, one may be to have a separate object map per object type.

UndertaleModLib/Models/UndertaleFont.cs
```diff
@@ -76,8 +87,33 @@ public void Unserialize(UndertaleReader reader)
...
-                Shift = reader.ReadUInt16();
-                Offset = reader.ReadUInt32();
+                Shift = reader.ReadInt16();
+                if (reader.undertaleData.GeneralInfo?.Major >= 2 || (reader.undertaleData.GeneralInfo?.Major == 1 && reader.undertaleData.GeneralInfo?.Build == 9999))
+                {
+                    Offset = reader.ReadInt16();
+                    Kerning = reader.ReadUndertaleObject<UndertaleSimpleListShort<GlyphKerning>>();
+                } else
+                {
+                    Offset = reader.ReadInt32(); // Maybe? I don't really know, but this definitely used to work
```
krzys-h: Maybe it was two bytes of data and two bytes of padding, or maybe the list always existed but was empty?

UndertaleModLib/Models/UndertaleFont.cs
```diff
@@ -95,6 +131,8 @@ public void Serialize(UndertaleWriter writer)
             writer.WriteUndertaleObjectPointer(Texture);
             writer.Write(ScaleX);
             writer.Write(ScaleY);
+            if (writer.undertaleData.GeneralInfo?.Major >= 2 && writer.undertaleData.GeneralInfo?.BytecodeVersion >= 17)
```
krzys-h: Are you sure this requires GMS2? I guess just bytecode 17 should do it? I don't think it's even possible to have 17 in GMS1.4 (and if you can, who knows what the behavior will be anyway)
```suggestion
            if (writer.undertaleData.GeneralInfo?.BytecodeVersion >= 17)
```

UndertaleModLib/Models/UndertaleFont.cs
```diff
@@ -112,6 +150,8 @@ public void Unserialize(UndertaleReader reader)
             Texture = reader.ReadUndertaleObjectPointer<UndertaleTexturePageItem>();
             ScaleX = reader.ReadSingle();
             ScaleY = reader.ReadSingle();
+            if (reader.undertaleData.GeneralInfo?.Major >= 2 && reader.undertaleData.GeneralInfo?.BytecodeVersion >= 17)
```
krzys-h:
```suggestion
            if (reader.undertaleData.GeneralInfo?.BytecodeVersion >= 17)
```

UndertaleModLib/Models/UndertaleRoom.cs
```diff
@@ -644,7 +644,8 @@ public void Serialize(UndertaleWriter writer)
 
                 public void Unserialize(UndertaleReader reader)
                 {
-                    _Background = reader.ReadUndertaleObject<UndertaleResourceById<UndertaleBackground, UndertaleChunkBGND>>();
+                    _Background = new UndertaleResourceById<UndertaleBackground, UndertaleChunkBGND>(); // see comment in UndertaleGlobalInit.Unserialize
```
krzys-h: whoops, missed that during my testing

UndertaleModLib/UndertaleData.cs
```diff
@@ -78,7 +80,9 @@ public static UndertaleData CreateNew()
             data.FORM.Chunks["OBJT"] = new UndertaleChunkOBJT();
             data.FORM.Chunks["ROOM"] = new UndertaleChunkROOM();
             data.FORM.Chunks["DAFL"] = new UndertaleChunkDAFL();
+            data.FORM.Chunks["EMBI"] = new UndertaleChunkEMBI();
             data.FORM.Chunks["TPAG"] = new UndertaleChunkTPAG();
+            data.FORM.Chunks["TGIN"] = new UndertaleChunkTGIN(); // Maybe a studio 2 option may come in handy
```
krzys-h: Yeah, CreateNew() right now generates a GMS1.4 file, so these shouldn't be here at all. But really, who would even attempt making a whole game from scratch in UndertaleModTool :P
```suggestion
```

UndertaleModLib/UndertaleData.cs
```diff
@@ -78,7 +80,9 @@ public static UndertaleData CreateNew()
             data.FORM.Chunks["OBJT"] = new UndertaleChunkOBJT();
             data.FORM.Chunks["ROOM"] = new UndertaleChunkROOM();
             data.FORM.Chunks["DAFL"] = new UndertaleChunkDAFL();
+            data.FORM.Chunks["EMBI"] = new UndertaleChunkEMBI();
```
krzys-h:
```suggestion
```

UndertaleModLib/Models/UndertaleFont.cs
```diff
@@ -76,8 +87,33 @@ public void Unserialize(UndertaleReader reader)
...
+                {
+                    Offset = reader.ReadInt16();
+                    Kerning = reader.ReadUndertaleObject<UndertaleSimpleListShort<GlyphKerning>>();
+                } else
+                {
+                    Offset = reader.ReadInt32(); // Maybe? I don't really know, but this definitely used to work
+                }
+            }
+
+            public class GlyphKerning : UndertaleObject
```
krzys-h: So I guess the editor needs an update...
How many entries of this do you usually get? I'm just wondering what controls would be best for this, since now every table row can have subrows...
(and font editing for translators just got harder, ugh)

krzys-h:
Also, please let me know if I missed anything we already know about bytecode 17 that should be listed on this page: https://github.com/krzys-h/UndertaleModTool/wiki/Bytecode-17

I should probably document other recent changes like the Spine sprites, the new font things and updates do GeneralInfo and Options flags, but I guess none of these are really related to the new bytecode right?

Ugh, we should seriously start writing proper documentation of the format like https://pcy.ulyssis.be/undertale/unpacking-corrected (but way more detailed, probably) instead of all these short patch documents

UndertaleModLib/Models/UndertaleFont.cs
colinator27: Actually, I feel stupid because I just realized none of the things I've tested on use the list at all. It doesn't really seem necessary to allocate too much space for them (if any).

UndertaleModLib/Models/UndertaleFont.cs
colinator27: I guess maybe it can assume there's always a list? No real harm in doing so.

krzys-h:
I was about to say that this suggestion feature really needs a way to accept multiple suggestions in one commit but...

> * To apply the change in its own commit, click **Commit suggestion**.
> * To add the suggestion to a batch of changes, click **Add suggestion to batch**. Continue to add the suggested changes you want to include in a single commit. When you've finished adding suggested changes, click **Commit suggestions**.

colinator27:
Wait, what?? I saw that button but it was grayed out...

krzys-h:
No idea, this was the first time I used that feature, I don't even know how it looks on your side :P

Kneesnap:
Is this resolved?

Kneesnap:
If so, it should be closed.

PR #108, opened 2018-12-01
Add basic support for the swap variant of pop instruction

colinator27:
There is a seemingly undiscovered variant of the pop instruction I just noticed while writing more of the compiler. It seems, from all I've found so far, used for incrementing/decrementing chain/stacktop variable references inside of expressions. All cases I've seen are exactly in the form `pop.e.v 5` according to the GM:S 1.4 debugger, though the number may vary (it's just in TypeInst apparently). I added support for assembly/disassembly, and a temporary exception for when the instruction is encountered in the decompiler in case anyone gets errors.

Since I can't really explain it well, here's some GML:
```gml
a++;
a.b.c++;
show_message(a++);
show_message(++a);
show_message(a.b.c++); // uses the instruction
show_message(++a.b.c); // uses the instruction
```

This now disassembles to this (previously it would fail to recognize the missing 4 bytes at the end of this instruction and fail to load the file):
```
00000: push.v self.a
00002: push.e 1
00003: add.i.v
00004: pop.v.v self.a

00006: push.v self.a
00008: conv.v.i
00009: push.v [stacktop]b
00011: conv.v.i
00012: dup.i 0
00013: push.v [stacktop]c
00015: push.e 1
00016: add.i.v
00017: pop.i.v [stacktop]c

00019: push.v self.a
00021: dup.v 0
00022: push.e 1
00023: add.i.v
00024: pop.v.v self.a
00026: call.i show_message(argc=1)
00028: popz.v

00029: push.v self.a
00031: push.e 1
00032: add.i.v
00033: dup.v 0
00034: pop.v.v self.a
00036: call.i show_message(argc=1)
00038: popz.v

00039: push.v self.a
00041: conv.v.i
00042: push.v [stacktop]b
00044: conv.v.i
00045: dup.i 0
00046: push.v [stacktop]c
00048: dup.v 0
00049: pop.e.v 5
00050: push.e 1
00051: add.i.v
00052: pop.i.v [stacktop]c
00054: call.i show_message(argc=1)
00056: popz.v

00057: push.v self.a
00059: conv.v.i
00060: push.v [stacktop]b
00062: conv.v.i
00063: dup.i 0
00064: push.v [stacktop]c
00066: push.e 1
00067: add.i.v
00068: dup.v 0
00069: pop.e.v 5
00070: pop.i.v [stacktop]c
00072: call.i show_message(argc=1)
00074: popz.v
```

There are other decompilation issues with this, though I'll put that in a separate issue.

### DIFF
--- a/UndertaleModLib/Decompiler/Assembler.cs
+++ b/UndertaleModLib/Decompiler/Assembler.cs
@@ -88,7 +88,12 @@ namespace UndertaleModLib.Decompiler
 
                 case UndertaleInstruction.InstructionType.PopInstruction:
                     UndertaleInstruction.InstanceType inst = instr.TypeInst;
-                    instr.Destination = ParseVariableReference(line, vars, localvars, ref inst, instr, lookOnStack);
+                    if (instr.Type1 == UndertaleInstruction.DataType.Int16 && instr.Type2 == UndertaleInstruction.DataType.Variable)
+                    {
+                        // Special scenario?
+                        inst = (UndertaleInstruction.InstanceType)Int32.Parse(line);
+                    } else
+                        instr.Destination = ParseVariableReference(line, vars, localvars, ref inst, instr, lookOnStack);
                     instr.TypeInst = inst;
                     line = "";
                     break;
--- a/UndertaleModLib/Decompiler/Decompiler.cs
+++ b/UndertaleModLib/Decompiler/Decompiler.cs
@@ -830,6 +830,8 @@ namespace UndertaleModLib.Decompiler
                         break;
 
                     case UndertaleInstruction.Opcode.Pop:
+                        if (instr.Destination == null)
+                            throw new Exception("Unsupported pop.e.v, this is a bug seemingly with incrementing/decrementing in expressions");
                         ExpressionVar target = new ExpressionVar(instr.Destination.Target, new ExpressionConstant(UndertaleInstruction.DataType.Int16, instr.TypeInst), instr.Destination.Type);
                         Expression val = null;
                         if (instr.Type1 != UndertaleInstruction.DataType.Int32 && instr.Type1 != UndertaleInstruction.DataType.Variable)
--- a/UndertaleModLib/Models/UndertaleCode.cs
+++ b/UndertaleModLib/Models/UndertaleCode.cs
@@ -427,7 +427,8 @@ namespace UndertaleModLib.Models
                         byte TypePair = (byte)((byte)Type2 << 4 | (byte)Type1);
                         writer.Write(TypePair);
                         writer.Write((byte)Kind);
-                        writer.WriteUndertaleObject(Destination);
+                        if (Destination != null)
+                            writer.WriteUndertaleObject(Destination);
                     }
                     break;
 
@@ -562,7 +563,11 @@ namespace UndertaleModLib.Models
                         Type1 = (DataType)(TypePair & 0xf);
                         Type2 = (DataType)(TypePair >> 4);
                         if(reader.ReadByte() != (byte)Kind) throw new Exception("really shouldn't happen");
-                        Destination = reader.ReadUndertaleObject<Reference<UndertaleVariable>>();
+                        if (Type1 == DataType.Int16 && Type2 == DataType.Variable)
+                        {
+                            // Special scenario?
+                        } else
+                            Destination = reader.ReadUndertaleObject<Reference<UndertaleVariable>>();
                     }
                     break;
 
@@ -678,7 +683,10 @@ namespace UndertaleModLib.Models
                         sb.Append(TypeInst.ToString().ToLower());
                         sb.Append(".");
                     }
-                    sb.Append(Destination.ToString());
+                    if (Destination != null)
+                        sb.Append(Destination.ToString());
+                    else
+                        sb.Append(TypeInst.ToString());
                     break;
 
                 case InstructionType.PushInstruction:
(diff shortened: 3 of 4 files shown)


krzys-h:
So here is what the stack probably looks like AFTER every instruction, for the last sample (++a.b.c)

[] denotes getting the value stored in a variable

```
push.v self.a                       ; self.a
conv.v.i                            ; [self.a]
push.v [stacktop]b                  ; [self.a].b
conv.v.i                            ; [[self.a].b]
dup.i 0                             ; [[self.a].b] [[self.a].b]
push.v [stacktop]c                  ; [[self.a].b] [[self.a].b].c
push.e 1                            ; [[self.a].b] [[self.a].b].c 1
add.i.v                             ; [[self.a].b] [[self.a].b].c+1
dup.v 0                             ; [[self.a].b] [[self.a].b].c+1 [[self.a].b].c+1
pop.e.v 5                           ; [[self.a].b].c+1 [[self.a].b] [[self.a].b].c+1 (MAGIC?! assuming our knowledge of pop.i.v behavior is correct, it needs to pop the value and instance, in that order)
pop.i.v [stacktop]c                 ; [[self.a].b].c+1
call.i show_message(argc=1)         ; (void)
popz.v                              ; stack empty
```

So.. that would be a swap of some kind? Exact semantics still unclear. The 5 is probably a type and e just marks that 'no this is in fact not a pop instruction', so the more correct representation would be `swap.i.v`, probably. The one from pcy.ulyssis.be would then be `swap.s.v`, meaning there may be more ways to trigger this opcode.

krzys-h:
So I created a new data.win from scratch (omg that seriously worked) and put the following test assembly code into it:
```
pushi.e 1
conv.i.v
pushi.e 2
conv.i.v
pushi.e 3
conv.i.v
pushi.e 4
conv.i.v
pushi.e 5
conv.i.v
pushi.e 6
conv.i.v

pop.e.v 5

call.i show_message(argc=1)
popz.v
call.i show_message(argc=1)
popz.v
call.i show_message(argc=1)
popz.v
call.i show_message(argc=1)
popz.v
call.i show_message(argc=1)
popz.v
call.i show_message(argc=1)
popz.v
```

And the result was: `5 0 3 2 1`, so it clearly messes with the top 3 stack values in some way. In my case it removed one though, I still can't figure out what the exact semantics of this operation are

krzys-h:
I can't figure out how the types work exactly, but I'm fairly certain that "5" is a type (changing it causes weird stack item size misalignment things)

Now, for the actual syntax... I hate to put something in the assembler that is not an actual opcode, but there is no way we are calling this operation a pop, so I propose `swap.t1.t2` for now instead of `pop.e.t1 t2`, unless anybody has any better ideas

colinator27:
By the way, I actually dug into the compiler and apparently it's 5 when it's not an array, and 6 when it is.

I agree with the naming though, `swap.t1.t2` would be much more fitting. This thing keeps getting more convoluted... how many levels deep does this go...?

krzys-h:
Wait, 6? That matches with pcy.ulyssis.be, but then, I'm not sure that value is actually a type. Why would they encode an array as 'string'.

krzys-h:
So this instruction is kinda getting weird

```
pushi.e 1
pushi.e 2
pushi.e 3
conv.i.v
pushi.e 4
conv.i.v

pop.e.v 6

call.i show_message(argc=1)
popz.v
conv.i.v
call.i show_message(argc=1)
popz.v
conv.i.v
call.i show_message(argc=1)
popz.v
call.i show_message(argc=1)
popz.v
```

Output is apparently `4 2 1 4`?!?! Which works for what it's supposed to do, I guess, but it's no longer a simple swap...

I also got it to work with 7:
```
pushi.e 1
pushi.e 2
pushi.e 3
pushi.e 4
conv.i.v
pushi.e 5
conv.i.v

pop.e.v 7

call.i show_message(argc=1)
popz.v
conv.i.v
call.i show_message(argc=1)
popz.v
conv.i.v
call.i show_message(argc=1)
popz.v
conv.i.v
call.i show_message(argc=1)
popz.v
call.i show_message(argc=1)
popz.v
```
This is apparently `5 3 2 1 5`. The pattern continues for pop.e.v 8. So apparently, we start dropping one of values on the stack an replace it with a copy of another one, but since they are supposed to be identical anyway it magically works. I wonder if that's an unnoticed bug in the runner or something :P

krzys-h:
(probably worth noting: i did the tests on the GMS1.4 runner because it's what UndertaleModTool runs on F5)

colinator27:
Yeah, this is *really weird*...

krzys-h:
I'm doing more testing. Seems like the number has something to do with the size, where something of type integer is '1' and something of type variable is '4' - or at least that assumption seems to give somewhat sensible results, or at least it crashes less :P Still no idea what the instruction DOES though, since

```
00000: pushi.e 1
00001: pushi.e 2
00002: conv.i.v
00003: pushi.e 3
00004: conv.i.v
00005: pushi.e 4
00006: conv.i.v
00007: pop.e.v 8
00008: call.i show_message(argc=1)
00010: popz.v
00011: call.i show_message(argc=1)
00013: popz.v
00014: call.i show_message(argc=1)
00016: popz.v
00017: conv.i.v
00018: call.i show_message(argc=1)
00020: popz.v
```
is apparently `4 2 4 1` and
```
00000: pushi.e 1
00001: conv.i.v
00002: pushi.e 2
00003: conv.i.v
00004: pushi.e 3
00005: conv.i.v
00006: pushi.e 4
00007: conv.i.v
00008: pop.e.v 12
00009: call.i show_message(argc=1)
00011: popz.v
00012: call.i show_message(argc=1)
00014: popz.v
00015: call.i show_message(argc=1)
00017: popz.v
00018: call.i show_message(argc=1)
00020: popz.v
```
is `4 3 1 4`

I think I'm just going to give up and hardcode the '5' and '6' values used by GMS compiler at this point :P

colinator27:
So how are we going to denote this in assembly text now, with these findings?

krzys-h:
Ugh... no idea. How about just changing `pop.e.X` to `swap.X` and keeping everything else as is? And we should also make sure that `pop.e.X` does not compile as the encoding for this variant is different.

PR #126, opened 2018-12-13
Some reasonable changes to TTFFonts.csx

BreakinBenny:
Not only did I scale the font sizes down to some more reasonable levels, I also cranked up the font range of fonts with letters past range 127, up to 255, to support Scandinavian languages. KÖR HÅRT, MINA EUROPEISKA GRANNAR!


krzys-h:
Does this even do anything? The end parameter was supposed to be ignored when loading from ttf file, and none of the characters I tested to check if it works were in the 0-127 range...

krzys-h:
Make sure the encoding of the lang file is correct. I'm not sure what it's supposed to be, but I heard some people had problems with that.

BreakinBenny:
Well, I edited the lang_en.json in Notepad++. I can't imagine that application just changing the encoding as it pleases.

BreakinBenny:
Here it is: https://github.com/BreakinBenny/UndertaleModTool/blob/master/lang_en.json

PR #127, opened 2018-12-16
Scaling down fonts...

BreakinBenny:
If the font range isn't going to work for others without my modified lang_en.json (translated into Swedish), then at the very least I can just reduce the font size to something reasonable.


krzys-h:
The font sizes were correct for Undertale, and TTFFonts is not uet officially supported for Deltarune - there is more stuff required for that to work correctly. Also, they are not 'something reasonable', all the parameters were actually copied from the normal font configs so that they match exactly.

BreakinBenny:
And since I had Deltarune in mind, I actually went ahead and edited TTFFonts accordingly. But maybe I should have madd a copy of UT's version and then altered the one for the latter game instead… seems logical if I do it like that.

Issue #40, opened 2018-11-03
REQUEST - Add TTFFonts script but for DeltaRune


Eleiber:
I am trying to do a translation to the spanish, so I need to add some characters like: "´", "¡" and "¿". So I modify a bit the TTFFonts to make it work good with Deltarune, but I don't know how to add the DotumChe font.
![image](https://user-images.githubusercontent.com/13054092/47958309-4b3ec880-df9f-11e8-8c50-22008a9b8843.png)
I already checked and yes, I have the DotumChe.ttf file in my SURVEY_PROGRAM folder, but it just doesn't works, it is the only font that is not working for me for DeltaRune.

D3fau4:
I recommend you better draw them with Photoshop, take less time

krzys-h:
It's going to be a bit more complicated than that because previously I used a terrible hack to force the IDs to be what the game expects and it's not going to work on deltarune because I think I've seen it use font_add_sprite-ext itself. It should be possible to hook into the new translation system to do replacements though.

Eleiber:
But what is the best option? Is it not possible to do with that script? I really would appreciate to know how to that.

krzys-h:
It's not possible to easily do what that script did in Deltarune, it needs to be completly rewritten. @D3fau4 is talking about adding the characters manually, which still works

Eleiber:
> For what it's worth. I applied the TTF script on top of Deltarune and it "works" but the line-height and letter-spacing are messed up. So I assume it's safer to edit the font sprites :\ (see screenshot below)
> ![unknown](https://user-images.githubusercontent.com/1409924/47970070-580dfb80-e035-11e8-8be3-0cb314e68649.png)
> 
> EDIT: nevermind the sizing is messed up too
> ![screen shot 2018-11-04 at 1 31 55 pm](https://user-images.githubusercontent.com/1409924/47970146-39f4cb00-e036-11e8-9b27-f5540600d979.png)

That happened to me, you only have to modify the size of the font in the TTF script.

eramdam:
@Eleiber Ah, neat. I ended up re-creating spritesheets and glyphs map in the meantime 🙈

NotedChord:
wingding.ttf like used in dark world

AsrielOver:
una pregunta me puedes decir como modificar los TTFONTS para añadir la á é í ó ú y la ñ ?

LinXP:
@krzys-h 
[...]

PR #128, opened 2018-12-16
If Undertale's TTFFonts has its own things...

BreakinBenny:
...then I may as well just include a _separate_ TTFFonts.csx specifically for Deltarune. Do what you want with it.

### DIFF
--- a/UndertaleModTool/SampleScripts/TTFFonts (Deltarune).csx
+++ b/UndertaleModTool/SampleScripts/TTFFonts (Deltarune).csx
@@ -1,0 +1,175 @@
+EnsureDataLoaded();
+
+// Remove all current fonts
+// This is necessary because I need to add fonts under IDs that are normally used by these resources
+// TODO: Japanese fonts won't work at all because I didn't add support for that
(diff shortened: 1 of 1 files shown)


krzys-h:
Related: #40

Does this even work properly? If I remember correctly, Deltarune has a totally different order of the fonts, AND it uses font_add_sprite_something itself for the damage font, breaking the hacky assumption about font IDs that made the original script work.

BreakinBenny:
Besides Noelle's "tiny Noelle" text not being shown, the normal text in the light world (and during CONTACT) works well if you also discount the dialogue choices being lower positioned lower on the Y axis than usual

The text during the Legend also works it uses the same programming (and fonts, by extension, I think). The answer is there for a yes, and a no.

krzys-h:
Yeah... the script is supposed to reproduce the original fonts 100%. It worked like that in Undertale.

Grossley:
It's fine to merge this, but this script will be added to the Experimental Scripts since it has not been tested thoroughly.

PR #162, opened 2019-04-10
Decompiler improvements

Kneesnap:
- Lots of minor changes to make the decompiled code fully supported by GMS1. (GMS2 untested)
- Support && + || if statements.
- Fixed numerous bugs, including `if (temp_var_1 - 1)` and a `break` bug causing some code to be omitted from the decompilation.
- Improved return logic,
```
Before:
if (condition)
 return 1
else
 return 0

After:
return condition
```
- Added option for scripts to show progress bars.


Kneesnap:
FYI, this pull request is ready.

colinator27:
Wait, that "decompilation bug" with the decimal... isn't that what `CultureInfo.InvariantCulture` is for? Like how it's used [here](https://github.com/krzys-h/UndertaleModTool/blob/619de0d387742cac756828951910cf4423da59ab/UndertaleModLib/Decompiler/Decompiler.cs#L248).

Kneesnap:
> Wait, that "decompilation bug" with the decimal... isn't that what `CultureInfo.InvariantCulture` is for? Like how it's used [here](https://github.com/krzys-h/UndertaleModTool/blob/619de0d387742cac756828951910cf4423da59ab/UndertaleModLib/Decompiler/Decompiler.cs#L248).

Oh, good catch. The problem with using that snippet of code is that it allows exporting decimals in exponential form, which GameMaker studio does not support.

colinator27:
> > Wait, that "decompilation bug" with the decimal... isn't that what `CultureInfo.InvariantCulture` is for? Like how it's used [here](https://github.com/krzys-h/UndertaleModTool/blob/619de0d387742cac756828951910cf4423da59ab/UndertaleModLib/Decompiler/Decompiler.cs#L248).
> 
> Oh, good catch. The problem with using that snippet of code is that it allows exporting decimals in exponential form, which GameMaker studio does not support.

Ah, alright. I guess that makes sense then.

UndertaleModLib/Decompiler/AssetTypeResolver.cs
```diff
@@ -484,13 +490,28 @@ public class AssetTypeResolver
             { "room_first", AssetIDType.Room },
             { "room_last", AssetIDType.Room },
             { "room", AssetIDType.Room },
+            { "currentroom", AssetIDType.Room },
```
krzys-h: This is UT/DR specific variable too, the GMS version of this is just "room"

UndertaleModLib/Decompiler/Decompiler.cs
```diff
@@ -1597,10 +1776,13 @@ private static List<Statement> HLDecompile(Dictionary<uint, Block> blocks, Block
 
             // TODO: This doesn't propagate to tempvars in broken switch statements but that will be cleaned up at some point
             AssetIDType propType = AssetIDType.Other;
-            if (code.Name.Content == "gml_Script___view_set_internal" || code.Name.Content == "gml_Script___view_get")
-                propType = AssetIDType.e__VW;
-            if (code.Name.Content == "gml_Script___background_set_internal" || code.Name.Content == "gml_Script___background_get_internal")
-                propType = AssetIDType.e__BG;
+            // NOTE: This was disabled because it wasn't valid.
+            // 1. The type only applied to the first case statement.
+            // 2. Even if each case statement had the proper type, it still gives an error when compiling.
```
krzys-h: It throws an error because the enum needs to be defined in a script, see `C:\ProgramData\GameMakerStudio2\Cache\runtimes\runtime-X.X.X.XXX\lib\compatibility.zip`, specifically `background\__init_background.gml` and `view\__init_view.gml`

When going for full decompilation you may want to copy the scripts from there (if GMS2 is installed) instead of decompiling them since they have proper comments etc. - although it is technically possible for somebody to modify them...

And type propagation in the case statement is another problem, it happens because switch statements are resolved after type propagation so during type propagation you have some tempvar nonsense going on

UndertaleModLib/Models/UndertaleGameObject.cs
```diff
@@ -373,6 +373,7 @@ public enum EventType : uint
         KeyRelease = 10, // subtype is key ID, see EventSubtypeKey
         Gesture = 11, // TODO: mapping is a guess // TODO: subtypes
         Asynchronous = 12, // TODO: mapping is a guess // TODO: subtypes
+        PreCreate = 14
```
krzys-h: wait, 14? what was 13 then? I don't really remember... at least add an Unknown13 value here

UndertaleModLib/UndertaleModLib.csproj
```diff
@@ -30,11 +30,13 @@
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="PresentationCore" />
     <Reference Include="System" />
     <Reference Include="System.Drawing" />
     <Reference Include="System.Xml.Linq" />
     <Reference Include="Microsoft.CSharp" />
     <Reference Include="System.Net.Http" />
+    <Reference Include="WindowsBase" />
```
krzys-h: UndertaleModLib should avoid having any Windows-specific dependencies so that it can be easily used in cross-platform projects

UndertaleModLib/Models/UndertaleTexturePageItem.cs
```diff
@@ -27,6 +30,13 @@ public class UndertaleTexturePageItem : UndertaleResource, INotifyPropertyChange
         private ushort _BoundingHeight;
         private UndertaleResourceById<UndertaleEmbeddedTexture, UndertaleChunkTXTR> _TexturePage = new UndertaleResourceById<UndertaleEmbeddedTexture, UndertaleChunkTXTR>();
 
+        private static int maxSpriteWidth = 0;
+        private static int maxSpriteHeight = 0;
+        private static UInt32[] pixelBuffer = null; // a memory buffer for the largest possible sprite.
```
krzys-h: Keeping a large buffer forever sounds like a bad idea, and I don't really like these being cached as static in the model class anyway. It would probably be better to have a separate TextureExporter class for this that can be properly garbage collected when you are done exporting the whole batch of textures (since I assume this was for performance reasons).

And while you're at it, feel free to replace the UndertaleModTool exporting implementation with this since that one reuses WPF display code instead of doing it properly and I feel like it's still sometimes buggy. And add importing for sprites where no scaling or cropping is required since it looks way easier with this approach than the old one.

UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -761,39 +767,70 @@ private void MenuItem_RunBuiltinScript_SubmenuOpened(object sender, RoutedEventA
...
                 MessageBox.Show(exc.Message, "Script compile error", MessageBoxButton.OK, MessageBoxImage.Error);
             }
             catch (Exception exc)
             {
-                Debug.WriteLine(exc.ToString());
-                CommandBox.Text = exc.Message;
-                MessageBox.Show(exc.Message, "Script error", MessageBoxButton.OK, MessageBoxImage.Error);
+                Console.WriteLine(exc.ToString());
+                Dispatcher.Invoke(() => CommandBox.Text = exc.Message);
+                ScriptMessage("Script Error: " + exc.Message + "\n" + exc.ToString());
```
krzys-h: I preferred the MessageBox tbh, it grabs attention more than the output window

UndertaleModLib/Models/UndertaleTexturePageItem.cs
```diff
@@ -76,5 +86,99 @@ public override string ToString()
...
+                    if ((int)texPageItem.BoundingHeight > maxSpriteHeight)
+                        maxSpriteHeight = texPageItem.BoundingHeight;
+                }
+                pixelBuffer = new UInt32[maxSpriteWidth * maxSpriteHeight]; // Allocate the memory buffer.
+            }
+
+            for (int x = 0; x < pixelBuffer.Length; x++)
+                pixelBuffer[x] = (UInt32)0U;
+
+            System.Windows.Int32Rect rect = default;
```
krzys-h: My Visual Studio complains that `= default` is a part of C# 7.1 and the project is configured for C# 7.0

UndertaleModLib/Models/UndertaleTexturePageItem.cs
```diff
@@ -76,5 +86,99 @@ public override string ToString()
...
+            // Clear the render target.
+            if (pixelBuffer == null)
+            {
+                foreach (var texPageItem in data.TexturePageItems)
+                {
+                    if ((int)texPageItem.BoundingWidth > maxSpriteWidth)
+                        maxSpriteWidth = texPageItem.BoundingWidth;
+                    if ((int)texPageItem.BoundingHeight > maxSpriteHeight)
+                        maxSpriteHeight = texPageItem.BoundingHeight;
+                }
```
krzys-h: What if somebody adds a new, big TexturePageItem after the buffer was already allocated?

UndertaleModLib/Decompiler/AssetTypeResolver.cs
```diff
@@ -484,13 +490,28 @@ public class AssetTypeResolver
...
             { "object_index", AssetIDType.GameObject },
             { "sprite_index", AssetIDType.Sprite },
             { "image_blend", AssetIDType.Color },
             { "event_object", AssetIDType.GameObject },
             { "keyboard_key", AssetIDType.KeyboardKey },
             { "keyboard_lastkey", AssetIDType.KeyboardKey },
             { "os_type", AssetIDType.Enum_OSType },
+            
+            // These ones are ordinary variables with specific meaning within Undertale/Deltarune.
+            // It would probably be better to only use these if UT/DR are being loaded.
```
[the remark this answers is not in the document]
Kneesnap: Agreed, eventually, we should be able to remove these.

UndertaleModLib/Decompiler/AssetTypeResolver.cs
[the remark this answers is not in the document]
Kneesnap: Oh. Yeah, appears so. We should remove it when we remove the others. Maybe we should move it down to where the others are just so we don't forget.

UndertaleModLib/Decompiler/Decompiler.cs
```diff
@@ -219,18 +221,33 @@ public override string ToString()
...
-                        return "'" + (char)val + "'";
-                    if (val > 0 && Enum.IsDefined(typeof(EventSubtypeKey), (uint)val))
-                        return ((EventSubtypeKey)val).ToString();
+                        return ((char)val) == '\'' ? (isGMS2 ? "\"\\\"\"" : "'\"'")
+                            : (((char) val) == '\\' ? (isGMS2 ? "\"\\\\\"" : "\"\\\"")
+                            : "\"" + (char)val + "\"");
                 }
 
+                if (Value is float) // Prevents scientific notation by using high bit number.
+                    return ((decimal) ((float) Value)).ToString(CultureInfo.InvariantCulture);
```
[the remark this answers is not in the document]
Kneesnap: Technically "F99" will give you 99 0s, buuuuuut it makes more of a mess than this, because you have to get rid of trailing zeros, then handle the decimal point, and it can get really messy. This seemed like the cleanest fix.

UndertaleModLib/Decompiler/Decompiler.cs
```diff
@@ -132,41 +133,37 @@ internal override bool IsDuplicationSafe()
                 return true;
             }
 
+            public int GetIntegerValue()
+            {
+                try
+                {
+                    return Convert.ToInt32(Value);
+                }
+                catch (OverflowException)
```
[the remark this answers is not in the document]
Kneesnap: Will take a look at that.

UndertaleModLib/Decompiler/Decompiler.cs
```diff
@@ -647,10 +717,13 @@ internal override AssetIDType DoTypePropagation(AssetIDType suggestedType)
...
                 if (current == AssetIDType.Other && suggestedType != AssetIDType.Other)
-                    current = assetTypes[Var] = suggestedType;
+                    current = suggestedType;
                 AssetIDType builtinSuggest = AssetTypeResolver.AnnotateTypeForVariable(Var.Name.Content);
                 if (builtinSuggest != AssetIDType.Other)
-                    current = assetTypes[Var] = builtinSuggest;
+                    current = builtinSuggest;
+
+                if ((VarType != UndertaleInstruction.VariableType.Array || (ArrayIndex1 != null && !(ArrayIndex1 is ExpressionConstant))))
+                    assetTypes[Var] = current; // This is a messy fix to arrays messing up exported variable types.
```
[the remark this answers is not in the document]
Kneesnap: There was a script, which used global.flag. It did global.flag[some_index] = caster_load. It made the rest of the script think every usage of global.flag must be a sound type. So, it was replacing numbers (including decimal numbers) with sound ids. If I can find what script it was, I'll post back here.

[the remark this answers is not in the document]
Kneesnap: It was obj_ambientrain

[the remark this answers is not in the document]
krzys-h: Ugh, why would he put a sound in the flags... doesn't this break most other cases of using arrays of sprites/sounds/etc though? Things like
```
sounds[0] = snd_sth1
sounds[1] = snd_sth2
snd_play(sounds[0])
```

UndertaleModLib/Decompiler/Decompiler.cs
```diff
@@ -779,14 +852,21 @@ internal static void DecompileFromBlock(Block block, List<TempVarReference> temp
...
-                        if (!(popped is ExpressionTempVar))
-                            statements.Add(popped);
+                        if (stack.Count > 0)
+                        {
+                            Expression popped = stack.Pop();
+                            if (!(popped is ExpressionTempVar))
+                                statements.Add(popped);
+                        } else
+                        {
+                            statements.Add(new CommentStatement("WARNING: Popz'd an empty stack."));
```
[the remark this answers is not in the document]
Kneesnap: I wasn't able to fix the bug, and there was no functional consequence in the specific scenario. Makes sense though.

Here's what the code looks like, with the comments:
```
switch dialog_step {
  case 0:
    {
     if (global.enemy_hp_check <= (global.enemy_hp_max * 0.75)) {
      dialog_step = 1
      dialog_active = 1
      cur_message = 14
      text_location = 3
      text_flipped = 1
      alarm[3] = room_speed
      exit
     }
  // Something was wrong with the stack, reentered the block with 1 variables instead of 0, ignoring
  }
  break
  case 1:
    {
     if (global.enemy_hp_check <= (global.enemy_hp_max * 0.5)) {
      dialog_step = 2
      dialog_active = 1
      cur_message = 16
      text_location = 3
      text_flipped = 1
      alarm[3] = room_speed
      exit
     }
  // Something was wrong with the stack, reentered the block with 1 variables instead of 0, ignoring
  }
 // WARNING: Popz'd an empty stack.
 alarm[1] = room_speed
 // Something was wrong with the stack, reentered the block with 1 variables instead of 0, ignoring
 // Something was wrong with the stack, reentered the block with 1 variables instead of 0, ignoring
}
```

Source:
obj_barghest2_controller_Step_0 - Heartbound Demo

I reviewed the byte-code, and I wasn't sure what was going wrong in the decompiler, but I determined that these issues were related and weren't supposed to leave any actual decompiled code behind.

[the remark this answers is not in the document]
Kneesnap: If we can fix the underlying issue, that would be the best solution.

UndertaleModLib/Decompiler/Decompiler.cs
Kneesnap: Good to note. Hmm. I think if we want to expect someone to grab stuff from a compatibility zip file, we should tell them somehow, if not grab it ourselves.

UndertaleModLib/Models/UndertaleGameObject.cs
Kneesnap: I don't know what 13 was, but I can say with confidence that 14 is PreCreate. Will add unknown 13.

UndertaleModLib/UndertaleModLib.csproj
Kneesnap: Hmm. What am I even using that for? I can't find it lol. Just a moment.

Kneesnap: Odd. "Project file must include the .NET Framework assembly 'WindowsBase' in the reference list."

Kneesnap: I'm not very familiar with C#, so I don't really know how to deal with that.

UndertaleModTool/MainWindow.xaml.cs
Kneesnap: Huh? It still makes a MessageBox.

UndertaleModLib/Models/UndertaleTexturePageItem.cs
Kneesnap: Good point. This is why we'll be making a TextureExporter class which can be instanced and thrown out.

UndertaleModLib/Models/UndertaleTexturePageItem.cs
Kneesnap: Good point, it can't be thread-safe either if it's some static buffer like that. I'll fix that.

I won't add importing right now, but I'll add it to my TODO list.

UndertaleModLib/Decompiler/AssetTypeResolver.cs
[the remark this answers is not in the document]
Kneesnap: (Moved)

UndertaleModLib/Decompiler/Decompiler.cs
```diff
@@ -628,13 +683,28 @@ public override string ToString()
...
+                    name = name + "[" + ArrayIndex1.ToString() + ", " + ArrayIndex2.ToString() + "]";
+                else if (ArrayIndex1 != null)
                     name = name + "[" + ArrayIndex1.ToString() + "]";
-                if (ArrayIndex2 != null)
-                    name = name + "[" + ArrayIndex2.ToString() + "]";
-                name = InstType.ToString() + "." + name;
 
-                return name;
+                string prefix = InstType.ToString() + ".";
+                if (InstType is ExpressionConstant) // Only use "global." and "other.", not "self." or "local.". GMS doesn't recognize those.
```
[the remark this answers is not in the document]
Kneesnap: Done. I'm not sure if you had anywhere in mind, so I put it where the little preview highlighted.

UndertaleModLib/UndertaleModLib.csproj
Kneesnap: If I were to guess, it might be the image exporter, since it uses "System.Windows". I'm not sure if that's windows like the OS, or something else GUI related though. I'll wait for your input.

UndertaleModLib/Models/UndertaleTexturePageItem.cs
Kneesnap: Will fix.

UndertaleModLib/Decompiler/Decompiler.cs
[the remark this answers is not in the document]
krzys-h: Ok, leave it then and open an issue so that we don't forget about this

UndertaleModTool/MainWindow.xaml.cs
krzys-h: Ah, right, my bad, ScriptMessage is still a MessageBox 😂 It doesn't have the error icon though.

UndertaleModLib/UndertaleModLib.csproj
krzys-h: Yeah, the `System.Windows.Media` namespace is part of WPF which is not supported outside of Windows (https://www.mono-project.com/docs/gui/wpf/) and it's the reason why the main UndertaleModTool GUI is Windows-only. It may be better to use the classes from `System.Drawing` instead (and it looks like that one is even already included)

In case you find it useful when implementing this, here is the code I used when I made the Jevil hitboxes video that uses System.Drawing for drawing on the sprites: https://gist.github.com/krzys-h/81ee306d4b799972ed50c62b71d5c8cc

UndertaleModTool/MainWindow.xaml.cs
Kneesnap: True, could probably fix that.

UndertaleModLib/UndertaleModLib.csproj
Kneesnap: Thanks. Should help me.

UndertaleModLib/Decompiler/Decompiler.cs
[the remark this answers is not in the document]
Kneesnap: I believe so.

UndertaleModLib/Decompiler/Decompiler.cs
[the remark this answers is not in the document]
Kneesnap: Issue created.

UndertaleModTool/MainWindow.xaml.cs
Kneesnap: Follow-up, I will fix it so it won't use ScriptMessage, but use MessageBox.Show again. That was a temporary thing I was doing for testing.

Kneesnap:
@krzys-h Issue discovered. Scripts like ToggleDebug.csx throw errors upon usage, because they're no longer running on the UI thread. I'd like your input on how best we handle this.

UndertaleModLib/UndertaleModLib.csproj
```diff
@@ -30,6 +30,7 @@
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="PresentationCore" />
```
krzys-h: PresentationCore is a part of WPF as well and should be removed from here too - I don't think it's even used anywhere

krzys-h:
> @krzys-h Issue discovered. Scripts like ToggleDebug.csx throw errors upon usage, because they're no longer running on the UI thread. I'd like your input on how best we handle this.

Yeah... I was afraid things like this would come up if we try to make it multithreaded, you can't add new objects in scripts now because that needs to trigger an update to the object list on the left. The only reasonable fix I can think of right now that doesn't involve rewriting half of the tool is to run the script on the main thread and somehow force the progress bar to update when UpdateProgress is called.

UndertaleModLib/UndertaleModLib.csproj
Kneesnap: Done.

UndertaleModLib/Decompiler/Decompiler.cs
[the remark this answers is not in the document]
Kneesnap: It's been fixed to not need the exception.

However, I still don't understand why some of the values are Int64 in the situations they're in. We may want to look further into it to make sure the fix has no side-effects.

Kneesnap:
> > @krzys-h Issue discovered. Scripts like ToggleDebug.csx throw errors upon usage, because they're no longer running on the UI thread. I'd like your input on how best we handle this.
> 
> Yeah... I was afraid things like this would come up if we try to make it multithreaded, you can't add new objects in scripts now because that needs to trigger an update to the object list on the left. The only reasonable fix I can think of right now that doesn't involve rewriting half of the tool is to run the script on the main thread and somehow force the progress bar to update when UpdateProgress is called.

The script thread issue has been resolved.

UndertaleModLib/Models/UndertaleTexturePageItem.cs
Kneesnap: Importing has been added.

Kneesnap:
I have no more changes planned for this PR.

krzys-h:
Nice! I'll double-check everything and probably merge it tpday.

Kneesnap:
I discovered a last-minute bug regarding exporting textures which seemed to happen on some computers and not others. I've fixed it though. I think I'm gonna do a few more minor decompiler improvements while I'm working on this. I don't mean this to delay the merge though.

Kneesnap:
@krzys-h I'm done.

Kneesnap:
Alright, I think that's everything. I'm able to export a working project without issue.

UndertaleModLib/Decompiler/Decompiler.cs
```diff
@@ -1450,9 +1717,37 @@ private static BlockHLStatement HLDecompileBlocks(ref Block block, Dictionary<ui
...
                     alreadyVisited.Add(block);
                 }
-                foreach (var stmt in block.Statements)
+                
+                for (int i = 0; i < block.Statements.Count; i++)
+                {
+                    Statement stmt = block.Statements[i];
                     if (!(stmt is PushEnvStatement) && !(stmt is PopEnvStatement))
+                    {
+                        bool handled = false;
```
krzys-h: unused variable

colinator27:
Previously it just represented negative origin as very large positive numbers (>~2.14 billion) due to the nature of signed/unsigned integers. This obviously didn't lose any data, but it makes it harder to deal with. So that fixes it for the library and editor in one, though the editor didn't need to be touched at all.

krzys-h:
dog2

PR #241, opened 2019-11-20
Persistent configuration pull

DevCortez:
Added persistent configuration to use within scripts


Kneesnap:
This pull request should be tweaked to keep UndertaleModLib completely platform independent. (IE: Don't use any Windows-specific-code, and I believe the .NET core is Windows-specific.) Other than that this seems like a good addition.

DevCortez:
I'm pretty sure the code is os agnostic but i don't have the means to test it, whenever im able ill do so

Grossley:
Changing branch of pull request as dev branch has been obsoleted

NyakoFox:
What's the status on this PR?

PR #242, opened 2019-11-21
AssetTypeResolver improvements for Undertale and Deltarune

Grossley:
This should improve the decompiled code quality for both Undertale and Deltarune.


colinator27:
Seems all good, and shouldn't really break much (if anything at all)

PR #248, opened 2019-12-24
Patching grievous 96 DPI scaling error in the UI

Grossley:
I have made a patch for importing embedded texture sheets and individual images via the UI. 

To correct scripts that apply images directly they will need:

```
using System.IO;
using System.Drawing;
```
and directly setting the DPI of the image to be imported using the [Bitmap.SetResolution(Single, Single) Method as described in the docs.](https://docs.microsoft.com/en-us/dotnet/api/system.drawing.bitmap.setresolution?view=netframework-4.8) in a similar manner to this: 

```
Bitmap bitmap = new Bitmap(100, 100);
bitmap.SetResolution(96.0F, 96.0F);
```

Also other asset resolution improvements and making a more descriptive error message for a certain kind of error.


colinator27:
This also includes many improvements to asset types in UT/DR. Though, the commit history will be flooded with "Update AssetTypeResolver.cs". Doesn't matter that much at this point so I think it's fine to merge.

PR #249, opened 2019-12-25
DPI-Fix (for scripts)

Mc-muffin:
Test this one pls


Grossley:
I don't think this should be merged. I was not aware that the DPI commits break UndertaleModTool's ability to "recreate a byte-for-byte exact copy from the decoded data" at all, much less to the extent that I saw here. Also, by its nature, the TPAG padding commit also breaks UndertaleModTool's "byte-for-byte exact copy" ability. This accuracy is important.

Mc-muffin:
With the latest commits that concern is no more.

PR #254, opened 2020-01-14
Improve color decompilation.

nkrapivin:
Don't represent colors as HEX values because when copied to GM:S 1.4 they don't work.
And if the color matches the color from a constant, return constant's name as a string.
$00FFFFFF which is supposed to be white, becomes yellow.
UInts on the other hand don't break like this.

### DIFF
--- a/UndertaleModLib/Decompiler/Decompiler.cs
+++ b/UndertaleModLib/Decompiler/Decompiler.cs
@@ -29,26 +29,29 @@ namespace UndertaleModLib.Decompiler
         public Dictionary<string, AssetIDType[]> scriptArgs = new Dictionary<string, AssetIDType[]>();
 		
 		// Color dictionary for color resolving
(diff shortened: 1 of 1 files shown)


nkrapivin:
YellowAfterlife said not to use the Constants dictionary from BuiltinList.cs but instead make a new one only with colors somewhere outside of ToString method, maybe take his idea into count?
So we'll have two dictionaries, one with colors only and second with all the constants +colors.

Kneesnap:
Colors are represented as hex in GMS1.4. The color names are just macros for certain hex colors. The problem with converting all colors to names is that those numbers aren't always used in the context of a color. It's possible to be using a number for a non-color purpose that happens to have the same hex representation as a color. For instance, $000000 is a color, but it's also the number 0. We don't want to replace every usage of the number 0 wirh a color name.

PR #265, opened 2020-02-01
Fix for Pizza Tower

Mc-muffin:
Simple fix to allow adding objects to Pizza Tower (GameMaker 2.2.2.302+ games) without crashing the runner.


colinator27:
I apologize for forgetting, but what does this fix exactly? I think having the field static may cause issues when loading multiple data files.

Grossley:
It does cause issues when loading multiple data files. This alters the room format such that editing a room does not crash the runner.

colinator27:
Okay, well, I'm going to close this because of the issues loading multiple data files. However, I just came up with a completely different fix which I'll push to master in a moment.

PR #304, opened 2020-07-28
Fix Border Images File Path Issue

omerakgoz34:


colinator27:
I think that this looks pretty good, but one thing I'm not sure about is this line: https://github.com/krzys-h/UndertaleModTool/pull/304/files#diff-3c281b570aa066288fe955b4551014d2R123.

It might be fine, but I'm not sure if that's the best way to get the application directory or not.

omerakgoz34:
@colinator27 When UMT opened via clicking on data.win file, the script wasn't able to find border images because C# tells to UMT exe working path is the path of the data.win file.

To be honest, I don't know the C# language but I made some searches on forums and I found this code as best way to get the actual path of the UMT exe. And it works without any problem :D

colinator27:
Alright, well I think it should work, so I'll merge it. If there's any issues with that function then we can always fix it later.

PR #305, opened 2020-08-02
Document function classifications and other unknown values.

nkrapivin:
self explanatory.
Kneesnap told me that "UTMT should understand this value more"
now it kinda does.


colinator27:
Oh, cool, though I have one question about this: when is this field ever used?
https://github.com/krzys-h/UndertaleModTool/pull/305/files#diff-d03ce460188701d4a8e02235dcb31534R15

nkrapivin:
As for now, nowhere. I don't quite understand how do I implement this.

PR #306, opened 2020-08-06
Direct Padding

Snaggly:
Jumps straight to desired address without looping through the file while loading or saving. By default, jumping to a higher address on a new file will zero the previous bytes, so there is no need to constantly write zeros. This may increase performance...

### DIFF
--- a/UndertaleModLib/Models/UndertaleEmbeddedAudio.cs
+++ b/UndertaleModLib/Models/UndertaleEmbeddedAudio.cs
@@ -25,8 +25,7 @@ namespace UndertaleModLib.Models
 
         public void SerializePadding(UndertaleWriter writer)
         {
-            while (writer.Position % 4 != 0)
-                writer.Write((byte)0);
+            writer.Position += (4 - (writer.Position % 4)) % 4;
         }
 
         public void Unserialize(UndertaleReader reader)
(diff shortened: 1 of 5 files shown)


colinator27:
I doubt this will increase performance, but it's nice to have nonetheless I suppose. Though, is that behavior of filling in zero bytes documented anywhere? I just want to make sure it doesn't break on obscure implementations or something.

Snaggly:
https://docs.microsoft.com/en-us/dotnet/api/system.io.filestream.position?view=netcore-3.1#System_IO_FileStream_Position
Under Remarks: "In Microsoft Windows NT and newer, any data added to the end of the file is set to zero."

colinator27:
Hmmm I'm a bit worried, then, since that might cause issues with things like Mono

Snaggly:
I'm using Wine+Mono here, it works as it should. Though that's all I can say, as I don't know exactly how they implemented the handling here.

colinator27:
So I'll merge it now, and if anybody reports any issues in the future, we could always revert it, but hopefully that won't be the case.

colinator27:
Okay so, apparently this is currently causing issues for some people, and now that I think about it, skipping those bytes while reading is a bad idea in the first place (as it doesn't actually verify the padding), so I'm going to revert this for now.

Issue #355, opened 2020-12-27
UTMT makes incorrect assumptions when creating new room layers


BenjaminUrquhart:
The room editor assumes the last room in the room list will have the highest layer id (see https://github.com/krzys-h/UndertaleModTool/blob/master/UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs#L434). However, this assumption is invalidated once new layers are added, meaning multiple layers will share the same id. This can lead to the game hard crashing or not responding under certain circumstances.

Video (triggered by suspending and reactivating all objects a few times, then debug restarting the game. Can also be triggered by leaving the room):  

https://user-images.githubusercontent.com/24701883/103180875-0e617a00-4868-11eb-83c6-8549ebdf9650.mp4

BenjaminUrquhart:
I've manually updated the layer ids and it does indeed fix the issue.

colinator27:
Fixed via #356

PR #356, opened 2020-12-27
Ensure layers ids are not duplicated.

BenjaminUrquhart:
See #355


colinator27:
The only problem I guess now would be modifying layers with scripts, which wouldn't increment the counter. Maybe it should be re-evaluated every time, which hopefully wouldn't be too costly for performance since it's only a few hundred iterations.

BenjaminUrquhart:
That's true.

BenjaminUrquhart:
That's true. Should I move the code to the editor then?

BenjaminUrquhart:
Thinking about it, I don't see people adding a bunch of layers by hand anyway so the potential performance impact shouldn't matter as much.

colinator27:
I think it should be fine to move that to the editor, yeah

PR #371, opened 2021-01-21
Temporary script to edit tile layers.

Lassebq:
Just a temporary script to edit tile layers. Currently GMS2 tile layer's data cannot be edited from editor window.


Grossley:
Can you specify that this script is for GMS2 tile data as opposed to GMS1 or compatibility tiles?

Lassebq:
> Can you specify that this script is for GMS2 tile data as opposed to GMS1 or compatibility tiles?

Alright, there you go.

PR #383, opened 2021-03-10
Extension editors, Timeline editor, Spine stuff is now using UndertaleObjects.

nkrapivin:
Works for me in The Swords of Ditto. Testing with other games would be appreciated. (UT BnP? Other fangames?)

Also all classes and types that were nested in UndertaleExtension were moved out as separate types.
This is because WPF hates nested types.
Yes I do hate WPF in return.

Also the timeline editor is pretty straightforward.

(PS: Do I have to add the product id field into the editor? It's not in the UndertaleExtension class but should be easy to fetch.)


Grossley:
Looking good

PR #387, opened 2021-03-14
Initial YYSWF support. Spine V1 support.

nkrapivin:
the title.
I need a hug so badly right now. :(
This stuff makes me very angry *and* sad.


nkrapivin:
Also added Spine V1 support for Bytecode >=15 games. No idea what old games actually use Spine, but here you go.

SWF is still broken and I am still very angry about it.

nkrapivin:
Should be done, tried with Chromaticity and other games by the same author, seems to work.
Still wouldn't mind a hug, I like 'em :/

Issue #394, opened 2021-03-21
Null file in extension editor causes crash


Grossley:

Grossley:
https://user-images.githubusercontent.com/42724744/111920359-b8a60380-8a4b-11eb-9bae-408eb3bce20b.mp4

Grossley:
@nkrapivin Please review this when you are able to

fireahmedgamerYT:
Actually, all I done was removing it not opening it...

nkrapivin:
Please look at #396.

Grossley:
Resolved in #396, closing

PR #396, opened 2021-03-22
Fix issue #394

nkrapivin:
The title.
Same thing is done in many modtool editors, just prevent setting object reference to null :/
But hey, it works.

cc @Grossley


Grossley:
Did you test this and ensure that this resolved #394

nkrapivin:
> Did you test this and ensure that this resolved #394

Well the " X " button is blocked so it's impossible to set the item reference to `null`.
So I guess it is resolved, if you set the reference manually in a script it will of course crash.

PR #402, opened 2021-03-30
Rewrite FancyRoomSelector script

BenjaminUrquhart:
- Made object persistent
- Fixed not being able to use in the same room twice in a row (open, close, then try to open again)
- Remove dependency on variable_global_exists; it's a stub function that always returns true on older runtimes (see Undertale 1.001)
- Remove F3 from obj_time for Undertale to avoid creating system_information_962
- Misc. tweaks to better fit being a persistent object.


BenjaminUrquhart:
Tested on various versions of Undertale and Deltarune. Let me know if this breaks on other games.

Grossley:
Issues addressed in #402 have been fulfilled in 986bb8f

PR #404, opened 2021-04-02
Minor corrections for very old games that use the old audio system.

nkrapivin:
For example Super Crate Box :/


nkrapivin:
image

Grossley:
Merged, thank you

PR #417, opened 2021-04-12
Little decompiler cleanup

Archie-osu:
Cleans up a long chain of if-statements - theoretically should also make it faster, because a switch statement utilizes a jump table.
Also cleans up confusing escaping of escapes using 4 variables, which makes the code a little more readable.


colinator27:
It was pretty outdated to the current decompiler so I just redid the switch statement myself, let me know if I missed anything

PR #429, opened 2021-04-23
Add convenience EventHandlerFor functions

BenjaminUrquhart:
This gives the option to pass an instance of UndertaleData instead of passing in Data.Strings, Data.Code, and Data.CodeLocals explicitly.


BenjaminUrquhart:
Oh also apparently this has a fix for empty SEQN chunk.

I'm bad at git.

PR #439, opened 2021-05-02
Implement deep copy of objects, allow for easy duplication of assets.

Grossley:


Grossley:
Closing until more testing can be done

PR #441, opened 2021-05-05
Dynamically resolve the System Drive

Archie-osu:
Use environment variables instead of hardcoding C:\ - this could lead to the GMS2 Runtime not being found if Windows is installed on D:\ for example.


Grossley:
Looks good to me

PR #472, opened 2021-06-20
IPC with AUMI

Archie-osu:
Fully implemented IPC with AUMI, which can be used by the script interface as well.

AUMI can now receive custom-made data such as bytecode to execute and send back a result, all without any Windows dependencies on UMT's side.

I tested this locally, ran without any errors or exceptions even when AUMI wasn't loaded (the new script TestAUMIConnection.csx can be used to verify if stuff works).

Link to AUMI: https://github.com/Archie-osu/AUMI

### DIFF
--- a/UndertaleModLib/AumiIPC.cs
+++ b/UndertaleModLib/AumiIPC.cs
@@ -1,0 +1,49 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Runtime.InteropServices;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace UndertaleModLib
+{
+    public struct IpcMessage_t
+    {
+        public short FuncID;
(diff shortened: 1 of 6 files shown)


UndertaleModLib/AumiIPC.cs
```diff
@@ -0,0 +1,49 @@
...
+    public struct IpcMessage_t
+    {
+        public short FuncID;
+        [MarshalAs(UnmanagedType.ByValArray, SizeConst = 510)]
+        public byte[] Buffer;
+
+        public byte[] RawBytes()
+        {
+            var stream = new MemoryStream();
+            var writer = new BinaryWriter(stream);
```
colinator27, line 20: I think you might want `using` here, otherwise these might not get cleaned up properly.

Archie-osu: Pushed a fix.

PR #479, opened 2021-06-25
Define function for easier creation of bytecode through a script

Archie-osu:
This commit defines a function that helps with the compilation of GML into bytecode, fixing variables and using AUMI to fix functions as well.

My script using AUMI's functionality to execute arbitrary GML has gone from 100+ lines to just 10 thanks to this.


Archie-osu:
Closing, as this is very old and not even needed.

PR #514, opened 2021-08-11
Code inlining, cleanup

Archie-osu:
Just cleaned up the code a little, all original functionality should be intact - the CheckDecompiler.csx script yields the exact same values before / after the changes.

I don't know much C#, so code criticism is welcome, and I'll gladly update this PR.

### DIFF
--- a/UndertaleModLib/Decompiler/Decompiler.cs
+++ b/UndertaleModLib/Decompiler/Decompiler.cs
@@ -129,10 +129,9 @@ namespace UndertaleModLib.Decompiler
                                 if (ev.CodeId == code)
                                 {
                                     Object = obj;
-                                    goto LoopEnd;
+                                    return;
                                 }
             }
-            LoopEnd: return;
         }
 
         public TempVar NewTempVar()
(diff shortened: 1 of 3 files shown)


UndertaleModLib/Decompiler/Decompiler.cs
```diff
@@ -268,20 +267,16 @@ public bool EqualsNumber(int TestNumber)
...
-                else if (val is float)
+
+                else if (val is double || val is float)
                 {
-                    var v = Convert.ToSingle(val);
-                    int res = (int)v;
-                    if (v == res)
-                        return res;
+                    // Handle float and doubles properly
+                    var Converted = (val is double) ? Convert.ToDouble(val) : Convert.ToSingle(val);
```
colinator27, line 274: I think this might actually result in a performance degradation as compared to the old method, since it not only requires a conversion to double for floats, but it also checks the value's type at least twice, and then casts to int twice (though that is probably easily optimized by the compiler, at least).

UndertaleModLib/Decompiler/Decompiler.cs
```diff
@@ -3127,7 +3122,7 @@ private static Block DetermineSwitchEnd(Block start, Block end, Block meetPoint)
 
                 if (test == end)
                     return end;
-                if (test == meetPoint)
+                else if (test == meetPoint)
```
colinator27, line 3125: This `else` isn't necessary at all.

Miepee:
Is this going to be obsoleted by the new decompiler?

colinator27:
This is pretty much fully obsolete now with the new decompiler, so I'll close this, at least for now.

PR #552, opened 2021-09-19
Add support for GMS2.3+ weird dup variants

krzys-h:
This PR adds support for the weird GMS2.3+ dup variants:

* The two-argument `dup` is actually a weird stack swap thing that takes stack locations in bytes as arguments
* The `dup.v` instruction is a special case for variable duplication
* The old base `dup` code was rewritten to properly take stack element sizes into account

I also commented out some code that was causing a lot of decompilation in DR Chapter 2 to fail... not sure what it was supposed to do because I haven't seen that much GMS2.3 assembly yet

This brings the count of scripts that fail to decompile from 931 to 531

### DIFF
--- a/UndertaleModLib/Decompiler/Decompiler.cs
+++ b/UndertaleModLib/Decompiler/Decompiler.cs
@@ -1480,6 +1480,32 @@ namespace UndertaleModLib.Decompiler
             }
         }
 
+        static int GetTypeSize(UndertaleInstruction.DataType type)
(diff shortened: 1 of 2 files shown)


colinator27 (approved):
Definitely should improve output, was planning to remove that builtin variable code soon too

BenjaminUrquhart (approved):
GitHub lets me do this. Hi.

PR #561, opened 2021-09-21
Add preliminary GMS2.3 method support

krzys-h:
**EXPERIMENTAL CODE AHEAD**

This adds preliminary supports for the GMS2.3 methods.

* Methods defined using inline functions will now properly inline the body of the function
* Calls to child code objects that are actually method definitions will get resolved to the method name (no more `gml_Script_...`)
* Recursive asset resolution *should* now work with child code blocks (not fully tested)

Needs more testing to make sure it doesn't break any existing decompilation, because I had to mess with how the decompiler generates entry points quite a bit. I'm especially worried about the consequences of removing the unreachable code detector that was apparently fixing the loop detection sometimes.

### DIFF
--- a/UndertaleModLib/Compiler/AssemblyWriter.cs
+++ b/UndertaleModLib/Compiler/AssemblyWriter.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Globalization;
 using System.Linq;
 using System.Text;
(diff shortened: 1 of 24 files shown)


krzys-h:
Support for local vars also needs to be double-checked (I'm not sure which code locals object it's using now)

krzys-h:
I think this is ready for testing and merging now. I already make sure that these changes don't alter the decompilation output for Undertale and Deltarune Chapter 1 at all, and only change the method stuff in Chapter 2.

Also, `ExportAllCode` works now for GMS2.3. It never exports the anonymous methods by themselves, so we should not run into any filename length problems.

krzys-h:
The compiler is now more broken than it was before because of the short names, this has to be fixed before merging

Archie-osu:
What's the current status on this?

Issue #590, opened 2021-09-25
Int64 decompiler System.NotImplementedException exception


LukeSaward1:
![image](https://user-images.githubusercontent.com/55454721/134766687-de808044-88af-4777-a15e-dad884467c68.png)
I get the above error when putting this code into gml_Object_DEVICE_MENU_Step_0 code entry:
```  
global.isTimerRunning = 1
if (global.isTimerRunning == 1)
{
    global.timeStart = get_timer()
}
```
using the timer code from the Undertale RTA timer mod.

Is there any reason why this would happen?

Archie-osu:
It may actually be a compiler / decompiler issue, as GMS 2.3 is not yet fully supported - at this point in time, it is recommended to edit assembly, and not decompiled code.

LukeSaward1:
I wish I knew assembly, I guess I need to learn it if I want to mod DRC2 rn.
Kinda would have been better if toby went with the version of gms before 2.3, although I don't know if that would have made it easier to mod; as I don't know if the exceptions are caused by GMS2.3 or if they would also appear on GMS2 versions before 2.3

Archie-osu:
I'll try implementing a potential fix myself - where exactly inside gml_Object_DEVICE_MENU_Step_0 have you added the supplied code?

Archie-osu:
![image](https://user-images.githubusercontent.com/61205459/134767090-11eadc8e-4b16-475b-a318-1cf29ce0dba6.png)
Was it like this, at the top?

LukeSaward1:
It was at the bottom in the
```
if(START_GAME == 1)
{
}
```
block I think

Grossley:
Fixed in 62fe347a0017476505f5e0415c2b5f549f690a17

PR #591, opened 2021-09-25
Add Decompiler typesizes for Double and Int64

Archie-osu:
Solves #590 - tried putting in the mentioned code, the game ran just file and UMT's decompiler worked as well.


colinator27 (approved):
Should fix "dup.l 0" instructions

PR #608, opened 2021-09-30
Unbreak compiler

Grossley:
The compiler takes an extremely long time (probably 20x-100x slower than normal at the minimum) due to linking up simplified names with internal global script entries. The simplification of the names is unnecessary where the unsimplified names produces accurate disassembly of about the same quality. Therefore by removing the unsimplified names aspect it is possible to remove the (and I am stressing this for emphasis) ***extremely slow (practically making the compiler unusable)*** linking code. 

It is recommended to either unsimplify the names (as is crudely done in this pull request) or to find a proper remedy to make the compiler usable, as in its current state it is not adequate for release of any kind, not now nor in the future. 

Edited to use the minimum number of changes to achieve this.


Grossley:
This may be obsoleted by 41f8ae60ceed8df1c4a900702e7316b42f4822b1, feel free to close it if this is the case.

PR #640, opened 2021-10-14
Non-blocking UI script results

VladiStep:


Archie-osu:
Why are you not utilize the "using" keyword in the main window code? Would simplify the Dispose calls.

PR #644, opened 2021-10-16
Limiting "Scripts" menu while "data.win" is not loaded.

VladiStep:
Keep disabled scripts that use "_data.win_", until it will be loaded.


nkrapivin:
What if I want to run a script that doesn't require a data file (e.g. backup)?

VladiStep:
I think this pull request is not required, so I close it.

PR #651, opened 2021-10-17
Update ExportAllCode2_3.csx

maximmax42:
Swapped two lines of code around so that the LUT file actually starts to list the code names at line 2 (currently it does so from line 1 right after the text explaining it's at line 2).


Grossley:
Thank you. Merged.

PR #652, opened 2021-10-18
cli: fix publish by disabling single file build

gurgalex:
Not ideal, but no more crashes about unable to find CodeAnalysis
assembly locations.

Not sure what is different from the GUI, but we can probably figure that out later.

image


gurgalex:
I should probably figure out how to exclude the different languages for code analysis.
image

gurgalex:
localizations for code analysis removed in #653

PR #672, opened 2021-10-28
Simplify flow graph generation

### LINKED ISSUE #671: NullReferenceException when decompiling a do-until loop
I compiled the following script:
```
var value = 0;
do {
    value = irandom(100);
} until (value < 50);
```
This is the resulting assembly:
```
.localvar 0 arguments
.localvar 1 value 1

:[0]
pushi.e 0
pop.v.i local.value

:[1]
pushi.e 100
conv.i.v
call.i irandom(argc=1)
pop.v.v local.value
pushloc.v local.value
pushi.e 50
cmp.i.v LT
bf [1]

:[end]
```
After that, I tried to decompile it back into a script, but got an error:
```
/* EXCEPTION!
   System.NullReferenceException: Object reference not set to an instance of an object.
   at UndertaleModLib.Decompiler.Decompiler.ComputeReverseDominators(Dictionary`2 blocks, Block entryBlock) in D:\kod\UndertaleModTool\UndertaleModLib\Decompiler\Decompiler.cs:line 2978
   at UndertaleModLib.Decompiler.Decompiler.HLDecompile(DecompileContext context, Dictionary`2 blocks, Block entryPoint, Block rootExitPoint) in D:\kod\UndertaleModTool\UndertaleModLib\Decompiler\Decompiler.cs:line 3318
   at UndertaleModLib.Decompiler.Decompiler.Decompile(UndertaleCode code, GlobalDecompileContext globalContext) in D:\kod\UndertaleModTool\UndertaleModLib\Decompiler\Decompiler.cs:line 3432
   at UndertaleModTool.UndertaleCodeEditor.<>c__DisplayClass26_1.<DecompileCode>b__1() in D:\kod\UndertaleModTool\UndertaleModTool\Editors\UndertaleCodeEditor.xaml.cs:line 344
*/
```

I suspect there's a problem with flow graph generation. Here's a screenshot from debugger:
obraz

mwiatrowski:
This is an attempt to fix #671.

### DIFF
--- a/UndertaleModLib/Decompiler/Decompiler.cs
+++ b/UndertaleModLib/Decompiler/Decompiler.cs
@@ -2315,134 +2315,118 @@ namespace UndertaleModLib.Decompiler
 
         public static Dictionary<uint, Block> DecompileFlowGraph(UndertaleCode code, List<uint> entryPoints)
         {
-            Dictionary<uint, Block> blockByAddress = new Dictionary<uint, Block>();
-            foreach(uint entryPoint in entryPoints)
-                blockByAddress[entryPoint] = new Block(entryPoint);
-            Block entryBlock = new Block(null);
-            Block finalBlock = new Block(code.Length / 4);
-            blockByAddress[code.Length / 4] = finalBlock;
-            Block currentBlock = entryBlock;
+            uint finalBlockAddress = code.Length / 4;
 
-            foreach (var instr in code.Instructions)
+            Dictionary<uint, Block> blocksByAddress = new Dictionary<uint, Block>();
+
+            // Identify addresses where any block starts
+            var blockStartingPoints = new List<uint>();
+            foreach (uint entryPoint in entryPoints) {
+                blockStartingPoints.Add(entryPoint);
+            }
+            blockStartingPoints.Add(finalBlockAddress);
+            foreach (var instruction in code.Instructions) {
+                switch (instruction.Kind) {
+                case UndertaleInstruction.Opcode.B:
+                case UndertaleInstruction.Opcode.Bt:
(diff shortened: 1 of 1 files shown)


mwiatrowski:
A couple of observations:
* My version has a minor bug where entries in `Block::entryPoints` sometimes get duplicated. I fixed it, but it didn't change the output.
* Original algorithm has a quirk where it sometimes adds an uninitialized block to `entryPoints` of some other block (I think this is what is causing #671). Additionally, `entryPoints` isn't sorted. I made a change to the original code (on local master branch) to remove bad block and sort the list, but the decompilation output didn't change either.
* Dictionary of blocks generated by the original code isn't sorted. I changed it so that they would be and this time output exactly matched what my branch produces.

It appears that type resolution depends on the order in which blocks are processed. In particular, `cityscape_alpha = 0,15` happened because `ExpressionConstant` was incorrectly recognized as some other type than `Float`/`Double` and formatted with basic `ToString()` method instead of the dedicated function.

@Grossley I think this PR can't be merged until the issue with type resolution is fixed.

Grossley:
What is the current status of this pull request now that #674 has been implemented?

colinator27:
If #674 is working then this is probably good to merge! We should test accuracy using the scripts on some games first, though, just to be sure.

Miepee:
What's the status of this PR? Info is a bit over the place here, are there any current known caveats / differences in generated code?

Grossley:
What is the status of this pull request @mwiatrowski @colinator27 ?

Miepee:
Is this going to be obsoleted by the new decompiler?

colinator27:
This has indeed been obsoleted by the new decompiler.

PR #685, opened 2021-11-11
Added scripts to import and export rooms and game objects

SolventMercury:


colinator27:
The only real issue I have with this is that code formatting is inconsistent with most of the project, but since these are just scripts that's not a very big problem. I haven't actually tested the scripts, but they do seem pretty useful!

SolventMercury:
Closing since y'all have not reviewed this in ages.

Miepee:
That's normal for larger software projects. Things get missed, or forgotten about.
If you want to have this feature implemented in UMT, a simple "any news on this?" comment would be more helpful than just closing.

Miepee:
Also not sure if you're still interested in this @SolventMercury but I found two issues:
- The scripts are actually not bundled. You have to select the files in VS and change the "copy to output directory" attribute to "copy if newer"
- Importing rooms doesn't quite work correctly. If one is trying to import any room (even if it's a just freshly exported room without any changes) something gets messed up somewhere, as when trying to save the data.win one is greeted with this lovely error:
image

SolventMercury:
> Also not sure if you're still interested in this @SolventMercury but I found two issues:
> 
>     * The scripts are actually not bundled. You have to select the files in VS and change the "copy to output directory" attribute to "copy if newer"
> 
>     * Importing doesn't quite work correctly. If one is trying to import any room or object (even if it's a just freshly exported without any changes) something gets messed up somewhere, as when trying to save the data.win one is greeted with this lovely error:
>       image

I actually fixed the latter issue a while back. I might have forgotten to apply it to the version in my repo, I'll check tomorrow. The former tip I was unaware of. I'm better at writing code than fiddling with the IDE, so I appreciate the heads-up.

PR #688, opened 2021-11-13
Script-related changes.

VladiStep:
1. Added sound for `ScriptQuestion()` and `ClickableTextOutput()` _(code search results window)_.
2. If code search results are not fully displayed, there will be notification - _"Displaying the results, please wait..."_.
3. _Ctrl+C_ copies selected code search results as _plain_ text.

### DIFF
--- a/UndertaleModTool/MainWindow.xaml.cs
+++ b/UndertaleModTool/MainWindow.xaml.cs
@@ -1356,7 +1356,7 @@ namespace UndertaleModTool
 
                 //TODO: find the way to scroll to the code item and highlight it.
                 
-                CodeEditorDecompile = editorDecompile; //-1 - unstated, 0 - don't decompile, 1 - decompile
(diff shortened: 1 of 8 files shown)


UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -1349,7 +1349,7 @@ public void OpenCodeFile(string name, sbyte editorDecompile)
 
                 //TODO: find the way to scroll to the code item and highlight it.
                 
-                CodeEditorDecompile = editorDecompile;
+                CodeEditorDecompile = editorDecompile; //-1 - unstated, 0 - don't decompile, 1 - decompile
```
colinator27, line 1352: Not sure about the full context here, but why isn't this an enum?

VladiStep: I just didn't think about it.

PR #700, opened 2021-11-26
Fixes of the code import scripts and minor improvements.

VladiStep:
1. Fixed the following scripts _(by adding "Functions" and "Variables" to `SyncBinding())`_ : _"ImportASM.csx"_, _"ImportASM_2_3.csx"_, _"ImportGML.csx"_, _"ImportGML_2_3.csx"_, _"ExportAndConvert_2_3_ASM.csx"_ and _"FindAndReplace.csx"_.
2. Replaced `Data.Code.ByName()` calls to the `code` variable reference where it's necessary _(for a slight performance boost)_.
3. `ReplaceTextInGML()` now throws `ScriptException` instead of calling `ScriptError()`. So, any script that would use that method will stop if there is no code entry with the requested name.
4. `CompileGMLText()` now also writes the code name on error.
5. Fixed bug, when clicked code search entry was overwritten by previously edited and _not compiled_ code.
_(This also should fix similar behavior of UMT with tabs, when you switch to another code tab without compiling)_
6. Fixed display of code editor compile error.
_(Sometimes, it was shown only after closing the "Compiling, please wait..." window)_

_(IDK why it retrieved that old commits, actual first commit of this PR is **"Some script fixes and minor improvements"**)_


Archie-osu:
This is the 700th pull request, congratulations!

PR #730, opened 2022-01-22
Improved/fixed update system.

VladiStep:
1. Download progress bar displays how many MB are downloaded.
2. Works on every build version _(non-single, .NET-bundled, etc.)_.
3. Shows the _"Check your internet connection"_ error.
4. Auto-detects current build version and OS type _(32 or 64-bit)_.
5. Checks if there is enough space on the system drive _(at least 500 MB is needed)_.
6. Detects if update is needed or not _(on the button click)_.
7. Update check button is aligned properly.
8. If it's running on Windows 7 and an SSL connection error is occured, it offers you to open a specific Windows update page to fix the issue with that _(KB2992611)_.
9. Update button is disabled while updating.
10. Updater app console has title now.
11. Updater app console waits for key press on the error.

Other minor changes:
1. `SetProgressBar()` without arguments just shows the progress bar.
2. You can specify `ShowQuestion()` icon _(error, warning, etc.)_.


NyakoFox:
LGTM

PR #733, opened 2022-01-27
add new community script

Trylobot:
doing some modding work for another game maker game community, thought I'd pop this over for inclusion in the community scripts; it's been tested with Samurai Gunn 2 version EA4_2


Grossley:
This seems to do the same thing as the script that is already bundled with UndertaleModTool in "Repackers" called "ImportAllEmbeddedTextures.csx". Does this do something unique that said script does not do? Also, the art is nice, but it is really big, taking up a whole 100 lines of the 161 line script.

Grossley:
This seems to do the same thing as the script that is already bundled with UndertaleModTool in "Repackers" called "ImportAllEmbeddedTextures.csx". Does this do something unique that said script does not do? Also, the art is nice, but it is really big, taking up a whole 100 lines of the 161 line script. @Trylobot

PR #738, opened 2022-02-01
Add a .yyp converter Experimental

MetalManZ:
Adding a .yyp converter would expand modding abilities, since the basic GameMaker Studios 2 is free. It would also possibly help you create more modding features for GMS 2.


BenjaminUrquhart:
For legal reasons UTMT will never have a Gamemaker Project exporter. Also, in the future, please make suggestions via the Issues section instead of a dud Pull Request. 

That being said, you can do most things GMS 2.2.5 and lower can do directly in the modding tool (2.3 and later is a work in progress).

PR #777, opened 2022-03-04
UMT Tabs

SolventMercury:
- Added tabs to UMT, allowing users to open multiple assets at once and switch between them
- Added export and import scripts for rooms and game objects


Grossley:
> Added export and import scripts for rooms and game objects

This should be in a separate pull request.

Miepee:
Was already posted this in the other PR, but wanted to post it in here as well:
The importing scripts currently don't work. The reason for those is, that new UndertaleStrings are added, but not added to the Data itself.
Also, formatting seems very inconsistent with the rest of the project.

Miepee:
Was closed in favour of #824 and #815.

PR #793, opened 2022-03-17
Fix v2022.2, part 2

Jacky720:
I won't call this ready, it is highly untested, but putting it out there. We've officially got v2022.2 detection and storage of the mystery new font value.


Jacky720:
Credit to @mandar1jn and @cgytrus for initial implementation of this fix, and sorry for being so trigger-happy with it.

mandar1jn:
@Jacky720 I think you did my coauthor property wrong

PR #803, opened 2022-03-26
Add struct decompilation

BenjaminUrquhart:
This is a mess, if someone else could figure out how to improve it please do so.

I also don't know what impact this has on decompiling other entries but from what few things I've tried, nothing seems broken (except possibly some asset resolution on struct member vars)


colinator27:
Looks good - did some testing, and it seems to work great. If anything major is broken, we can fix it later, although I can't see what would break at the moment.

PR #806, opened 2022-03-26
First attempt at adding in-code documentation

Miepee:
This is the first batch / attempt of adding in-code documentation.
The following has been annotated so far:
- UndertaleAnimationCurves.cs (except for channels)
- UndertaleBackgrounds.cs (except for a few GMS2 specific things)
- A bit of UndertaleCode.cs
- UndertaleAudio.cs
- UndertaleEmbeddedTexture.cs (except for the blob serializers)
- UndertaleFont.cs (except for a few attributes)
- UndertaleGameObject.cs (except for a few methods 
- Most of UndertaleGeneralInfo.cs
- UndertaleGlobalInit.cs
- Some of UndertaleRoom.cs
- UndertaleScript.cs
- Most of UndertaleShader.cs
- UndertaleSound.cs
- UndertaleSprite.cs except for most of the GMS2 stuff and padding methods.
- UndertaleString.cs
- UndertaleTextureGroupInfo
- UndertaleTexturePageItem
- UndertaleTimeline.cs
- Most of UndertaleData.cs
- A bit of UndertaleBaseTypes
- Most of IScriptingInterface.cs

In addition, I also did the following changed to IScriptingInterface:
- rename `Make_New_File` to `MakeNewFile`. This may break a few Scripts, I will check later if any are broken.
- Put the implementation of `SendAUMIMessage` right into the scripting Interface.
- Put the implementation `AreFilesIdentical` right into the scripting interface.
- Rename `IncProgress()` `AddProgressP` and `IncProgressP` to `IncrementProgress`, `AddProgressParallel` and `IncrementProgressParallel` respectively. This likely breaks a few scripts,  but I do not know which ones right now.
- rename `SyncBinding(bool enable = false);` to `DisableAllBindings()`
- rename `StartUpdater` and `StopUpdater` to `StartProgressBarUpdater` and `StopProgressBarUpdater` respectively. This likely breaks a few scripts, but I do not know which ones right now.
- Put the implementation of the dummy methods right into the scripting interface.

In addition, I also added a lot of TODOs, which mostly indicate things that should still be documented before I put this PR out of Draft, but also some other structural changes that should be done.

Lastly, I also enabled it, so every public field which is undocumented shows up as a warning now.

Not related to the codebase, but somewhat to this PR. I put up a quick basic doxygen HTML render of the current documentation here, which may or may not help with reviewing:
https://miepee.github.io/UMTDoc/annotated.html

### DIFF
--- a/UndertaleModLib/Scripting/IScriptInterface.cs
+++ b/UndertaleModLib/Scripting/IScriptInterface.cs
@@ -1,5 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.IO;
+using System.IO.Pipes;
 using System.Linq;
 using System.Text;
 using System.Threading;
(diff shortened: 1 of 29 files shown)


Miepee:
I've fixed the big issues, so it's ready for review now.
There are still a few undocumented things in the models, but am not that knowledgeable on them.

UndertaleModLib/Scripting/IScriptInterface.cs
```diff
@@ -22,84 +24,469 @@ public ScriptException(string msg) : base(msg)
...
+        /// </summary>
         void HideProgressBar();
+
+        /// <summary>
+        /// Enables the UI.
+        /// </summary>
         void EnableUI();
+
+        /// <summary>
+        /// Allows scripts to synchronize their assets with the lists from the UI.
```
VladiStep, line 405: _"Allows scripts to modify asset lists from non-UI thread _(e.g. from within `Task.Run()`)_."_ would be better, because it describes better what it actually used for.

VladiStep: Or maybe _"Calls `BindingOperations.EnableCollectionSynchronization()` for every listed asset list.
It allows scripts to modify asset lists from non-UI thread _(e.g. from within `Task.Run()`)_."_

UndertaleModLib/Scripting/IScriptInterface.cs
```diff
@@ -22,84 +24,469 @@ public ScriptException(string msg) : base(msg)
...
+        /// <param name="enable">Whether to enable or disable the synchronization.</param>
+        //TODO: Having resourceType as a comma separated list just screams for error. Make it use some array of predefined assets it can use.
         void SyncBinding(string resourceType, bool enable);
-        void SyncBinding(bool enable = false);
-        void StartUpdater();
-        Task StopUpdater();
 
-        Task<bool> GenerateGMLCache(ThreadLocal<Decompiler.GlobalDecompileContext> decompileContext = null, object dialog = null, bool isSaving = false);
+        /// <summary>
+        /// Stops the synchronization of all previously enabled assets.
```
VladiStep, line 413: _"Calls `BindingOperations.DisableCollectionSynchronization()` for all the asset lists that have enabled synchronization."_

VladiStep (changes requested):

UndertaleModLib/Scripting/IScriptInterface.cs
Miepee: Disagree with this.
Different GUIs / implementations could handle this differently, there is no guarantee that this exact method will be called.

UndertaleModLib/Scripting/IScriptInterface.cs
Miepee: Disagree with the second comment, see description below as to why.

Regarding the first one, are they really modifying the asset list in the non-UI thread tho? As far as I can see they just modify their own assets which then get synced to the UI list.
In ImportGML.csx for example:
```cs
SyncBinding("Strings, Code, CodeLocals, Scripts, GlobalInitScripts, GameObjects, Functions, Variables", true);
await Task.Run(() => {
    foreach (string file in dirFiles)
    {
        IncrementProgress();

        ImportGMLFile(file, doParse, false, stopOnError);
    }
});
DisableAllSyncBindings();
```
Nothing actually accesses the list from the UI. `ImportGMLFile` changes assets inside of `Data` which then get synced with the UI list later because `SyncBinding` was called before.

VladiStep: Most of asset lists in `Data` _(code, sprites, etc.)_ always bound to the _"Data"_ tree list.
And `ImportGMLFile()` modifies some of them while being within `Task.Run()` _(non-UI thread)_.
So, without `SyncBinding()`, it throws an error.

UndertaleModLib/Scripting/IScriptInterface.cs
VladiStep: There is no other method that does that _(at least for now)_.

UndertaleModLib/Scripting/IScriptInterface.cs
Miepee: Ah was not aware it throws an error otherwise.
Is 
`Allows scripts to modify asset lists from the non-UI thread. If this isn't called before attempting to modify them, an exception will be thrown.` 
fine?

VladiStep: Yes, but _"... `System.NotSupportedException` will be thrown."_

PR #815, opened 2022-03-29
Scripts for importing/exporting objects and rooms

Miepee:
This adds scripts for importing/exporting objects and rooms.
These scripts were part of #777, I've excluded them from there and cleaned them up a bit.

### DIFF
--- a/UndertaleModTool/Repackers/ImportGameObject.csx
+++ b/UndertaleModTool/Repackers/ImportGameObject.csx
@@ -195,6 +195,9 @@ UndertaleGameObject.EventAction ReadAction(ref Utf8JsonReader reader) {
 		newAction.ActionName  = null;
 	} else {
 		newAction.ActionName  = new UndertaleString(actionName);
+
+		if (!Data.Strings.ToList().Any(s => s == newAction.ActionName))
+			Data.Strings.Add(newAction.ActionName);
 	}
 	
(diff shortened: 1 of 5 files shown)


wertercatt:
I wish there was a script to fully unpack and repack data.win.

Miepee:
This is kinda the wrong place for this.
I personally don't think it's that useful. But all scripts are already in place, so it's just a matter of writing a script that calls all other scripts.
Feel free to do it yourself, it's a great learning experience!

NyakoFox:
I'm curious if this has been looked at yet?

wapordev:
Im getting bugs when using this with Samurai Gunn 2. (It's not free, but contact me if you need to see the .win or anything)
To test it, I just modified a tile in a room, exported, and reimported in to a fresh copy. I also made sure this worked with undertale so it wasn't an issue with my setup.
It gets through exporting just fine, but has trouble when saving after reimporting. I get this error:
image

Miepee:
> using this with Samurai Gunn 2

Which GMS version does it use? (for games gms2+ and higher, you'd probably be better using this snippet: https://gist.github.com/JohnnyonFlame/fdd79303423392c5911cf44eed9e17cb)

Miepee:
Thanks, will need to double check if there has been something new added by those versions that isn't tracked.

Issue #467, opened 2021-06-05
UndertaleModTool cannot load and save an unmodified copy of "Death's Gambit v1.2"


Grossley:
The problem seems to be in "room_darknessfalls_extra" (room 338) which has a massive amount of null instances in "Compatibility_Instances_Depth_0". This makes me believe that the tool is having trouble actually loading the room properly since this data is unexpected and obviously wrong. 
![image](https://user-images.githubusercontent.com/42724744/120902329-9fe1ce80-c5f4-11eb-8320-aca4c641c9eb.png)

Miepee:
@JohnnyonFlame can you retest please if this has been fixed?

JohnnyonFlame:
@Miepee File now saves successfully, testing it required `git checkout 7094c547 -- UndertaleModLib/UndertaleChunks.cs` since breakage from a73a3a8d is still present.

Miepee:
But since that's a duplicate of #833, this can be closed, right?

JohnnyonFlame:
Guess so.

PR #848, opened 2022-04-20
Null instance handling in deserialization

Jacky720:
Hopefully resolves all these "object reference not set to an instance of an object" issues in saving. Might crash in loading if there are two consecutive null objects (if that's even possible) and will not re-write the objects 1:1.

Also, change a couple of if-else chains to switches.


Miepee:
@JohnnyonFlame can you check if this fixes #467 ?

JohnnyonFlame:
I can, but I have no idea how to guarantee the correctness of this solution.

PR #849, opened 2022-04-20
Added the "Frame index" and "Image speed" fields.

### LINKED ISSUE #834: Add frame and speed parameters to object instances for GMS 2.3 games
Add frame and speed parameters to object instances for GMS 2.3 games, these are missing from the room editor.

VladiStep:
1. Added _"Frame index"_ and _"Image speed"_ fields for game object instances in the room editor.
2. Fixed crash when changing frame index of sprite instance with `null` sprite.
3. `UndertaleRoomRenderer` _(used in "ExportAllRoomsToPng.csx")_ now takes into account frame index of sprite instances _(+ object instances)_.
4. Added _"DataFieldOneTimeConverter"_ for one-time binding to `bool` fields.

Closes #834

### DIFF
--- a/UndertaleModLib/Models/UndertaleRoom.cs
+++ b/UndertaleModLib/Models/UndertaleRoom.cs
@@ -699,7 +699,7 @@ namespace UndertaleModLib.Models
             /// </summary>
             public int ImageIndex { get; set; }
             /// <summary>
-            /// A wrapper of <see cref="ImageIndex"/> that prevents entering index that's out of bounds.
+            /// A wrapper of <see cref="ImageIndex"/> that prevents using an out of bounds index.
             /// </summary>
             public int SafeImageIndex { get => ImageIndex; set { ImageIndex = Math.Clamp(value, 0, (ObjectDefinition?.Sprite?.Textures?.Count ?? 1) - 1); OnPropertyChanged(); } }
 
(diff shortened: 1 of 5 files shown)


UndertaleModLib/Models/UndertaleRoom.cs
```diff
@@ -698,6 +698,10 @@ public class GameObject : UndertaleObjectLenCheck, RoomObject, INotifyPropertyCh
             /// The image index of this object. Game Maker: Studio 2 only.
             /// </summary>
             public int ImageIndex { get; set; }
+            /// <summary>
+            /// A wrapper of <see cref="ImageIndex"/> that prevents entering index that's out of bounds.
```
Miepee, line 702:
```suggestion
            /// A wrapper of <see cref="ImageIndex"/> that prevents using an out of bounds index.
```

UndertaleModTool/Converters/UndertaleCachedImageLoader.cs
```diff
@@ -300,7 +300,12 @@ public object Convert(object[] values, Type targetType, object parameter, Cultur
             if (textures is null)
                 return null;
 
-            int index = (int)(float)values[1];
+            int index = -1;
+            if (values[1] is int indexInt)
+                index = indexInt;
+            else if (values[1] is float indexFloat)
+                index = (int)indexFloat;
+
```
Miepee, line 308: Can't this be replaced with just
```cs
int index = (int)values[1];
```
?

VladiStep: `values[1]` is `object`, so `(int)values[1]` here is _"unboxing"_, not a type conversion.
So no, it can't be replaced.

PR #888, opened 2022-05-05
More room editor changes.

### LINKED ISSUE #880: Bug with zooming in/out in the room editor.
If you click on some room object and zoom in/out, then you can't release that object

https://user-images.githubusercontent.com/45438625/166121039-ecc353ac-58cb-4016-985a-933b02d9cc0f.mp4

.

VladiStep:
1. **`UndertaleBackground` and `UndertaleSprite` items can be drag-and-dropped to the room _(`UndertaleSprite` will work only on GMS 2 games)_.**
2. Fixed persistent room editor vertical scroll bar.
3. **Fixed bug that caused errors on game saving if there are some `SpriteInstance` copies with auto-generated name _("graphic_XXXXXXXX")_.**
4. Added `SpriteInstance.GenerateRandomName()`.
5. Added `insertIndex` argument to `AddObjectCopy()` _(if defined, it inserts an object to the list)_.
6. Improved `Rectangle_MouseDown()` so it uses `AddObjectCopy()`.
7. **Fixed room object moving while zooming in/out.**
8. **Following layer properties are now properly working/displayed: depth, offset, visibility.**
9. **If layer is selected, all its objects also will flash.**
10. New layer is visible by default.
11. **Fixed adding and removing of layers.**
12. Fixed saving game data with newly created asset layers _(`AssetsData.NineSlices` was null)_.
13. Depth of new layer automatically increases.
14. `VisualUpwardSearch()` and `FindNearestParent()` methods of `MainWindow` are `public`.
15. Added `Layer.ToString()` override for ease of debugging.
16. Fixed background layer tiling with offset.
17. New layer name is increasing: _"name" -> "name1", "name9" -> "name10"_.
18. **If layers of opened room are not depth-ordered, it will be suggested to rearrange them.**
19. Removed redundant `BGLayerVisConverter`.
20. Following new background layer data properties are properly set: _visibility, color_.
21. Removed redundant _"flashAnimation"_ resource from the room editor.
22. Flashing animation error message is more precise.
23. Fixed potentional crash on layer adding/removing.
24. Applied WPF error _("Cannot find governing FrameworkElement or FrameworkContentElement for target element.")_ fix to `UndertaleRoomRenderer`.
25. Fixed wrong offset display of tile layer with a lot non-transparent tiles _(more than 1000)_ and size larger than room _(`TileLayerImage` stretch mode wasn't specified)_.

---
Closes #880.
Here's the link to final builds for testing: [artifacts](https://github.com/VladiStep/UndertaleModTool/actions/runs/2303709378#artifacts)

### DIFF
--- a/UndertaleModLib/Models/UndertaleRoom.cs
+++ b/UndertaleModLib/Models/UndertaleRoom.cs
@@ -1527,10 +1527,10 @@ namespace UndertaleModLib.Models
                 Rotation = reader.ReadSingle();
             }
 
-            public static UndertaleString GenerateRandomName()
+            public static UndertaleString GenerateRandomName(UndertaleData data)
             {
                 // The same format as in "GameMaker Studio: 2".
-                return new UndertaleString("graphic_" + ((uint)new Random().Next(-int.MaxValue, int.MaxValue)).ToString("X8"));
+                return data.Strings.MakeString("graphic_" + ((uint)new Random().Next(-int.MaxValue, int.MaxValue)).ToString("X8"));
             } 
 
             public override string ToString()
--- a/UndertaleModLib/Models/UndertaleRoom.cs
+++ b/UndertaleModLib/Models/UndertaleRoom.cs
@@ -1058,7 +1058,7 @@ namespace UndertaleModLib.Models
             /// <summary>
             /// The room this layer belongs to.
             /// </summary>
-            public UndertaleRoom ParentRoom { get => _ParentRoom; set { _ParentRoom = value; PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(nameof(ParentRoom))); UpdateParentRoom(); } }
+            public UndertaleRoom ParentRoom { get => _ParentRoom; set { _ParentRoom = value; OnPropertyChanged(); UpdateParentRoom(); } }
 
             /// <summary>
             /// The name of the layer.
@@ -1084,7 +1084,7 @@ namespace UndertaleModLib.Models
             public float YOffset { get; set; }
             public float HSpeed { get; set; }
             public float VSpeed { get; set; }
-            public bool IsVisible { get; set; }
+            public bool IsVisible { get; set; } = true;
             public LayerData Data { get; set; }
             public LayerInstancesData InstancesData => Data as LayerInstancesData;
             public LayerTilesData TilesData => Data as LayerTilesData;
@@ -1093,6 +1093,10 @@ namespace UndertaleModLib.Models
             public LayerEffectData EffectData => Data as LayerEffectData;
 
             public event PropertyChangedEventHandler PropertyChanged;
+            protected void OnPropertyChanged([CallerMemberName] string name = null)
+            {
+                PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(name));
+            }
 
             // GMS 2022.1+
             public bool EffectEnabled { get; set; }
@@ -1106,6 +1110,7 @@ namespace UndertaleModLib.Models
                 if (TilesData != null)
                     TilesData.ParentLayer = this;
             }
+            public void UpdateZIndex() => OnPropertyChanged("LayerDepth");
 
             public void Serialize(UndertaleWriter writer)
             {
@@ -1165,6 +1170,11 @@ namespace UndertaleModLib.Models
                     case LayerType.Effect: Data = reader.ReadUndertaleObject<LayerEffectData>(); break;
                     default: throw new Exception("Unsupported layer type " + LayerType);
                 }
+            }
+
+            public override string ToString()
+            {
+                return GetType().FullName + " - \"" + LayerName?.Content + '\"';
             }
 
             public class LayerInstancesData : LayerData
@@ -1355,7 +1365,7 @@ namespace UndertaleModLib.Models
                 public UndertalePointerList<Tile> LegacyTiles { get; set; }
                 public UndertalePointerList<SpriteInstance> Sprites { get; set; }
                 public UndertalePointerList<SequenceInstance> Sequences { get; set; }
-                public UndertalePointerList<SpriteInstance> NineSlices { get; set; } // Removed in 2.3.2, before ever used
+                public UndertalePointerList<SpriteInstance> NineSlices { get; set; } // Removed in 2.3.2, before never used
 
                 public void Serialize(UndertaleWriter writer)
                 {
--- a/UndertaleModLib/Models/UndertaleRoom.cs
+++ b/UndertaleModLib/Models/UndertaleRoom.cs
@@ -1527,6 +1527,12 @@ namespace UndertaleModLib.Models
                 Rotation = reader.ReadSingle();
             }
 
+            public static UndertaleString GenerateRandomName()
+            {
+                // The same format as in "GameMaker Studio: 2".
+                return new UndertaleString("graphic_" + ((uint)new Random().Next(-int.MaxValue, int.MaxValue)).ToString("X8"));
+            } 
+
             public override string ToString()
             {
                 return "Sprite " + Name?.Content + " of " + (Sprite?.Name?.Content ?? "?") + " (UndertaleRoom+SpriteInstance)";
--- a/UndertaleModLib/Models/UndertaleRoom.cs
+++ b/UndertaleModLib/Models/UndertaleRoom.cs
@@ -167,6 +167,23 @@ namespace UndertaleModLib.Models
         public UndertaleSimpleList<UndertaleResourceById<UndertaleSequence, UndertaleChunkSEQN>> Sequences { get; private set; } = new UndertaleSimpleList<UndertaleResourceById<UndertaleSequence, UndertaleChunkSEQN>>();
 
         public void UpdateBGColorLayer() => OnPropertyChanged("BGColorLayer");
+        public void RearrangeLayers(Layer selectedLayer = null)
+        {
+            if (Layers.Count > 0)
+            {
+                Layer[] orderedLayers = Layers.OrderBy(l => l.LayerDepth).ToArray();
+
+                // ensure that room objects tree will have the layer to re-select
+                if (selectedLayer is not null)
+                    Layers[Array.IndexOf(orderedLayers, selectedLayer)] = selectedLayer;
+
+                for (int i = 0; i < orderedLayers.Length; i++)
+                {
+                    if (Layers[i] != orderedLayers[i])
+                        Layers[i] = orderedLayers[i];
+                }
+            }
+        }
 
         /// <summary>
         /// The layer containing the background color.<br/>
@@ -1297,13 +1314,13 @@ namespace UndertaleModLib.Models
                 public float CalcScaleX { get; set; }
                 public float CalcScaleY { get; set; }
 
-                public bool Visible { get; set; }
+                public bool Visible { get; set; } = true;
                 public bool Foreground { get; set; }
                 public UndertaleSprite Sprite { get => _Sprite.Resource; set { _Sprite.Resource = value; OnPropertyChanged(); ParentLayer.ParentRoom.UpdateBGColorLayer(); } }
                 public bool TiledHorizontally { get => _TiledHorizontally; set { _TiledHorizontally = value; OnPropertyChanged(); } }
                 public bool TiledVertically { get => _TiledVertically; set { _TiledVertically = value; OnPropertyChanged(); } }
                 public bool Stretch { get => _Stretch; set { _Stretch = value; OnPropertyChanged(); } }
-                public uint Color { get; set; }
+                public uint Color { get; set; } = 0xFF000000;
                 public float FirstFrame { get; set; }
                 public float AnimationSpeed { get; set; }
                 public AnimationSpeedType AnimationSpeedType { get; set; }
--- a/UndertaleModLib/Models/UndertaleRoom.cs
+++ b/UndertaleModLib/Models/UndertaleRoom.cs
@@ -169,19 +169,19 @@ namespace UndertaleModLib.Models
         public void UpdateBGColorLayer() => OnPropertyChanged("BGColorLayer");
         public void RearrangeLayers(Layer selectedLayer = null)
         {
-            if (Layers.Count > 0)
+            if (Layers.Count == 0)
+                return;
+
+            Layer[] orderedLayers = Layers.OrderBy(l => l.LayerDepth).ToArray();
+
+            // ensure that room objects tree will have the layer to re-select
+            if (selectedLayer is not null)
+                Layers[Array.IndexOf(orderedLayers, selectedLayer)] = selectedLayer;
+
+            for (int i = 0; i < orderedLayers.Length; i++)
             {
-                Layer[] orderedLayers = Layers.OrderBy(l => l.LayerDepth).ToArray();
-
-                // ensure that room objects tree will have the layer to re-select
-                if (selectedLayer is not null)
-                    Layers[Array.IndexOf(orderedLayers, selectedLayer)] = selectedLayer;
-
-                for (int i = 0; i < orderedLayers.Length; i++)
-                {
-                    if (Layers[i] != orderedLayers[i])
-                        Layers[i] = orderedLayers[i];
-                }
+                if (Layers[i] != orderedLayers[i])
+                    Layers[i] = orderedLayers[i];
             }
         }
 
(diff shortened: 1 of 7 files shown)


UndertaleModLib/Models/UndertaleRoom.cs
```diff
@@ -1527,6 +1554,12 @@ public void Unserialize(UndertaleReader reader)
                 Rotation = reader.ReadSingle();
             }
 
+            public static UndertaleString GenerateRandomName(UndertaleData data)
+            {
+                // The same format as in "GameMaker Studio: 2".
```
Miepee, line 1559: No need for quotation marks.

UndertaleModLib/Models/UndertaleRoom.cs
```diff
@@ -1527,6 +1554,12 @@ public void Unserialize(UndertaleReader reader)
                 Rotation = reader.ReadSingle();
             }
 
+            public static UndertaleString GenerateRandomName(UndertaleData data)
```
Miepee, line 1557: Can you add documentation please?

UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs
```diff
@@ -44,6 +45,10 @@ public partial class UndertaleRoomEditor : DataUserControl
 
         public static readonly PropertyInfo visualOffProp = typeof(Canvas).GetProperty("VisualOffset", BindingFlags.NonPublic | BindingFlags.Instance);
         private static readonly MainWindow mainWindow = Application.Current.MainWindow as MainWindow;
+        private static readonly Regex endNumberEx = new(@"\d+$", RegexOptions.Compiled);
+
+        // used for the flashing animation
```
Miepee, line 50: adding "when an element is selected" makes it a bit clearer on what kind of flashing anim it is.

UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs
```diff
@@ -44,6 +45,10 @@ public partial class UndertaleRoomEditor : DataUserControl
...
+        public static Dictionary<UndertaleObject, FrameworkElement> ObjElemDict { get; } = new();
```
Miepee, line 51: Bit of a nitpick and out of scope for this PR, but this along with the other dicts here should get renamed to have clearer names.

UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs
```diff
@@ -44,6 +45,10 @@ public partial class UndertaleRoomEditor : DataUserControl
 
         public static readonly PropertyInfo visualOffProp = typeof(Canvas).GetProperty("VisualOffset", BindingFlags.NonPublic | BindingFlags.Instance);
         private static readonly MainWindow mainWindow = Application.Current.MainWindow as MainWindow;
+        private static readonly Regex endNumberEx = new(@"\d+$", RegexOptions.Compiled);
```
Miepee, line 48: Personally don't like how this is named.

UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs
```diff
@@ -155,6 +165,19 @@ private void UndertaleRoomEditor_DataContextChanged(object sender, DependencyPro
...
                 if (room.Layers.Count > 0) // if GMS 2+
                 {
+                    LayerZIndexConverter.ProcessOnce = true;
+
+                    bool ordered = true;
+                    for (int i = 0; i < room.Layers.Count - 1; i++)
+                    {
+                        if (room.Layers[i].LayerDepth > room.Layers[i + 1].LayerDepth)
+                            ordered = false;
+                    }
```
Miepee, line 175:
```suggestion
                    {   
                        if (room.Layers[i].LayerDepth > room.Layers[i + 1].LayerDepth)
                            ordered = false;
                            
                        if (!ordered) break;
                    }
```

UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs
```diff
@@ -155,6 +165,19 @@ private void UndertaleRoomEditor_DataContextChanged(object sender, DependencyPro
...
+
+                    if (!ordered)
+                        if (MainWindow.ShowQuestion("Room layers are not ordered by depth.\nRearrange them?", MessageBoxImage.Warning) == MessageBoxResult.Yes)
+                            room.RearrangeLayers();
```
Miepee, line 179:
```suggestion
                    if (!ordered)
                    {
                        if (MainWindow.ShowQuestion("Room layers are not ordered by depth.\nRearrange them?", MessageBoxImage.Warning) == MessageBoxResult.Yes)
                            room.RearrangeLayers();
                     }
```

UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs
```diff
@@ -1221,22 +1237,75 @@ private void AddLayer<T>(LayerType type, string name) where T : Layer.LayerData,
...
+            int layerDepth = 0;
+            if (room.Layers.Count > 0)
+            {
+                layerDepth = room.Layers.Select(l => l.LayerDepth).Max();
+                if (layerDepth + 100 > int.MaxValue)
+                {
+                    if (layerDepth + 1 > int.MaxValue)
+                    {
+                        layerDepth -= 1;
+                        MainWindow.ShowWarning("Warning - a maximum layer depth is reached.\nYou probably should change the depth of the new layer.");
```
Miepee, line 1249:
```suggestion
                        MainWindow.ShowWarning("Warning - the maximum layer depth is reached.\nYou probably should change the depth of the new layer.");
```

UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs
```diff
@@ -1221,22 +1237,75 @@ private void AddLayer<T>(LayerType type, string name) where T : Layer.LayerData,
...
+                        name = baseName + nameNum;
+                    }
+                    else
+                    {
+                        // "name" => "name1"
+                        name += 1;
+                    }
+                }
+                else
+                    name = baseName + (++nameNum);
```
Miepee, line 1287: Would prefer if this whole thing were a tad better commented.

UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs
```diff
@@ -1221,22 +1237,75 @@ private void AddLayer<T>(LayerType type, string name) where T : Layer.LayerData,
...
 
             if (layer.LayerType == LayerType.Assets)
             {
-                // create a new pointer list (if null)
+                // "??=" - assign if null
                 layer.AssetsData.LegacyTiles ??= new UndertalePointerList<Tile>();
-                // create new sprite pointer list (if null)
                 layer.AssetsData.Sprites ??= new UndertalePointerList<SpriteInstance>();
-                // create new sequence pointer list (if null)
                 layer.AssetsData.Sequences ??= new UndertalePointerList<SequenceInstance>();
```
Miepee, line 1304: Don't see why there was a need to remove the comments tbh.

UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs
```diff
@@ -1221,22 +1237,75 @@ private void AddLayer<T>(LayerType type, string name) where T : Layer.LayerData,
...
+
+                // if GMS version is less that 2.3.2
+                if (!data.GMS2_3_2)
```
Miepee, line 1307: Comment is incorrect or check is incorrect.
`!data.GM2_3_2` can also get triggered on versions above gms232

UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs
```diff
@@ -1571,20 +1544,164 @@ public static void GenerateSpriteCache(UndertaleRoom room)
                 loader.Convert(texture, null, "generate", null);
             });
         }
+
+        private void TileDataExport_Click(object sender, RoutedEventArgs e)
+        {
+            if (RoomObjectsTree.SelectedItem is Layer layer)
+            {
```
Miepee, line 1551: Make this return early

UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs
```diff
@@ -1571,20 +1544,164 @@ public static void GenerateSpriteCache(UndertaleRoom room)
...
+                    return;
+                }
+
+                MainWindow.ShowMessage("Exported file path: " + filePath);
+            }
+        }
+        private void TileDataImport_Click(object sender, RoutedEventArgs e)
+        {
+            if (RoomObjectsTree.SelectedItem is Layer layer)
+            {
```
Miepee, line 1581: Make this return early

UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs
```diff
@@ -1571,20 +1544,164 @@ public static void GenerateSpriteCache(UndertaleRoom room)
...
+
+                    MainWindow.ShowMessage("Imported successfully.");
+                }
+            }
+        }
+
+        private void LayerCanvas_DataContextChanged(object sender, DependencyPropertyChangedEventArgs e)
+        {
+            // raised on layers rearrange
+            if (e.NewValue is Layer layer)
```
Miepee, line 1646: Make this return early

Miepee, line 1646: Make this return early

UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs
```diff
@@ -1571,20 +1544,164 @@ public static void GenerateSpriteCache(UndertaleRoom room)
...
+
+                canvas.CurrentLayer = layer;
+                ObjElemDict[layer] = canvas;
+            }
+        }
+        private void LayerCanvas_Initialized(object sender, EventArgs e)
+        {
+            LayerCanvas canvas = sender as LayerCanvas;
+
+            if (canvas?.DataContext is Layer layer)
```
Miepee, line 1658: Make this return early

UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs
Miepee: Also, not sure yet if popping up this question every time could get annoying.

Miepee:
You forgot to note that it fixes #880. Otherwise everything seems fine.

UndertaleModLib/Models/UndertaleRoom.cs
```diff
@@ -167,6 +167,23 @@ public enum RoomEntryFlags : uint
...
+                // ensure that room objects tree will have the layer to re-select
+                if (selectedLayer is not null)
+                    Layers[Array.IndexOf(orderedLayers, selectedLayer)] = selectedLayer;
+
+                for (int i = 0; i < orderedLayers.Length; i++)
+                {
+                    if (Layers[i] != orderedLayers[i])
+                        Layers[i] = orderedLayers[i];
+                }
+            }
```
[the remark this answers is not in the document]
VladiStep: Will be applied separately _(not with "Commit suggestion")_.

UndertaleModLib/Models/UndertaleRoom.cs
VladiStep: There's a `:` character in the title, so I think they should stay here.

UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs
VladiStep: But how else to call it?

UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs
VladiStep: I'll consider it,

UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs
Miepee: Something akin to `trailingNumbersPattern`?

UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs
VladiStep: I've thought about it.
What if ask another question on _"No"_, like _"Don't ask this question for this session?"_ ?

Miepee: Personally, not that much of a fan of that.

UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs
VladiStep: Yeah, then it will be `trailingNumberRegex`.

UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs
Miepee: That's weird, I'm pretty sure they're still used in 2022 something right now.

VladiStep: No, I think only the source comment and my comment should be rewritten

UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs
VladiStep: So, then I guess it should stay as is _(curly brackets too)_.

UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs
VladiStep: It's small, so IMO it should stay as is.

UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs
Miepee: Yeah, if it gets annoying at one point, someone'll probably complain. Or me.

UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs
Miepee: Still looks cleaner IMO

Miepee: Still looks cleaner IMO

VladiStep: I've changed my mind, so it will be changed.

UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs
VladiStep: But is it necessary to insert comments between `if-else`, so they make a gap between them?

VladiStep: E.g.
```
}
// if base name is already extracted, increment "nameNum" and append it to the base name
else
```

VladiStep: Also, why you've extracted `nameNum++` to a separate line?
Why not using the possibility to simplify that to `(++nameNum)` ?

Miepee: > But is it necessary to insert comments between if-else, so they make a gap between them?

Because that actually makes it clearer from the outside what the individual chunks do.

> Why not using the possibility to simplify that to (++nameNum) ?

because it's harder to read.

VladiStep: And what about this?
```c#
}
else // if base name is already extracted, increment "nameNum" and append it to the base name
{
```

VladiStep: Hmm... then fine.
But the second null check for `baseName` is for extracting the name and number only once, so it shouldn't be deleted.

Miepee: It is an impossible scenario.
There is no way that basename is null in the first null check, but not null in the second

Miepee (approved):

PR #886, opened 2022-05-04
Optimize TexData (mostly QoiConverter)

cgytrus:
Optimizes QoiConverter, makes TexData store a bitmap instead of PNG byte array and replaces most TextureBlob usages with a direct reference to the bitmap
The memory usage deserializing QOI images is drastically lower:
image

~~There isn't much that can be optimized in the serialization but it's still a bit better:~~
<details>
<summary>old screenshot</summary>

image
</details>

Memory usage here is also way lower in serialization now (about 40MB, instead of 380MB on a random 4K wallpaper image)

PR #892, opened 2022-05-07
Fix games with PNG images

cgytrus:
## Description
fixes games with png images not working after the qoi optimization pr (#886)

### Caveats
none?

### Notes
what.


Miepee:
Confirmed by me that it fixes the issue.

PR #893, opened 2022-05-08
Revert qoi changes

Miepee:
## Description
Revert #886 and #892. While the benefit of them reducing memory usage drastically is nice, they currently break and/or crash a lot of functionality, due to barely any testing done.

### Caveats
Bring memory usage from qoi converting back into potential gigabytes again.


Miepee:
Closed in favor of #895

PR #921, opened 2022-05-21
Correct function definitions and 2.3 FancyRoomSelect

### LINKED ISSUE #951: GM2022 game with a new room can't be saved.
### Describe the bug

When you create a room layer in GM2022 game, it can't be saved, because `Layer.EffectProperties` is `null`.

### Reproducing steps

1. Create (add) a new layer (of any type) in GM2022 game.
2. Try to save the file.
3. Error.

### Setup Details

UTMT version - commit [94ea048](https://github.com/krzys-h/UndertaleModTool/commit/94ea0488245bedd4edbecfa085b20e149843390f).

Jacky720:
## Description
Adds support for `function scr_myscript()` formatting in GMS 2.3+ in both compilation and decompilation. Minor tweaks to FancyRoomSelect.csx to function on 2.3+ with strange room orders.

### Caveats
Struct definition isn't 100% due to me not knowing how struct definition works. The new decompiler code hackily uses `FunctionDefinition` as both an Expression and a Statement depending on context. The code does not differentiate in decompilation; `scr_myscript = function()` no longer appears even if it is the format used, although it is still accepted in compilation.

I didn't bother to restrict the `myroom_last` variable to 2.3, so technically it defines one additional unused variable when the script is used on older games. This will not cause performance issues.

### Notes
Closes #926. Closes #951.

### DIFF
--- a/UndertaleModLib/Compiler/Parser.cs
+++ b/UndertaleModLib/Compiler/Parser.cs
@@ -693,14 +693,13 @@ namespace UndertaleModLib.Compiler
                 result.Children.Add(ParseStatement(context));
                 if (expression)
                     return result;
-                else // Whatever you call non-anonymous
+                else // Whatever you call non-anonymous definitions
                 {
-                    Statement trueresult = new Statement(Statement.StatementKind.Assign, new Lexer.Token(TokenKind.Assign));
-                    trueresult.Children.Add(destination);
-                    trueresult.Children.Add(new Statement(Statement.StatementKind.Token, trueresult.Token));
-                    trueresult.Children.Add(result);
-                    return trueresult;
-
+                    Statement trueResult = new Statement(Statement.StatementKind.Assign, new Lexer.Token(TokenKind.Assign));
+                    trueResult.Children.Add(destination);
+                    trueResult.Children.Add(new Statement(Statement.StatementKind.Token, trueResult.Token));
+                    trueResult.Children.Add(result);
+                    return trueResult;
                 }
             }
 
--- a/UndertaleModLib/Compiler/Parser.cs
+++ b/UndertaleModLib/Compiler/Parser.cs
@@ -660,10 +660,19 @@ namespace UndertaleModLib.Compiler
             {
                 Statement result = new Statement(Statement.StatementKind.FunctionDef, EnsureTokenKind(TokenKind.KeywordFunction).Token);
                 Statement args = new Statement();
+                bool expression = true;
+                Statement destination = null;
+
+                if (GetNextTokenKind() == TokenKind.ProcFunction)
+                {
+                    expression = false;
+                    Statement s = remainingStageOne.Dequeue();
+                    destination = new Statement(Statement.StatementKind.ExprFuncName, s.Token) { ID = s.ID };
+                }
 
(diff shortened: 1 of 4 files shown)


UndertaleModLib/Compiler/Parser.cs
```diff
@@ -682,7 +691,17 @@ private static Statement ParseFunction(CompileContext context)
                 if (EnsureTokenKind(TokenKind.CloseParen) == null) return null;
 
                 result.Children.Add(ParseStatement(context));
-                return result;
+                if (expression)
+                    return result;
+                else // Whatever you call non-anonymous
```
Miepee, line 696: Else is unnecessary here, due to the return from above.

UndertaleModLib/Compiler/Parser.cs
```diff
@@ -682,7 +691,17 @@ private static Statement ParseFunction(CompileContext context)
...
+                {
+                    Statement trueresult = new Statement(Statement.StatementKind.Assign, new Lexer.Token(TokenKind.Assign));
```
Miepee, line 698: variable should get named to `trueResult`

UndertaleModLib/Compiler/Parser.cs
```diff
@@ -682,7 +691,17 @@ private static Statement ParseFunction(CompileContext context)
...
+                    trueresult.Children.Add(destination);
+                    trueresult.Children.Add(new Statement(Statement.StatementKind.Token, trueresult.Token));
+                    trueresult.Children.Add(result);
+                    return trueresult;
+
```
Miepee, line 703: newline can be removed here.

UndertaleModTool/CommunityScripts/FancyRoomSelect.csx
```diff
@@ -411,7 +408,7 @@ exiting = 0
 ", Data);
 
 
-ScriptMessage("Successfully applied for Gamemaker " + (gms2 ? 2 : 1).ToString() + @"
+ScriptMessage("Successfully applied for Gamemaker " + (Data.GMS2_3 ? "2.3" : (gms2 ? "2" : "1")) + @"
```
Miepee, line 411: Would personally prefer if the GM variable would get split into seperate statement(s), rather than trying to force it into the scriptMessage here.

BenjaminUrquhart:
So, the reason why struct decompilation uses `FunctionDefinition` is because as far as I can tell, the bytecode is identical. The only difference is a variable in the code entry that's set to an identical struct. Go figure.

Jacky720:
Oh, that worked pretty well.

No more conflicts, right?

Miepee (approved):
Doesn't seem to be.
Would still personally prefer to get rid of the else to reduce nesting, but that's not something that's merge blocking or anything.

UndertaleModLib/Compiler/Parser.cs
Jacky720: I feel that keeping both returns on the same level makes it clear at a glance that neither is especially a "default". You can't miss the if anymore and assume it always does the one thing.

Miepee: Which is fair, I just dislike the nesting. Probably a better thing to have in this case though.

PR #927, opened 2022-05-22
Fix message boxes except everything's broken

### LINKED ISSUE #878: MessageBoxes show behind the main window / other boxes
It's fairly common for message boxes to get hidden behind other windows. It would be a very nice improvement if they'd *not* do that, and/or the UMT icon would flicker if a new messagebox showed up and UMT is not in focus.

Archie-osu:
## Description
Fixes #878 again, meaning message boxes don't appear behind the UMT window now.

### Caveats
Literally everything is broken (???)
Nyako and Miepee tried very hard to fix it, but failed (I think???)

### Notes
This is very much incomplete and needs review.

### DIFF
--- a/UndertaleModTool/Controls/UndertaleObjectReference.xaml.cs
+++ b/UndertaleModTool/Controls/UndertaleObjectReference.xaml.cs
@@ -95,9 +95,9 @@ namespace UndertaleModTool
         {
             if (ObjectReference is null)
             {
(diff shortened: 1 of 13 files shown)


VladiStep:
There are still some untouched `Application.Current.MainWindow as MainWindow`.
Also, `MainWindow.ShowX()` methods would crash if `wait = false` - there shouldn't be `owner` in `MessageBox.Show()` if `wait = false`.

Issue #555, opened 2021-09-20
Add try/catch support to the decompiler


krzys-h:

Miepee:
Since me (and potentially future people) was confused by this:
It is referring to GMS2' try catch finally feature
https://manual.yoyogames.com/GameMaker_Language/GML_Overview/Language_Features/try_catch_finally.htm

DT-is-not-available:
This would be incredibly helpful.

PR #930, opened 2022-05-24
First pass at try-catch decompilation

BenjaminUrquhart:
## Description
This PR attempts to add try-catch decompilation to the tool. It's an absolute mess and incomplete right now which is why this is a draft PR.

### Caveats
- Nested try-catch blocks hide code that comes after them
- Finally has not been implemented yet
- Literally a hack to a hack this PR only exists for completeness

### Notes
I hate everything


Miepee:
Aware that it is draft, but still fixes #555

Raltyro:
We really need this with compilation too

PR #931, opened 2022-05-25
New grid related features

plrusek:
## Description
1. The grid now has been moved from being rendered in the background of the canvas (meaning it could be covered up by any room object easily) to the foreground of the canvas (meaning it will render on top of everything), thanks to the use of an opacity mask on the canvas.
2. The automatically calculated grid height and width per room (based on the width and height of the most used tile in said room) can now be overriden by global properties that can be enabled/disabled and then set in the settings.
3. The settings window itself had a slight reordering of it's contents (slighty tidyup, empty horizontal space was filled out)


UndertaleModTool/Editors/UndertaleRoomEditor.xaml
```diff
@@ -1053,7 +1080,6 @@
                         <Setter Property="RenderOptions.BitmapScalingMode" Value="NearestNeighbor"/>
                         <Setter Property="UIElement.SnapsToDevicePixels" Value="True"/>
                     </Style>
-
                     <DataTemplate DataType="{x:Type undertale:UndertaleRoom+Layer}">
```
VladiStep, line 1083:
```suggestion

                    <DataTemplate DataType="{x:Type undertale:UndertaleRoom+Layer}">
```

UndertaleModTool/Editors/UndertaleRoomEditor.xaml
```diff
@@ -1081,7 +1107,6 @@
                             </ItemsControl.Style>
                         </ItemsControl>
                     </DataTemplate>
-
                     <DataTemplate DataType="{x:Type undertale:UndertaleRoom+Background}">
```
VladiStep, line 1110:
```suggestion

                    <DataTemplate DataType="{x:Type undertale:UndertaleRoom+Background}">
```

UndertaleModTool/Editors/UndertaleRoomEditor.xaml
VladiStep: This empty line was placed intentionally.

UndertaleModTool/Editors/UndertaleRoomEditor.xaml
VladiStep: This empty line was placed intentionally.

VladiStep (changes requested):

UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs
```diff
@@ -832,6 +832,7 @@ private void SelectObject(UndertaleObject obj)
 
                 resListView.IsExpanded = true;
                 resListView.BringIntoView();
+                resListView.Focus();
```
VladiStep, line 835: It should be after line 856 (`resItem.IsSelected = true;`).

UndertaleModLib/Models/UndertaleRoom.cs
```diff
@@ -403,11 +403,18 @@ public void SetupRoom(bool calculateGrid = true)
             }
 
             // If tiles exist at all, grab the most used tile size and use that as our grid size
+            // If a default starting grid is set in the settings, use that instead
```
Miepee, line 406: This comment is slightly misleading, as in this code block the starting grid is *not* get from the settings as the grid isn't even recalculated in that case.

UndertaleModLib/Models/UndertaleRoom.cs
```diff
@@ -403,11 +403,18 @@ public void SetupRoom(bool calculateGrid = true)
...
-                GridWidth = largestKey.X;
-                GridHeight = largestKey.Y;
+                if (calculateGridWidth)
+                {
+                    GridWidth = largestKey.X;
+                }
+                if (calculateGridHeight)
+                {
+                    GridHeight = largestKey.Y;
+                }
```
Miepee, line 417: Can be condensed
```suggestion
                if (calculateGridWidth)
                    GridWidth = largestKey.X;
                    
                if (calculateGridHeight)
                    GridHeight = largestKey.Y;            
```

UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs
```diff
@@ -1670,6 +1671,26 @@ private void LayerCanvas_Unloaded(object sender, RoutedEventArgs e)
...
+                }
+                else
+                {
+                    room.SetupRoom(false);
+                }
+            }
+            else
+            {
+                room.SetupRoom();
+            }
```
Miepee, line 1692: This logic here is very weird.
The way it is coded right now, it will only use a forced height, if a forced width is given as well.

UndertaleModLib/Models/UndertaleRoom.cs
```diff
@@ -129,7 +129,7 @@ public enum RoomEntryFlags : uint
     /// <summary>
     /// The thickness of the room grid in pixels.
     /// </summary>
-    public double GridThicknessPx { get; set; } = 1d;
+    public double GridThicknessPx { get; set; } = 0.5d;
```
Miepee, line 132: Obviously preference, but I think reverting the grid thickness would be better. Currently it looks way too thin IMO.

Miepee:
Input boxes need some work:
- I can add a lot of numbers here image
- As mentioned in the comments from above, only setting height currently does not work as expected.
- For a better user experience, it would be better if one couldn't input normal text in there
- Default grid sizes should probably be 16x16 instead of 0x0
- Setting negative grid sizes crashes the room editor
- putting too high or too low numbers shows an error message. Would be better if a user couldn't input them in the first place, or if the numbers would be snapped to what makes sense (aka ]0;Int32.MaxInt])
- same as above but with `NaN` and `-NaN`
- -0 is apparently a valid grid size
- Having thickness be enforceable via settings would also be nice
- Exporting the room as a PNG exports the grid as well which is undesirable.

UndertaleModTool/Controls/UndertaleRoomRenderer.xaml
```diff
@@ -81,12 +81,42 @@
                     <Canvas x:Name="RoomCanvas" Width="{Binding Width, Mode=OneTime}" Height="{Binding Height, Mode=OneTime}" IsItemsHost="True" AllowDrop="True" ClipToBounds="True"
                             SnapsToDevicePixels="True" RenderOptions.BitmapScalingMode="NearestNeighbor"
                             Loaded="RoomCanvas_Loaded">
+                        <Canvas.OpacityMask>
+                            <DrawingBrush TileMode="Tile" ViewportUnits="Absolute">
+                                <DrawingBrush.Viewport>
+                                    <MultiBinding Converter="{StaticResource GridConverter}">
+                                        <Binding Path="GridWidth" Mode="OneWay"/>
```
VladiStep, line 88: All the data bindings that you've changed/added should be `OneTime`.

Miepee:
Vlad brought up something in the server which I think is worth noting here.
Currently, with the revamped settings, columns are a bit hard to distinguish.
image

Something that could help with this, would be to move the checkboxes *before* the text, similar to how other tools do it. (using JBRider here as an example:)
image
The above path textboxes can be left as-is, but the textboxes for width/height should be then also rearranged to have the activation checkbox on the left, with the textbox on the right like here:
image

Miepee:
Thanks for the PR! Unfortunate, that you weren't able to get this one merged and that we needed to create our own PRs for it, but nevertheless your contributions are very welcomed :)

PR #946, opened 2022-05-29
File path related fixes and improvements.

### LINKED ISSUE #690: Improper access to "dotnet" directory while using ImportGraphics
unknown

### LINKED ISSUE #943: Incorrect path for saving while doing temp run involving GMLCache
image

VladiStep:
## Description

1. **Fixed bug when sometimes UTMT tries to access one of system folders instead of the app folder.**
2. Fixed the following scripts: "ImportShaderData.csx", "FindUnknownFunctions.csx".
3. Some of `System.IO` class references are simplified (e.g. `System.IO.Path` -> `Path`).
4. All the scripts that are modifying textures create "Packager" folder in UTMT folder.
5. GML cache load and save methods use `Path.Combine()` instead of `Path.Join()` for consistency with other code.

Closes #690, closes #943.

### DIFF
--- a/UndertaleModTool/MainWindow.xaml.cs
+++ b/UndertaleModTool/MainWindow.xaml.cs
@@ -1155,11 +1155,17 @@ namespace UndertaleModTool
         private async Task LoadGMLCache(string filename, LoaderDialog dialog = null)
         {
             await Task.Run(() => {
-                if (SettingsWindow.UseGMLCache && File.Exists(Path.Join("GMLCache", "index")))
+                if (SettingsWindow.UseGMLCache)
                 {
+                    string cacheDirPath = Path.Combine(ExePath, "GMLCache");
+                    string cacheIndexPath = Path.Combine(cacheDirPath, "index");
+
+                    if (!File.Exists(cacheIndexPath))
+                        return;
+
                     dialog?.Dispatcher.Invoke(() => dialog.ReportProgress("Loading decompiled code cache..."));
 
-                    string[] indexLines = File.ReadAllLines(Path.Join("GMLCache", "index"));
+                    string[] indexLines = File.ReadAllLines(cacheIndexPath);
 
                     int num = -1;
                     for (int i = 0; i < indexLines.Length; i++)
@@ -1172,7 +1178,7 @@ namespace UndertaleModTool
                     if (num == -1)
                         return;
 
-                    if (!File.Exists(Path.Join("GMLCache", num.ToString())))
+                    if (!File.Exists(Path.Combine(cacheDirPath, num.ToString())))
                     {
                         ShowWarning("Decompiled code cache file for open data is missing, but its name present in the index.");
 
@@ -1181,7 +1187,7 @@ namespace UndertaleModTool
 
                     string hash = GenerateMD5(filename);
 
-                    using (StreamReader fs = new(Path.Join("GMLCache", num.ToString())))
+                    using (StreamReader fs = new(Path.Combine(cacheDirPath, num.ToString())))
                     {
                         string prevHash = fs.ReadLine();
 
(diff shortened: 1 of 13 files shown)


UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -1155,11 +1155,17 @@ public string GenerateMD5(string filename)
         private async Task LoadGMLCache(string filename, LoaderDialog dialog = null)
         {
             await Task.Run(() => {
-                if (SettingsWindow.UseGMLCache && File.Exists(Path.Join("GMLCache", "index")))
+                if (SettingsWindow.UseGMLCache)
```
Miepee, line 1158: Probably better for future PR, but this (and savegmlcache) if condition can be negated.

UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -1155,11 +1155,17 @@ public string GenerateMD5(string filename)
...
                 {
+                    string cacheDirPath = Path.Join(ExePath, "GMLCache");
+                    string cacheIndexPath = Path.Join(cacheDirPath, "index");
```
Miepee, line 1161: Why Path.Join instead of Path.Combine here?

UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -1242,14 +1248,16 @@ private async Task SaveGMLCache(string filename, bool updateCache = true, Loader
                 {
                     dialog?.Dispatcher.Invoke(() => dialog.ReportProgress("Saving decompiled code cache..."));
 
-                    if (!File.Exists(Path.Join("GMLCache", "index")))
+                    string cacheDirPath = Path.Join(ExePath, "GMLCache");
+                    string cacheIndexPath = Path.Join(cacheDirPath, "index");
```
Miepee, line 1252: Same question as above.

UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -1242,14 +1248,16 @@ private async Task SaveGMLCache(string filename, bool updateCache = true, Loader
...
+                    if (!File.Exists(cacheIndexPath))
                     {
-                        Directory.CreateDirectory("GMLCache");
+                        Directory.CreateDirectory(cacheDirPath);
```
Miepee, line 1256: Unnecessary newline.

UndertaleModTool/Scripts/Repackers/ImportShaderData.csx
```diff
@@ -1,33 +1,23 @@
...
 EnsureDataLoaded();
 
-// "Select 'Import_Loc.txt' file in 'Shader_Data'"
 string importFolder = PromptChooseDirectory();
 if (importFolder == null)
     throw new ScriptException("The import folder was not set.");
 
-string[] dirFiles = Directory.GetFiles(importFolder, "*.*", SearchOption.AllDirectories);
-List<string> shadersToModify = new List<string>();
+var shadersToModify = Directory.GetDirectories(importFolder).Select(x => Path.GetFileName(x));
```
Miepee, line 14: This does not seem to be the equivalent of the foreach loop from before.

UndertaleModTool/MainWindow.xaml.cs
VladiStep: IIRC, because I thought that `Join()` will be enough here, as `Combine()` does more work.

UndertaleModTool/Scripts/Repackers/ImportShaderData.csx
VladiStep: Yes, because the algorithm is changed completely.
This approach is better.

UndertaleModTool/MainWindow.xaml.cs
Miepee: Combine does more work, but it's also a little safer, and considering that the rest of the codebase uses combine consistency would be nice here.

UndertaleModTool/Scripts/Repackers/ImportShaderData.csx
Miepee: Double checked, see now.

UndertaleModTool/MainWindow.xaml.cs
VladiStep: Yeah, so I'll change this.

Miepee (approved):

PR #947, opened 2022-05-31
Room layers rearrangement related fixes.

VladiStep:
## Description

1. **Fixed the crashes on following events: opening a room with layers that are not ordered by depth, new layer addition (not consistent).**
2. **Fixed new layer depth overflow check that caused negative depth of the new layer.**

### DIFF
--- a/UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs
+++ b/UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs
@@ -1240,7 +1240,7 @@ namespace UndertaleModTool
                 }
             }
 
-            int layerDepth = 0;
+            long layerDepth = 0;
             if (room.Layers.Count > 0)
             {
                 layerDepth = room.Layers.Select(l => l.LayerDepth).Max();
@@ -1293,7 +1293,7 @@ namespace UndertaleModTool
             layer.LayerName = data.Strings.MakeString(name);
             layer.LayerId = largest_layerid + 1;
             layer.LayerType = type;
-            layer.LayerDepth = layerDepth;
+            layer.LayerDepth = (int)layerDepth;
             layer.Data = new T();
             room.Layers.Add(layer);
             room.UpdateBGColorLayer();
--- a/UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs
+++ b/UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs
@@ -103,6 +103,12 @@ namespace UndertaleModTool
             visualOffProp.SetValue(roomCanvas, prevOffset);
         }
 
+        /// <summary>
+        /// Checks if the room layers are ordered by depth. If they are not, the user will be prompted,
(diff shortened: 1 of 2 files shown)


UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs
```diff
@@ -103,6 +103,23 @@ public void SaveImagePNG(Stream outfile)
             visualOffProp.SetValue(roomCanvas, prevOffset);
         }
 
+        public static void CheckAndRearrangeLayers(UndertaleRoom room)
```
Miepee, line 106: Documentation for the method would be nice.

UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs
```diff
@@ -1240,7 +1246,7 @@ private void AddLayer<T>(LayerType type, string name) where T : Layer.LayerData,
                 }
             }
 
-            int layerDepth = 0;
+            long layerDepth = 0;
```
Miepee, line 1249: Don't see it right now: why does this need to be a long?

VladiStep: To be able to check if new layer depth is more than `int.MaxValue`, because `int.MaxValue + 1` equals `int.MinValue`.

Miepee: Might wanna add a comment for that then

UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs
VladiStep: Is this fine?
```xml
/// <summary>
/// Checks if the layers are ordered by depth.<br/>
/// If they are not, then it will ask if they should be ordered.<br/>
/// If yes, then it will rearrange the layers.
/// </summary>
/// <param name="room">The <see cref="UndertaleRoom"/>, whose layers should be checked and rearranged.</param>
```

UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs
```diff
@@ -1246,13 +1254,15 @@ private void AddLayer<T>(LayerType type, string name) where T : Layer.LayerData,
                 }
             }
 
+            // "layerDepth" is "long", because otherwise it couldn't be checked if incremented value is larger than "Int32.MaxValue",
+            // because then it will overflow.
```
Miepee, line 1258:
```suggestion
            // "layerDepth" is "long", because otherwise one can't check if the incremented value is larger than "Int32.MaxValue",
            // because then it'd overflow.
```

Miepee: Yeah. One is a generic term for a person.

Miepee (approved):

PR #948, opened 2022-06-02
Fix of QOI textures display on different system scale setting.

VladiStep:
## Description

Textures of GMS 2022 games are displayed correctly independently of system display scale options.
Also, a embedded texture preview looks sharp now ("nearest neighbor" is used).

### Caveats
The preview can be considered "choppy" by some users, especially for large embedded textures.


Miepee:
What about the previews on the sprite/room/background/font/object editors?

VladiStep:
> What about the previews on the sprite/room/background/font/object editors?

Out of the ones you listed, only sprite mask preview doesn't use "nearest neighbor".

Miepee:
Might wanna add a fix for that as well then.

Miepee (approved):

PR #936, opened 2022-05-26
Fix MessageBox focus issues v2

Archie-osu:
## Description
Fixes #878 by using a custom ``OwnedMessageBox`` class, which acts as a wrapper for ``MessageBox.Show``.

### Caveats
Scripts have not been changed, afraid of breaking stuff.

### Notes
Please don't merge this until it's reviewed, we all know my C# knowledge isn't ideal.

PR #956, opened 2022-06-11
Fix MessageBox focus issues v3

### LINKED ISSUE #878: MessageBoxes show behind the main window / other boxes

Miepee:
## Description
Fixes #878, by using a custom MessageBox wrapper which in general makes the use of messageBoxes much cleaner.

### Caveats
None

### Notes
- Scripts have not been touched yet.
- Obsoletes #936 
- A few instances of messageBoxes have been kept as-is in order not to break stuff.

### DIFF
--- a/UndertaleModTool/Controls/UndertaleObjectReference.xaml.cs
+++ b/UndertaleModTool/Controls/UndertaleObjectReference.xaml.cs
@@ -104,7 +104,7 @@ namespace UndertaleModTool
 
                 if (mainWindow.Selected is null)
                 {
-                    MainWindow.ShowError("Nothing currently selected! This is currently unsupported.");
+                    mainWindow.ShowError("Nothing currently selected! This is currently unsupported.");
                     return;
                 }
                 else if (mainWindow.Selected is UndertaleGameObject gameObject)
@@ -116,7 +116,7 @@ namespace UndertaleModTool
                 }
                 else
                 {
-                    MainWindow.ShowError("Adding to non-objects is currently unsupported.");
+                    mainWindow.ShowError("Adding to non-objects is currently unsupported.");
                     return;
                 }
             }
--- a/UndertaleModTool/Converters/UndertaleCachedImageLoader.cs
+++ b/UndertaleModTool/Converters/UndertaleCachedImageLoader.cs
@@ -273,7 +273,7 @@ namespace UndertaleModTool
         {
             throw new NotImplementedException();
         }
-    }    
+    }
 
     // UndertaleCachedImageLoader wrappers
     public class CachedTileImageLoader : IMultiValueConverter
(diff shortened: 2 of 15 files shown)


VladiStep:
I suggest to return static `MessageBox` methods of `MainWindow`, but in separate file, and add static `MainWindow.Instance` property.

For example, `MainWindow.ShowMessage()` should call `Instance.ShowMessage()` with the parent arguments.
So, you can call `MainWindow.ShowMessage()` as it was before (without adding `mainWindow` for each class).

VladiStep:
Hmm... actually, `MainWindow.Instance` also can be used as shorter replacement of `Application.Current.MainWindow as MainWindow`

Miepee:
> I suggest to return static MessageBox methods of MainWindow, but in separate file, and add static MainWindow.Instance property.

IMO this complicates things, as you then have basically the same (extension) methods, but in multiple places, where the differences aren't fully clear. 
Since `MainWindow.Instance` is a full replacement for `Application.Current.MainWindow as MainWindow`, it's possible for classes which only use the main window for a few messagebox calls, to just use that directly, instead of defining it as a private property.

VladiStep:
> > I suggest to return static MessageBox methods of MainWindow, but in separate file, and add static MainWindow.Instance property.
> 
> IMO this complicates things, as you then have basically the same (extension) methods, but in multiple places, where the differences aren't fully clear. (Makes me wish tho that the "extend everything" feature would've been implemented...) Since `MainWindow.Instance` is a full replacement for `Application.Current.MainWindow as MainWindow`, it's possible for classes which only use the main window for a few messagebox calls, to just use that directly, instead of defining it as a private property.

Is it really complicates that?
If some other class wants to show some message, then it should use `MainWindow.ShowMessage()` (without using `mainWindow`).
And if it's a `Window`, then it should use `this.ShowMessage()`.

And each static `MessageBox` method will contain only one line of code.

Miepee:
> Is it really complicates that?

You have two classes, with the exact same names. From an outsider view, it's not fully clear why one should use `MainWindow.ShowMessage()` over `MainWindow mw = MainWindow.Instance; mw.ShowMessage()` or vice versa.
And from what I can recall, there also aren't a lot of classes, that'd benefit from such a change to begin with.

VladiStep:
> From an outsider view, it's not fully clear why one should use `MainWindow.ShowMessage()` over `MainWindow mw = MainWindow.Instance; mw.ShowMessage()` or vice versa.

To make it clear, all the classes except `Window` ones should use `MainWindow.ShowMessage()`.
So, if someone would want to show a message within `MainWindow`, they will see that in the class, `this.ShowMessage()` is used, so they also use that.
And if it's other class, they will see `MainWindow.ShowMessage()`, and use that instead.
If they try to use `this.ShowMessage()` there, then it will show an error, and (hopefully) they will guess that it should be `MainWindow.ShowMessage()`.

Also, if they decide to use `MessageBox.Show()` anyway, code reviewers should notice that and suggest to change that.


> ...there also aren't a lot of classes, that'd benefit from such a change to begin with.

That's for now - there can be more classes in the future.


This is only a suggestion, so if you don't want to do it, then you can not do that.
Then, I'll still approve it.

Miepee:
Yeah, it's a nice suggestion, but ATM I don't see the need for this, as well as in the near future.
Not to mention that "reference MainWindow's methods for anything outside of MainWindow, otherwise use the instance" isn't exactly that easy to get across.

VladiStep:
Then, I'm approving it.

VladiStep (approved):

PR #960, opened 2022-06-11
Direct loading and saving game data.

### LINKED ISSUE #520: An error occurred while trying to load: Array dimensions exceeded supported range
Was messing around with trying to extract a non-deltarune/undertale game, and ran into this issue. Assuming the file was too large: file size 2,444,982 kb.

VladiStep:
## Description
Game data loading and saving improvements:
1. **Game data is now loaded and saved directly, without buffering whole file data in memory** - new `FileBinaryReader` and `FileBinaryWriter` are used for that.
So, it's possible to **load and save data files that larger than 2 GB.**
2. **_"Writing references in parallel..."_ part of the data saving is a lot faster** - added the chunks type and object index dictionaries, added `IUndertaleSingleChunk` and `IUndertale(Simple)ListChunk` interfaces.
3. Messages of the final part of data saving are describing the process more precisely.
4. `ProfileSaveEvent()` and `UpdateProfile()` are running on different thread, so UI won't freeze on MD5 calculation.

Other changes:

1. **Fixed some warning/error message boxes not showing up on load or save.**

### Caveats
Game data loading and saving is now more depending on disk read/write load - if disk will be used while loading/saving the data, then it will process the data longer.

### Notes
Closes #520

### DIFF
--- a/UndertaleModLib/Util/FileBinaryWriter.cs
+++ b/UndertaleModLib/Util/FileBinaryWriter.cs
@@ -1,0 +1,74 @@
+﻿using System;
+using System.Buffers.Binary;
+using System.IO;
+using System.Text;
+
+namespace UndertaleModLib.Util
+{
+    public class FileBinaryWriter : BinaryWriter
+    {
+        private readonly Encoding encoding = new UTF8Encoding(false);
+
+        public uint Position
+        {
+            get => (uint)OutStream.Position;
+            set => OutStream.Position = value;
+        }
+
+        public Encoding Encoding { get => encoding; }
+        public Stream BaseStream => OutStream;
+
+        public FileBinaryWriter(Stream stream, Encoding encoding = null) : base(stream)
+        {
+            if (encoding is not null)
+                this.encoding = encoding;
+        }
+
+        public void Write(MemoryStream value)
+        {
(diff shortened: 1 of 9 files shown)


UndertaleModLib/Util/FileBinaryWriter.cs
```diff
@@ -0,0 +1,74 @@
...
+        private readonly Encoding encoding = new UTF8Encoding(false);
+
+        public uint Position
+        {
+            get => (uint)OutStream.Position;
+            set => OutStream.Position = value;
+        }
+
+        public Encoding Encoding { get => encoding; }
+        public Stream BaseStream => OutStream;
```
Miepee, line 19: Compiler warning here, that that `FileBinaryWriter.BaseStream` hides `BinaryWriter.BaseStream`.
from what I can see, you probably want to override it?

Miepee:
Tested out the Cafe game as well, seemed to load and save fine.

UndertaleModLib/Util/FileBinaryWriter.cs
VladiStep: Actually, no, that's just an oversight - it's not even used, so it can be removed

Miepee (approved):

PR #968, opened 2022-06-12
Add some GMS2.3 Methods

### LINKED ISSUE #464: "texture_prefetch" is not known to be a built-in GameMaker function
"texture_prefetch" is identified to be an unknown function when it is used in GameMaker bytecode 17 and later. Please add "texture_prefetch" to the list.

Miepee:
## Description
This adds some GMS2.3 methods (and one keyword) to the known list.
Fixes #464

### Caveats
None


UndertaleModLib/Compiler/BuiltinList.cs
```diff
@@ -1093,6 +1104,12 @@ public void Initialize(UndertaleData data)
             Functions["array_equals"] = new FunctionInfo(this, 2);
             Functions["array_create"] = new FunctionInfo(this, -1);
             Functions["array_copy"] = new FunctionInfo(this, 5);
+            if (data is not null && data.GMS2_3)
```
VladiStep, line 1107:
```suggestion
            if (data?.GMS2_3 == true)
```

VladiStep (approved):

PR #977, opened 2022-06-14
String-related fixes and improvements.

VladiStep:
## Description

1. Fixed missing `UndertaleString` name in the tab titles.
2. **Fixed crash on new `UndertaleString` addition.**
3. "Copy name to clipboard" feature now works with `UndertaleString` items.
4. `UndertaleString` tab title updates on string content change and properly shortens if it contains tab characters.
5. Fixed all the instances of "Name property not found on object of type ..." WPF binding error.

### DIFF
--- a/UndertaleModTool/MainWindow.xaml.cs
+++ b/UndertaleModTool/MainWindow.xaml.cs
@@ -140,9 +140,12 @@ namespace UndertaleModTool
             else if (obj is UndertaleString str)
             {
                 string stringFirstLine = str.Content;
-                int stringLength = StringTitleConverter.NewLineRegex.Match(stringFirstLine).Index;
-                if (stringLength != 0)
-                    stringFirstLine = stringFirstLine[..stringLength] + " ...";
+                if (stringFirstLine is not null)
+                {
+                    int stringLength = StringTitleConverter.NewLineRegex.Match(stringFirstLine).Index;
+                    if (stringLength != 0)
+                        stringFirstLine = stringFirstLine[..stringLength] + " ...";
+                }
 
                 title = "String - " + stringFirstLine;
             }
@@ -167,7 +170,7 @@ namespace UndertaleModTool
                 Debug.WriteLine($"Could not handle type {obj.GetType()}");
             }
 
-            if (title.Length > 64)
+            if (title?.Length > 64)
                 title = title[..64] + "...";
 
             return title;
--- a/UndertaleModTool/MainWindow.xaml.cs
+++ b/UndertaleModTool/MainWindow.xaml.cs
@@ -141,7 +141,7 @@ namespace UndertaleModTool
             {
                 string stringFirstLine = str.Content;
                 int stringLength = StringTitleConverter.NewLineRegex.Match(stringFirstLine).Index;
-                if (stringLength != -1)
+                if (stringLength != 0)
                     stringFirstLine = stringFirstLine[..stringLength] + " ...";
 
                 title = "String - " + stringFirstLine;
--- a/UndertaleModTool/MainWindow.xaml.cs
+++ b/UndertaleModTool/MainWindow.xaml.cs
@@ -1713,10 +1713,17 @@ namespace UndertaleModTool
                 }
             }
         }
-        private void CopyItemName(UndertaleNamedResource namedRes)
+        private void CopyItemName(object obj)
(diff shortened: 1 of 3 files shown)


Miepee:
- I dont think the tab shortening works (from the amount of tabs in there, it shouldnt even display #define matrix_world)
image
- there should be some default name on new string creation. this makes it very hard to edit if i misclicked somewhere else ("new string" would probably be enough)
image

VladiStep:
The first screenshot is "Strings" list, so it have 256 characters limit

Miepee:
Ah, whoops.

UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -167,8 +170,17 @@ public static string GetTitleForObject(object obj)
...
 
-            if (title.Length > 64)
-                title = title[..64] + "...";
+            if (title is not null)
+            {
+                // "\t" is displayed as 8 spaces.
+                // So, if the title contains "\t", replace them with spaces,
+                // in order to properly shorten the title.
+                if (title.IndexOf('\t') != -1)
+                    title = title.Replace("\t", "        ");
```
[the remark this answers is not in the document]
VladiStep: I assumed that it will create another string even if there's no occurrences, but now I see that it will not.

UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -1730,7 +1734,7 @@ private void CopyItemName(object obj)
 
             if (obj is UndertaleNamedResource namedRes)
                 name = namedRes.Name?.Content;
-            else if (obj is UndertaleString str)
+            else if (obj is UndertaleString str && str.Content?.Length > 0)
```
Miepee, line 1737: Why check for the length here, if title converter already does it?

UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -1914,6 +1918,8 @@ private void MenuItem_Add_Click(object sender, RoutedEventArgs e)
                     (obj as UndertaleNamedResource).Name = new UndertaleString(notDataNewName); // not Data.MakeString!
                 }
             }
+            else if (obj is UndertaleString str)
+                str.Content = "";
```
Miepee, line 1922: Would prefer if this would be set to something like "new string" instead of being a literal empty string.

UndertaleModTool/MainWindow.xaml.cs
VladiStep: So it wouldn't copy "(empty string)", as it's useless.

UndertaleModTool/MainWindow.xaml.cs
VladiStep: If it will do that, then you'll have to erase "new string" each time when you add the string.

Miepee: Which isn't a huge deal. Other stuff, like adding new code/script entries, behave similar.

Miepee: Well, currently when I add a new string, it is added as an empty string, which is suboptimal.
image

It would be better if it would start with something else. So i'm pretty sure I am talking about the content here, not name.

VladiStep: But how is it better?
When you are creating the string, you should already have a concept what to write to it.
Having "new string" there will lead only to unnecessary actions - selecting all and pressing Delete.

VladiStep: And now it's not a problem to have an empty string for some time, as you can properly click to it.

Miepee: You didn't get what I was trying to say.
Yes, having a literal empty string be represented by something else is great and should stay

What I'm asking for, is that when a new name gets *added*, that an actual placeholder string gets created. Currently the string does get created as an empty string, and only gets *displayed* differently, because of the precaution from above. But instead a string similar to "newstring1" or something should get created, similar to what happens if you create a new code/script/object/room. You don't just get a new code entry with an empty name, the program goes out of its way to generate a placeholder like "code1234" so its differentiable. And something like that should happen here as well.

Miepee (approved):

PR #983, opened 2022-06-16
Allow entries from the left list to be middle-mouse button'ed

### LINKED ISSUE #981: Make MMB on assets open them in new tabs

UI/UX

Would be nice if Middle-Mouse-Button click on assets would open them in new tabs just like RMB -> open in new tab does.

Miepee:
## Description
This PR makes it so that entries from the left list can be middle-mouse button'ed, which opens them in another tab.
Fixes #981 

### Caveats
- If you attempt to MMB the headers, they won't get highlighted, which may look a little weird
- MMB'ing results in the treeview centering the item that was clicked, which may look a little weird.


VladiStep:
Besides the caveats you've described, there are more issues:

1. The `MouseDown` handler is `MainTree_OnMouseDown()`, but it's assigned to every child of `TreeViewItem`.
2. It doesn't check the `UndertaleRoom` layers order that you are opening.
3. It doesn't open `UndertaleString` items.

Miepee:
> The MouseDown handler is MainTree_OnMouseDown(), but it's assigned to every child of TreeViewItem.

It is supposed to be assigned to the TreeViewItem children. The name is completely wrong, because it was from when I was still figuring things out which I later forgot to update xP

Miepee:
Closed in favour of #988

PR #988, opened 2022-06-17
Open asset in a new tab with MMB.

### LINKED ISSUE #981: Make MMB on assets open them in new tabs

VladiStep:
## Description
Any asset from the asset list (contains "Strings", "Code", "Sprites", etc.) can be opened in a new tab with a **m**iddle **m**ouse **b**utton.

### Notes
Obsoletes #983 
Closes #981

### DIFF
--- a/UndertaleModTool/MainWindow.xaml.cs
+++ b/UndertaleModTool/MainWindow.xaml.cs
@@ -1531,7 +1531,7 @@ namespace UndertaleModTool
         {
             if (e.NewValue is TreeViewItem)
             {
-                string item = (e.NewValue as TreeViewItem)?.Header?.ToString();
+                string item = (e.NewValue as TreeViewItem).Header?.ToString();
 
                 if (item == "Data")
                 {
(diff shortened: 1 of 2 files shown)


UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -1567,7 +1567,29 @@ private void MainTree_MouseDoubleClick(object sender, MouseButtonEventArgs e)
 
             OpenInTab(Highlighted);
         }
+        private void MainTree_MouseDown(object sender, MouseButtonEventArgs e)
+        {
+            if (e.ButtonState == MouseButtonState.Pressed && e.ChangedButton == System.Windows.Input.MouseButton.Middle)
+            {
+                DependencyObject obj = MainTree.InputHitTest(e.GetPosition(MainTree)) as DependencyObject;
+                if (obj is not TextBlock)
+                    return;
```
Miepee, line 1576: comment as to why this is necessary would be nice. You mentioned it in the discord, but not everyone would know it.

VladiStep: What exactly is necessary?
`InputHitTest()` instead of `VisualTreeHelper.HitTest()` ?

Miepee: that, and what the hitThat, and why one needs to hit test to begin with, cause wasnt that clear when i first saw it.
(Might also be my inexperience with wpf tho).

VladiStep: So, this will be enough?
```
// Get the clicked visual element by a mouse position (relatively "MainTree" itself).
// It should be "TextBlock".
```

Miepee:
```
// Gets the clicked visual element by the mouse position (relative to "MainTree"). 
```
Slight typo fix and removed second statement as `obj is not TextBlock` IMO is clear enough for that

VladiStep: Forgot to write about the method choice:
```
// Get the clicked visual element by a mouse position (relatively "MainTree" itself).
// It should be a "TextBlock".
// P. S.: "VisualTreeHelper.HitTest()" ignores visibility of the elements,
// and this led to "ghost" hits, if it's an empty space.
```

Miepee: Would personally change that to:
```
// Gets the clicked visual element by the mouse position (relative to "MainTree").
// This is used instead of "VisualTreeHelper.HitTest()" because that ignores the visibility of elements,
// which led to "ghost" hits on empty space.
```

PR #995, opened 2022-06-23
Texture size UI fields in the embedded texture editor.

### LINKED ISSUE #913: Feature: Display Size of Embedded Texture page

UI/UX

Having the size of the Texture page be displayed would be very useful, as you otherwise need to extract the texture in order to know it's size

VladiStep:
## Description
Added width and height UI fields to the embedded texture editor.

Closes #913


Miepee (approved):
Not a fault of the PR, but I'd wish disabled textboxes in WPF would stand out a little more.

PR #998, opened 2022-06-24
Make Settings window more compact

Miepee:
## Description
This PR makes the settings window more compact.

### Caveats
None

### Notes
Change 3 of #931 was seperated here, since Tek does not have the time to maintain the other PR.


VladiStep:
Everything is fine, but I think that window should have fixed height (so you can't resize it vertically).

Miepee:
Am personally not a fan of locking resizing, regardless of the circumstances.
Would've somewhat understood the motivation for width, but with height, it's personally reasonable to navigate it with scrollbars. Should I add them instead?

VladiStep:
Yeah, it will be better.
But make sure that it would be invisible if all content is visible.

VladiStep (approved):

PR #1000, opened 2022-06-24
Remove GMS 2.3 checks for profile mode and GMS 2.3 warning upon loading

Grossley:
Remove GMS 2.3 checks for profile mode and GMS 2.3 warning upon loading

## Description
Remove GMS 2.3 checks for profile mode and GMS 2.3 warning upon loading.

### Caveats
I might have missed some checks for profile mode, and there may or may not be reason for that GMS 2.3 warning to still exist.

### Notes
Long overdue, similar warnings need to be removed from the scripts.


VladiStep:
Besides what @Miepee have noticed, I don't see any issues.

PR #1001, opened 2022-06-25
Make room grid be in front, instead of in back

Miepee:
## Description
Makes the room grid to be at the very front, instead of at the back

### Caveats
None

### Notes
This is 1) of #931


VladiStep (approved):

VladiStep:
I've forgot about "ExportAllRoomsToPng.csx" - on line 38, change "Draw background grid?" to "Draw room grid?" (or something similar).

Issue #827, opened 2022-04-09
Tab issues


Miepee:
- [x] back button / history is not working properly (proposed solution: Add opened tab to the history (probably, `Stack`) on tab switching, and remove closed tab from it. (Addressed on #831)
- [x] tab scroll bar should only be visible if its necessary, not always shown (Addressed on #831)
- [x] make prettier close icons.

VladiStep:
+ _"tabs scroll bar sometimes overlaps data editor"_

NyakoFox:
I said this in the discord server but I'll repeat it here

The way tabs are handled right now is, they're like shortcuts to editors. That means if you're in a tab, and you switch to another tab, the tab you just clicked on will load what it's supposed to show. Most programs with tabs don't do this because that's very slow and comes with a whole bunch of issues.

An example of an issue from this is, try writing code, not compiling, and switching to another tab. This use-case is very common, as referencing other tabs while you're writing code happens quite a lot, at least for me.

But since you switched tabs, it throws away whatever the tab you were just on was. Meaning that your changes were never compiled, and the code you just wrote was thrown away. And don't worry, next time you click on that tab, it'll decompile the code again!

This is a really bad method of handling tabs, and I know it won't be easy to change the behavior, but it _really_ needs to be done.

https://user-images.githubusercontent.com/87289390/162821612-e0e28457-452c-4b76-ba1e-eb918e4c55d3.mp4

Jacky720:
- [ ] Tabs persist across data.win loads, appearing functional but referencing the wrong UndertaleData (which is not cleared from memory). Tabs should either close automatically on loading, or attempt to open the corresponding asset in the new Data.

NyakoFox:
- [ ] Tabs should be able to be reordered, preferably with drag and drop.

Miepee:
When a tab's title is too long, this happens:
![image](https://user-images.githubusercontent.com/38186597/166147014-8ec63073-db18-42be-9f7c-e554d7ff8bd1.png)

NyakoFox:
Here's the ideal tab functionality that I suggested on the server:

- The list on the left will open things in a new tab, _**unless**_ you right click, and choose "**Open in current tab**".
- For everything _inside_ of a tab which would _open_ a new tab, for example, clicking on a code reference inside of an object's tab, it'll be opened in the current tab, **_unless_** you either middle click, or right click and choose "**Open in new tab**".

[...]

PR #1003, opened 2022-06-25
Per-tab history.

VladiStep:
## Description

1. **Tab history is now stored per-tab.**
2. **Tab history can be navigated through in both directions** - added "Forward" button; back and forward mouse buttons work.
3. `Tab.OpenedObject` is renamed to `Tab.CurrentObject`.
4. Added `MainWindow.OnPropertyChanged()`.
5. Fixed object re-opening after closing all tabs with "Close all tabs".
6. Fixed bug that in some cases broke the tab close button after closing other tabs with "Close other tabs".
7. "Open in new tab" option always opens a object in a new tab.
8. Asset items do always open in the current tab on double-click.
9. `ChangeSelection()` (used by `UndertaleObjectReference` and more) opens an object in the current tab.
10. The history navigation button(s) icon opacity is reduced.

### Notes
The  last sub-issue of #827 can be closed.

### DIFF
--- a/UndertaleModTool/MainWindow.xaml.cs
+++ b/UndertaleModTool/MainWindow.xaml.cs
@@ -1846,7 +1846,10 @@ namespace UndertaleModTool
                 for (int i = 0; i < ClosedTabsHistory.Count - 1; i++)
                 {
                     if (ClosedTabsHistory[i] == ClosedTabsHistory[i + 1])
+                    {
                         ClosedTabsHistory.RemoveAt(i);
+                        i--;
+                    }
                 }
 
                 // remove all deleted object occurrences from all tab histories
@@ -1856,7 +1859,9 @@ namespace UndertaleModTool
                     {
                         if (tab.History[i] == obj)
                         {
-                            tab.HistoryPosition--;
+                            if (i < tab.HistoryPosition)
+                                tab.HistoryPosition--;
+
                             tab.History.RemoveAt(i);
                         }
                     }
@@ -1866,8 +1871,11 @@ namespace UndertaleModTool
                     {
                         if (tab.History[i] == tab.History[i + 1])
                         {
-                            tab.HistoryPosition--;
+                            if (i < tab.HistoryPosition)
+                                tab.HistoryPosition--;
+
                             tab.History.RemoveAt(i);
+                            i--;
                         } 
                     }
                 }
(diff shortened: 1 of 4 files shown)


UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -55,20 +56,38 @@ public class Tab : INotifyPropertyChanged
         public static readonly BitmapImage ClosedIcon = new(new Uri(@"/Resources/X.png", UriKind.RelativeOrAbsolute));
         public static readonly BitmapImage ClosedHoverIcon = new(new Uri(@"/Resources/X_Down.png", UriKind.RelativeOrAbsolute));
 
-        // it's actually used, but when it will be compiled (see the note in "MainWindow" class)
-        #pragma warning disable CS0067
+        private static readonly MainWindow mainWindow = Application.Current.MainWindow as MainWindow;
```
Miepee, line 59: Why is a reference to itself necessary?
Can't you just use `this`?

UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -189,10 +208,55 @@ public static string GetTitleForObject(object obj)
...
+            bool objString = obj is UndertaleString;
+
+            if (prevObj is not null)
+            {
+                bool pObjNamed = prevObj is UndertaleNamedResource;
+                bool pObjString = prevObj is UndertaleString;
+
+                // If both objects have the same type (one of above)
+                // or both objects are not "UndertaleNamedResource",
+                // then there's no need for changing the binding
```
Miepee, line 234:
```suggestion
                // then there's no need to change the binding
```

UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -3265,20 +3390,23 @@ private void OpenInTab(object obj, bool isNewTab = false, string tabTitle = null
...
-                    else
-                        tabSwitched = false;
+                            return;
+                        }
+                        else
+                            tabSwitched = false;
 
-                    break;
+                        break;
+                    }
```
Miepee, line 3409: You can invert the if check to reduce nesting

UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -1741,22 +1833,44 @@ private void DeleteItem(UndertaleObject obj)
                     Data.GMLEditedBefore?.Remove(codeName);
                 }
 
-                CloseTab(obj);
+                while (CloseTab(obj)) ;
```
Miepee, line 1836: CloseTab returns `void`, how does this while loop function?

Miepee:
Some notes while testing:
- maybe make the buttons in full greyscale, to better show that they're disabled?
- pretty sure we had this discussion before and realize i'm wrong, but IMO now that i toy around with it, I feel like when something should get opened in the current tab, it should always do that, and do *not* switch to earlier occurences of the same object

UndertaleModTool/MainWindow.xaml.cs
VladiStep: Check the line 3519 - `CloseTab(obj)` overload returns boolean.

Miepee: Missed that. That is *definitely* confusing though, since all other close tab functions only return void

UndertaleModTool/MainWindow.xaml.cs
VladiStep: `Tab` class is separate from `MainWindow`, so `this` will return the `Tab` instance.

UndertaleModTool/MainWindow.xaml.cs
VladiStep: What the if check exactly?
`if (Tabs[i].CurrentObject == obj)` ?

VladiStep:
> * maybe make the buttons in full greyscale, to better show that they're disabled?

I've tried that with adding `FormatConvertedBitmap` with `DestinationFormat="Gray8"` as a resource (so it won't be recreated on each button disable), but that doesn't work.
So I'll just reduce the opacity to `0.2` - this also seems to work.

UndertaleModTool/MainWindow.xaml.cs
Miepee: Ah, missed that context from the web UI. Really need a way to get PR diffs into an IDE...

UndertaleModTool/MainWindow.xaml.cs
Miepee: For clarification, the if check in question to get inverted was `if (!isNewTab)`

Miepee:
You missed the opacity for the second arrow
image

Miepee (approved):
Aside from the few small issues here and there, this was done surprisingly quick!

PR #1022, opened 2022-07-01
Bump version to 0.5

Miepee:
## Description
Bumps the version up to 0.5.0.0


Grossley:
This is not the only place where "0.4.0.4" is referenced, please update those other areas.

Archie-osu:
The version doesn't follow the format. Please make it say 0.5.0.0 to preserve it.

PR #1029, opened 2022-07-02
Implement loading overhaul

Grossley:
## Description
1. Allow it to load partial data if it encounters an error while loading. This would normally result in the entire game being inaccessible. Saving is disabled in this mode, however everything else is available for read only access. This also allows data from the data file to be included in any issues made on GitHub.
2. Record warnings and errors to a log in executable directory for user reference.
3. Allow streamlined creation of issue to GitHub if it fails to load the game properly, including log, game information, MD5, and steam page (if it exists).

### Caveats
UndertaleModTool may possibly crash in an unexpected way due to partially loaded data. This may require more error handling in the new catch block in ReadUndertaleData or in other parts of the code. However, what else needs to be accounted for is unknown. No mechanism for clearing old logs has been implemented at this time in the tool itself. That will need to be done manually. 

### Notes
A fix has been implemented to prevent crashing of loading a room when no embedded texture pages exist.
TODO: Investigate bytecode 15 games that use GMS2.

### DIFF
--- a/UndertaleModTool/MainWindow.xaml.cs
+++ b/UndertaleModTool/MainWindow.xaml.cs
@@ -1004,10 +1004,13 @@ namespace UndertaleModTool
             DisposeGameData();
             Highlighted = new DescriptionView("Welcome to UndertaleModTool!", "Double click on the items on the left to view them!");
             OpenInTab(Highlighted);
+            String current_date_and_time = DateTime.Now.ToString("_\\M\\D\\Y_\\H\\M\\S_MM-dd-yyyy_HH-mm-ss");
 
             Task t = Task.Run(() =>
             {
                 bool hadWarnings = false;
+                bool loadedCorrectly = true;
+                Exception loadingException = null;
                 UndertaleData data = null;
                 try
                 {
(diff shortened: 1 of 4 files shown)


Miepee (changes requested):
As mentioned on Discord, I am very displeased with this PR.
We were discussing earlier, that this feature is low priority right now, needs its implementation properly thought out and should be ignored until the feature freeze for 0.5.0.0 is over.

You went ahead, and made your own Pull Request, disregarded the suggestions we told you were likely flawed, and are now pushing for this to get merged as soon as you can.

This is also not the first time, that you have shown a lack of cooperating as a team, and I don't know the reasons why you keep acting that way, but it would be helpful if you could stop that kind of behavior in the future.

UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -1040,9 +1115,15 @@ private async Task LoadFile(string filename, bool preventClose = false)
...
                             CanSafelySave = false;
                         }
-                        else if (hadWarnings)
+                        else if (!(data.LoadedCorrectly))
+                        {
+                            this.ShowError("Something went wrong while loading. Data loss WILL occur when trying to save, probably to a severe degree. As a result, saving has been disabled.\n\nYou may still view the data file loaded, but there is likely to be missing data and unexpected errors may occur.\n\nPlease see log" + current_date_and_time + ".txt for more details regarding what errors happened during loading.", "Loading problems");
+                            CanSave = false;
+                            CanSafelySave = false;
+                        }
+                        else if (hadWarnings && data.LoadedCorrectly)
```
Jacky720, line 1124: Isn't `data.LoadedCorrectly` redundant here because it's an else if?

Issue #929, opened 2022-05-24
Display proper GM version in the window title


Archie-osu:
### Which component should be improved?

UI/UX

### Describe your feature suggestion in more detail

Currently, even the latest DELTARUNE Chapter 2 builds show up as GMS 2.0.0.0 - I understand this is the version that UMT reads directly from the data.win, but given we can identify if a game uses 2022.x accurately, it should in my opinion be displayed instead.

Even if we can't identify all of the versions separately (as maybe the WAD format didn't change), the window title should include at least an approximation of what the version might be (think ``GM2022.3+``).

Miepee:
Personally think that the approximate version should be shown in General Info (or in a another new tab?) rather than the title along with all the other extra stuff that currently isn't shown like `ShortCircuit` or `UseQoiFormat`.
Is there any reason why the GMS version and bytecode are shown in the title to begin with? Easier help for people posting screenshots I assume?

Archie-osu:
In my opinion, the data.win version is less useful than the approximation - the approximation being in the window title would help with easily spotting which GM versions the tool's having issues with.

I like your idea of showing the hidden variables in a separate tab, and personally think the (now not-updated) data.win version should be moved there.

Miepee:
> being in the window title would help with easily spotting which GM versions the tool's having issues with.

Well, if a game fails to load, then the window title won't even get a chance to update. If a game fails to save or other errors occur, then I don't think its that hard to ask for people to send what the variable is by looking at a certain tab. 
Alternatively, maybe put in a "copy troubleshoot information" button somewhere? Have seen some programs do that.

> personally think the (now not-updated) data.win version should be moved there.

The general info, options and language areas are all sorted by their respective chunks. Moving the version which is part of GEN8 out of the area that displays GEN8 info is not good UX.

I also think that this:
![image](https://user-images.githubusercontent.com/38186597/170456532-9e6532cd-0cbe-4b3c-bcc4-6cec2954aa8b.png)

Is much cleaner than this:
![image](https://user-images.githubusercontent.com/38186597/170456549-5c1e1c55-e6d7-48c7-8724-aca06fb1a02a.png)

Archie-osu:
> The general info, options and language areas are all sorted by their respective chunks
Why is this even the case? The tool doesn't allow these values to be edited, and as far as I know no-one has ever actually used them.
To alter their values, you'd have to use the console, at which point you don't even need the UI and can just use UMTLib itself.

[...]

PR #1047, opened 2022-07-12
Clean up version flags

### LINKED ISSUE #1045: GMS2 flag spam

Lib

Because GameMaker no longer updates the version number stored in data.win, the mod tool currently uses a series of 7 (soon to be 8 with #1043) boolean values: GMS2_3, GMS2_3_1, GMS2022_1, etc. Having so many variables is largely redundant and makes it more difficult to solve #929. These, and all references to such, should be condensed into one value; perhaps a version enum, or expanding the [existing version functions](https://github.com/krzys-h/UndertaleModTool/blob/master/UndertaleModLib/UndertaleData.cs#L483).

Jacky720:
## Description
- This PR removes the `UndertaleData.GMS2_2_2_302`, `UndertaleData.GMS2_3`, `UndertaleData.GMS2_3_1`, `GMS2_3_2`, `GMS2022_1`, `GMS2022_2`, `GM2022_3`, and `GM2022_5` booleans in favor of creating an `UndertaleGeneralInfo.GMSVersions` enum, as well as `UndertaleGeneralInfo.GMS2Version` using said enum. Closes #1045.
- Additionally, `UndertaleData.IsVersionAtLeast()` has been updated to accept these newer versions, and `UndertaleData.SetGMS2Version()` has been added to easily change the value of `GMS2Version`.
- All existing references to the previous booleans have been changed to use the above functions, as well as several checks for `reader.undertaleData.GeneralInfo.Major >= 2` being simplified to the existing `UndertaleData.IsGameMaker2()` function.
- Finally, `UndertaleGeneralInfo.ToString()` has been updated to read from `GMS2Version` and thus display a more accurate version in the title bar, helping with #929: 
image

### Caveats
- Breaks any external references to the booleans that are not part of the project or that I missed.
- The version displayed is inaccurate for GM2022.6, as well as versions between GMS2.0 and GMS2.2.2.302 due to the limited nature of the pre-existing flags and infrequency of changes to the data.win format. This may be fixed by adding special clarification (e.g. "GMS2.0-2.2.0") to the display.

### Notes
N/A


UndertaleModLib/UndertaleData.cs
```diff
@@ -490,19 +485,38 @@ private bool TestGMS1Version(uint stableBuild, uint betaBuild, bool allowGMS2 =
...
-        public bool IsVersionAtLeast(uint major, uint minor, uint release, uint build)
+        public bool IsVersionAtLeast(uint major, uint minor = 0, uint release = 0, uint build = 0)
         {
-            if (GeneralInfo.Major != major)
-                return (GeneralInfo.Major > major);
+            uint gameMajor = GeneralInfo.Major;
+            uint gameMinor = GeneralInfo.Minor;
+            uint gameRelease = GeneralInfo.Release;
+            uint gameBuild = GeneralInfo.Build;
+            switch (GeneralInfo.GMS2Version)
```
Miepee, line 494: Small comment for this would be nice, something along the lines of "if gms2version is set, we readjust major/minor/release/build based on those."

Miepee (commented):
Haven't checked the actual replacements yet, only the new approach.

UndertaleModLib/Models/UndertaleGeneralInfo.cs
```diff
@@ -380,6 +432,8 @@ public void Serialize(UndertaleWriter writer)
     /// <inheritdoc />
     public void Unserialize(UndertaleReader reader)
     {
+        if (GMS2Version < GMSVersions.GMS2_3 && reader.GMS2_3)
+            GMS2Version = GMSVersions.GMS2_3;
```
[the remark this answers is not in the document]
Jacky720: It's more a general safety check, as applied elsewhere. If the version has (somehow, I think in practice this is the first enum-affecting check to run) already been set to _greater_ than 2.3, we don't want to roll it back.

If you mean that it would never run because the reader should have already set it, it can't do that because the reader value is set at the start of reading but the GeneralInfo hasn't been created yet to store it. So it patiently bides its time until the GeneralInfo _is_ parsed, then sets it right at the start there.

UndertaleModLib/UndertaleData.cs
```diff
@@ -482,6 +437,46 @@ private bool TestGMS1Version(uint stableBuild, uint betaBuild, bool allowGMS2 =
...
         }
 
+        /// <summary>
+        /// Sets the GMS2+ version flag in GeneralInfo.
+        /// </summary>
+        /// <param name="major">The major version.</param>
+        /// <param name="minor">The minor version.</param>
+        /// <param name="release">The release version.</param>
+        /// <param name="build">The build version.</param>
+        public void SetGMS2Version(uint major, uint minor = 0, uint release = 0, uint build = 0)
```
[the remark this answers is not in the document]
Jacky720: `UndertaleGeneralInfo.GMSVersions.GMS2_2_2_302` qualifies as too much boilerplate, IMO.

Most invalid versions are actually just versions that don't have any distinguishing characteristics in the data file, but I see what you mean about major versions in particular. I can add a `throw` there.

UndertaleModLib/UndertaleData.cs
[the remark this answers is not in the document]
Jacky720: > why not

(mostly because I didn't know you could give enums attributes like that)

UndertaleModLib/UndertaleData.cs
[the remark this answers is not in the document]
Miepee: > UndertaleGeneralInfo.GMSVersions.GMS2_2_2_302 qualifies as too much boilerplate, IMO.

AFAIK, you don't have give the full name like that, `GMSVersions.GMS2_2_2_302` should be valid too, but may need some specific usings.

Jacky720:
To-do:
- [ ] Resolve all these suggestions.
- [ ] Confirm that more recent commits haven't broken anything.
- [ ] Realize I can move the `reader.AllChunkNames.Contains("SEQN")` check into GeneralInfo and do away with `reader.GMS2_3`.
- [ ] Add 2022.8 support.
- [ ] Say, are there any other new-ish chunks that can be used for version identification since 2.3? When was FEDS added?

UndertaleModLib/UndertaleData.cs
Jacky720: Added "If GMS2Version is set, we increase the version values to match."

UndertaleModLib/UndertaleData.cs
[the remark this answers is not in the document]
Jacky720: Added error: "Attempted to set a version of GameMaker {major} using SetGMS2Version". Not going to resolve this in case it's still possible to get in those usings and make it pretty.

UndertaleModLib/Models/UndertaleGeneralInfo.cs
```diff
@@ -478,7 +532,10 @@ private long GetInfoNumber(long firstRandom, bool infoTimestampOffset)
     /// <inheritdoc />
     public override string ToString()
     {
-        return DisplayName + " (GMS " + Major + "." + Minor + "." + Release + "." + Build + ", bytecode " + BytecodeVersion + ")";
+        if (Major == 2)
+            return DisplayName + " (" + Enum.GetName(typeof(GMSVersions), GMS2Version).Replace("_", ".") + ", bytecode " + BytecodeVersion + ")";
+        else
```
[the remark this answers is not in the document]
Jacky720: That's just my preference. Keeping all returns on the same level.

UndertaleModLib/Models/UndertaleGeneralInfo.cs
```diff
@@ -305,6 +305,58 @@ public enum FunctionClassification : ulong
     /// </summary>
     public bool InfoTimestampOffset { get; set; } = true;
 
+    /// <summary>
+    /// Enum for versions past GMS2.
+    /// </summary>
+    public enum GMSVersions
```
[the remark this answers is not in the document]
Jacky720: Renamed and added "GameMaker no longer updates the version constants in General Info, so it falls to UTMT to detect these versions."

UndertaleModLib/UndertaleData.cs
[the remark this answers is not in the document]
Jacky720: Who knew that using an enum at all was just like 20 extra lines? I just stole the regular GeneralInfo versions and amended serialization to max out at 2.0.0.0, seems much cleaner.

UndertaleModLib/UndertaleData.cs
```diff
@@ -289,26 +289,6 @@ public class UndertaleData : IDisposable
...
 
-        /// <summary>
-        /// Whether the data file is from version GMS2.2.2.302
-        /// </summary>
-        public bool GMS2_2_2_302 = false;
-
-        /// <summary>
-        /// Whether the data file is from version GMS2.3
-        /// </summary>
-        public bool GMS2_3 = false;
```
VladiStep, line 300: Revert this field removal, and update the description so it's clear that only this field is kept because of performance.
(see my commentary at `IsVersionAtLeast()`)

VladiStep (changes requested):

VladiStep:
I suggest instead of setting `CompileContext.GMS2_3` on its creation, you should:
1) Make the variable static.
2) Set it in either `GeneralInfo.Unserialize()` or `UndertaleChunkFORM.Unserialize()`.

UndertaleModLib/Models/UndertaleGeneralInfo.cs
```diff
@@ -478,7 +510,33 @@ private long GetInfoNumber(long firstRandom, bool infoTimestampOffset)
     /// <inheritdoc />
     public override string ToString()
     {
-        return DisplayName + " (GMS " + Major + "." + Minor + "." + Release + "." + Build + ", bytecode " + BytecodeVersion + ")";
+        if (Major == 1)
+            return DisplayName + " (GMS " + Major + "." + Minor + "." + Release + "." + Build + ", bytecode " + BytecodeVersion + ")";
+        else
+        {
+            StringBuilder sb = new(DisplayName?.ToString() ?? "");
+            sb.Append(" (GMS ");
```
VladiStep, line 518:
```suggestion
            if (Major < 2022)
                sb.Append(" (GMS ");
            else
                sb.Append(" (GM ");
```

VladiStep: If it's "GM**S** 2022.**2**" but "GM 2022.**3**", then:
```suggestion
            if (Major == 2022 && Minor >= 3)
                sb.Append(" (GM ");
            else
                sb.Append(" (GMS ");
```

Jacky720: Ah, but then you would have to add `|| Major > 2022`. 2023.1 is already out.

I'll commit my version, if you please.

VladiStep: Oh, yeah, I forgot.
But there's still no need for `||` - just make it `if (Major >= 2022 && Minor >= 3)`

VladiStep: Oh

UndertaleModLib/UndertaleData.cs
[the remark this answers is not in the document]
Jacky720: Marking as resolved after the rework to use the existing version variables.

UndertaleModLib/UndertaleChunks.cs
```diff
@@ -120,7 +117,7 @@ public class UndertaleChunkEXTN : UndertaleListChunk<UndertaleExtension>
 
         internal override void UnserializeChunk(UndertaleReader reader)
         {
-            if (reader.undertaleData.GMS2_3)
+            if (reader.undertaleData.IsVersionAtLeast(2, 3) && !reader.undertaleData.IsVersionAtLeast(2022, 6))
```
VladiStep, line 120: Is a 2022.6 check necessary?

UndertaleModLib/UndertaleChunks.cs
```diff
@@ -339,7 +336,7 @@ internal override void SerializeChunk(UndertaleWriter writer)
 
         internal override void UnserializeChunk(UndertaleReader reader)
         {
-            if (reader.undertaleData.GeneralInfo?.BytecodeVersion >= 17)
+            if (!reader.undertaleData.IsVersionAtLeast(2022, 2) && reader.undertaleData.GeneralInfo?.BytecodeVersion >= 17)
```
VladiStep, line 339: (the same question as above)

UndertaleModLib/UndertaleChunks.cs
Jacky720: Yes. That code block is to set the version to 2022.6, so if the versions n has already been set higher it would be redundant.

VladiStep:
Everything seems fine.
Just make sure that there are no scripts that you've missed.

VladiStep (approved):

Jacky720:
Rebasing to update the change from #1048...

PR #1085, opened 2022-10-05
Attempt fixes to reported erring scripts

Jacky720:
## Description
Remove Chapter 2 errors from DeltaMILK, fix other errors on DeltaHATE caused by changing variable names. Make it handle games that aren't named "UNDERTALE" a bit more gracefully, too.

### Caveats
DeltaHATE still doesn't work on Chapter 2.

### Notes
N/A

### DIFF
--- a/UndertaleModTool/Scripts/Builtin Scripts/DeltaHATE.csx
+++ b/UndertaleModTool/Scripts/Builtin Scripts/DeltaHATE.csx
@@ -2,6 +2,7 @@
 
 EnsureDataLoaded();
 
+// Is actually incompatible. Something broke when I (Space Core) tried to combine the UT and SURVEY_PROGRAM code.
 if (Data?.GeneralInfo?.DisplayName?.Content.ToLower() == "deltarune chapter 1 & 2")
 {
     ScriptError("Error 0: Incompatible with the new Deltarune Chapter 1 & 2 demo");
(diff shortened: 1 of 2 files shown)


UndertaleModTool/Scripts/Builtin Scripts/DeltaHATE.csx
```diff
@@ -116,7 +117,16 @@ void StringSwap(int n, int k)
...
 
-if (Data.GeneralInfo.Filename.Content.ToLower().Contains("undertale"))
+string GameName = Data.GeneralInfo.DisplayName.Content.ToLower();
+bool deltamode = false;
+if (GameName.Contains("undertale") || GameName.Contains("nxtale"))
+    deltamode = false;
+else if (GameName.Contains("survey_program"))
+    deltamode = true;
+else
+    deltamode = ScriptQuestion("Is this Deltarune Chapter 1 or a mod thereof?") ? true : false;
```
Miepee, line 127: `ScriptQuestion` already returns a bool, why do a `? true : false` after it?

Jacky720: Somehow I was unsure it was a bool, but was sure it could be used in a ternary. In retrospect, of course it's a bool. I'll fix that.

PR #1116, opened 2022-11-13
Experimental

UtkuGARIP:
Can a feature be added to the **UndertaleModTool v0.4.0.4** program for language translators to count how many characters the written sentence consists of?

As you know, especially in Arcade style games, after a certain number of characters, a bottom line code (\n) is used.
The problem is that counting each character as manul becomes extra difficult.

Guidance lines with maybe 50% transparency can be added. Once you reach this line, you must move to the next line. to remind.

I hope I was able to explain.

image
<center>A game with a 32 word limit.</center>

A transparent line that we can move back and forth, like in this example, will suffice. _I certainly did not count the characters over and over the picture._


Dobby233Liu:
I think you want to open an issue instead of a pull request

UtkuGARIP:
> I think you should open an issue instead of a pull request

Thank you. I opened the [issue](https://github.com/krzys-h/UndertaleModTool/issues/1117).

PR #1172, opened 2023-02-10
Fix of the README download links.

VladiStep:
## Description
The "Downloads" section links are updated now.
Here's the preview:
image

### Caveats
Someone should update README on every new release.


VladiStep:
TIL that renaming branch closes a PR with that branch

PR #1191, opened 2023-02-27
Fix returns/multiple scripts in GMS2.3+

### LINKED ISSUE #900: Bug: the compiler removes return statements sometimes
### Describe the bug

a return at the end of an if statement with other code in it removes the return when compiled

### Reproducing steps

1. Open a script (not sure if it works on the new functions thing, definitely works on legacy scripts)
2. Put the following code at the beginning:
```gml
if(some_var) {
    do_something()
    return;
}
```
3. Click away to compile the code
4. Observe the return disappear

### Setup Details

1. commit 61d0f89
2. win10 latest
3. Will You Snail?

Jacky720:
## Description
GMS 2.3 scripts always produce an `exit` instruction, even for scripts lacking a `return`. UTMT previously interpreted this as a return anyway, and the workaround of ignoring returns on the top level block caused issue #900. (This closes #900).

Commit 60ea473 additionally adds support for changing the offset, length, and argument counts of anonymous script entries, and creating new ones by simply invoking `function`, during script compilation.

### Caveats
The new workaround takes the list of blocks and `Sort()`s its keys to peek at the next and identify the extraneous `exit`s. This may result in performance degradation for complex scripts with many blocks and many returns.

### Notes
N/A


UndertaleModLib/Decompiler/Decompiler.cs
```diff
@@ -2065,17 +2065,38 @@ internal static void DecompileFromBlock(DecompileContext context, Dictionary<uin
...
-                        */
-                        statements.Add(stmt);
+                        // 2.3 scripts add exits to every script, even those that lack a return
+                        // This detects that type of exit using the next block.
+                        Block nextBlock = null;
+                        if (DecompileContext.GMS2_3 && instr.Kind == UndertaleInstruction.Opcode.Exit)
+                        {
+                            List<uint> blockAddresses = new(blocks.Keys);
+                            blockAddresses.Sort();
+                            int nextBlockIndex = blockAddresses.IndexOf(block.Address ?? 0) + 1;
```
VladiStep, line 2075:
```suggestion
                            uint[] blockAddresses = blocks.Keys.ToArray();
                            Array.Sort(blockAddresses);
                            int nextBlockIndex = Array.IndexOf(blockAddresses, block.Address ?? 0) + 1;
```

VladiStep: (`uint[]` is a little bit lighter than `List<uint>`)

PR #1203, opened 2023-03-03
Bytecode v17 Bandaid for big Switch Statements in Decompiler.cs

Radix-Comet:
Haven't entirely figured out why but in v17 bytecode, some switch statements can break entirely and eat literal gigabytes of memory infinitely trying to determine the end of the switch statement. I have tested with existing and working scripts inside of Pizza-Tower and found no noticeable differences before or after the patch, and the entire switch statement seems in-tact. Not entirely sure, very strange!

### DIFF
--- a/UndertaleModLib/Decompiler/Decompiler.cs
+++ b/UndertaleModLib/Decompiler/Decompiler.cs
@@ -3556,7 +3556,13 @@ namespace UndertaleModLib.Decompiler
                         var x = caseEntries.ElementAt(i);
                         Block temp = x.Key;
 
-                        Block switchEnd = DetermineSwitchEnd(temp, caseEntries.Count > (i + 1) ? caseEntries.ElementAt(i + 1).Key : null, meetPoint);
+                        // RadixComet: New bytecode (Mostly from Pizza Tower it seems) hangs here infinitely for some scripts
+                        // RadixComet: So I'm adding a special exception that seems to bandaid the issue for now
+                        Block switchEnd = null;
+                        if (context.GlobalContext.Data.GeneralInfo.BytecodeVersion == 17)
+                            switchEnd = meetPoint;
+                        else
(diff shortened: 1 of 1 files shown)


Jacky720:
Are you sure it's bytecode 17? There have been a lot of GameMaker versions since then.

Radix-Comet:
That is what Pizza Tower identifies itself as if I'm not mistaken.

Jacky720:
Well, yes, it's just that this might be an issue with GMS2.2, GMS2.3, GMS2022.1, GMS2022.2, GM2022.3, GM2022.4, GM2022.5, GM2022.7, GM2022.8, GM2022.9, or... Well, probably not 2023.1, the tool doesn't support that at all at the moment. But the bytecode hasn't changed in ages, and making a fix that applies to everything bytecode 17 would reach back as far as Deltarune Chapter 1 (2018).

VladiStep:
> Well, yes, it's just that this might be an issue with GMS2.2, GMS2.3, GMS2022.1, GMS2022.2, GM2022.3, GM2022.4, GM2022.5, GM2022.7, GM2022.8, GM2022.9, or... Well, probably not 2023.1, the tool doesn't support that at all at the moment. But the bytecode hasn't changed in ages, and making a fix that applies to everything bytecode 17 would reach back as far as Deltarune Chapter 1 (2018).

Yeah, 2023.1 also has bytecode 17.
image

Radix-Comet:
> Well, yes, it's just that this might be an issue with GMS2.2, GMS2.3, GMS2022.1, GMS2022.2, GM2022.3, GM2022.4, GM2022.5, GM2022.7, GM2022.8, GM2022.9, or... Well, probably not 2023.1, the tool doesn't support that at all at the moment. But the bytecode hasn't changed in ages, and making a fix that applies to everything bytecode 17 would reach back as far as Deltarune Chapter 1 (2018).

If the bytecode hasn't changed in ages why does Pizza Tower seem to mess up? Is the current bytecode decompiler not 100%?

Jacky720:
I mean, the listed bytecode version hasn't changed. It still says 17 even though they've added all sorts of stuff.

PR #1316, opened 2023-04-11
Fixed Update Button fix

zivmaor:
there were a couple of problems with the previous fix: #1312
this should work.

### DIFF
--- a/UndertaleModToolUpdater/Program.cs
+++ b/UndertaleModToolUpdater/Program.cs
@@ -63,7 +63,7 @@ namespace UndertaleModToolUpdater
             File.Delete(basePath + "Update.zip");
 
             Console.WriteLine("Replacing files with update...");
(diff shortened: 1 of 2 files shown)


UndertaleModToolUpdater/Program.cs
```diff
@@ -63,7 +63,7 @@ static void Main(string[] args)
             File.Delete(basePath + "Update.zip");
 
             Console.WriteLine("Replacing files with update...");
-            MoveDirectory(basePath + "Update", appPath);
+            MoveDirectory(basePath + $"Update{Path.DirectorySeparatorChar}windows-latest", appPath);
```
VladiStep, line 66:
```suggestion
```

VladiStep: Because my PR reverts the change

VladiStep (approved):

Issue #1343, opened 2023-04-21
Loading "Abomi Nation" results in high RAM usage and "out of bounds" error in SPRT chunk


Grossley:
### Describe the bug

<img width="587" alt="2023-04-21 16_24_06-" src="https://user-images.githubusercontent.com/42724744/233749963-e4ca8192-d0b4-4808-9da4-57891cadb610.png">

### Reproducing steps

This occurs with "Abomi Nation".

### Setup Details

Windows 11, git 4336bdd

PR #1352, opened 2023-04-25
Fix YYSWF parsing on 2022.1 and beyond

nkrapivin:
## Description

Fixes issue(s): #1343 

### Caveats

This *assumes* that the TPEIndex system is used on 2022.1 and beyond since this is what I had, older GM versions might use TPEIndex too but I don't have enough .wins nor free time to accurately determine this.

Also YYSWF objects are not pooled since they might overlap, i.e. an object might start directly at the parent object's address. If the pool system handles this correctly now, NoPool can be removed.

### Notes

YYG if you are reading this please, for god's sake, version your damn changes correctly.

### DIFF
--- a/UndertaleModLib/Models/UndertaleSprite.cs
+++ b/UndertaleModLib/Models/UndertaleSprite.cs
@@ -1177,6 +1177,9 @@ public class UndertaleYYSWFGradientFillData : UndertaleObject
     /// <summary>
     /// Unknown purpose. Probably to accomodate for new texture formats.
     /// </summary>
+    /// <remarks>
+    /// Presumably present in GM 2022.1+.
+    /// </remarks>
     public int? TPEIndex { get; set; }
 
     /// <inheritdoc />
--- a/UndertaleModLib/Models/UndertaleSprite.cs
+++ b/UndertaleModLib/Models/UndertaleSprite.cs
@@ -1174,11 +1174,17 @@ public class UndertaleYYSWFGradientFillData : UndertaleObject
     public UndertaleYYSWFGradientFillType GradientFillType { get; set; }
     public UndertaleYYSWFMatrix33 TransformationMatrix { get; set; }
     public UndertaleSimpleList<UndertaleYYSWFGradientRecord> Records { get; set; }
+    /// <summary>
+    /// Unknown purpose. Probably to accomodate for new texture formats.
+    /// </summary>
+    public int? TPEIndex { get; set; }
 
     /// <inheritdoc />
     public void Serialize(UndertaleWriter writer)
     {
         writer.Write((int)GradientFillType);
+        if (TPEIndex is not null)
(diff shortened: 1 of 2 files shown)


UndertaleModLib/Models/UndertaleSprite.cs
```diff
@@ -1174,11 +1174,17 @@ public class UndertaleYYSWFGradientFillData : UndertaleObject
     public UndertaleYYSWFGradientFillType GradientFillType { get; set; }
     public UndertaleYYSWFMatrix33 TransformationMatrix { get; set; }
     public UndertaleSimpleList<UndertaleYYSWFGradientRecord> Records { get; set; }
+    /// <summary>
+    /// Unknown purpose. Probably to accomodate for new texture formats.
+    /// </summary>
+    public int? TPEIndex { get; set; }
```
VladiStep, line 1180:
```suggestion
    /// <summary>
    /// Unknown purpose. Probably to accomodate for new texture formats.
    /// </summary>
    /// <remarks>
    /// Presumably present in GM 2022.1+.
    /// </remarks>
    public int? TPEIndex { get; set; }
```

UndertaleModLib/UndertaleIO.cs
```diff
@@ -344,7 +344,9 @@ private bool ProcessCountExc(uint poolSize = 0)
             if (poolSize != 0 && poolSize != objectPool.Count)
             {
                 SubmitWarning("Warning - the estimated object pool size differs from the actual size.\n" +
-                              "Please report this on UndertaleModTool GitHub.");
+                              "Please report this on UndertaleModTool GitHub.\n" +
+                              "Unless you have a .win with SWF sprites,\n" +
+                              "since YYSWF objects are not pooled due to overlapping.");
```
VladiStep, line 349: Are you sure that there could be the pool size mismatch? All the SWF related classes are not in the pool, so is that really possible?

UndertaleModLib/Models/UndertaleSprite.cs
```diff
@@ -1651,6 +1661,10 @@ public class UndertaleYYSWFBitmapData : UndertaleObject
     public UndertaleYYSWFBitmapType Type { get; set; }
     public int Width { get; set; }
     public int Height { get; set; }
+    /// <summary>
+    /// Unknown purpose. Probably to accomodate for new texture formats.
+    /// </summary>
+    public int? TPEIndex { get; set; }
```
VladiStep, line 1667:
```suggestion
    /// <summary>
    /// Unknown purpose. Probably to accomodate for new texture formats.
    /// </summary>
    /// <remarks>
    /// Presumably present in GM 2022.1+.
    /// </remarks>
    public int? TPEIndex { get; set; }
```

VladiStep (changes requested):

PR #1380, opened 2023-05-13
put each sprite in seperate folders

H-A-M-G-E-R:
when there's only 1 texture in a sprite it doesn't create a folder


H-A-M-G-E-R:
but it breaks ImportGraphics.csx

H-A-M-G-E-R:
i made it compatible with ImportGraphics.csx

UndertaleModTool/Scripts/Resource Repackers/ImportASound.csx
```diff
@@ -211,7 +211,7 @@ if (dialogRet == DialogResult.OK)
             File        = Data.Strings.MakeString(fname),
             Effects     = 0,
             Volume      = 1.0F,
-            Pitch       = 1.0F,
+            Pitch       = 0,
```
VladiStep, line 214: I think this is a redundant change that you should undo.
```suggestion
            Pitch       = 1.0F,
```

VladiStep: You should undo this, and instead create a separate PR with only this particular change.

Miepee:
obsoleted in favour of the above mentioned PR

PR #1391, opened 2023-05-20
fix object reference field background being transparent when empty

### LINKED ISSUE #1383: The object reference field background is transparent when empty.
### Describe the bug

image

It should be like this instead:
image

### Reproducing steps

1. Open sprite editor.
2. Add new frame.

### Setup Details

1. UTMT version - https://github.com/krzys-h/UndertaleModTool/commit/029a8e52fc4f91a0790787e938a18e6bda7708d2

zivmaor:
## Description

Fix #1383.
The problem was that the hint for when the object reference field is empty was set as the background of the textbox via a label, and thus the background became transparant.
I don't know how this can be fixed with the hint still being in the background, but I fixed it by making the hint the text of the textbox. This should be ok in this context because the textbox is readonly.
### Caveats

I used a valueConverter in weird way to make this work. If that is not ok I can make it work without that via making the UndertaleObjectReference use INotifyPropertyChanged

### DIFF
--- a/UndertaleModTool/Controls/UndertaleObjectReference.xaml
+++ b/UndertaleModTool/Controls/UndertaleObjectReference.xaml
@@ -13,7 +13,6 @@
             <MenuItem Header="Open in new tab" Click="OpenInNewTabItem_Click"/>
             <MenuItem Header="Find all references" Click="FindAllReferencesItem_Click"/>
         </local:ContextMenuDark>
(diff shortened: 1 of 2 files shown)


zivmaor:
A better way to do this would probably be to replace the visual of the visual brush with a canvas/grid containing the label and a rectangle of the correct color (or just a background of the canvas). I have a version of this working but I've had to do some janky things to get it working

PR #1409, opened 2023-06-07
Create ImportTileset.csx

AwfulNasty:
## Description

Includes ImportTIleset script that works in the same way as ImportGraphics. Auto imports Tile IDs, a previously tedious process.
### Caveats

It feels odd to have a script that functions the same way as ImportGraphics, but also able to do tilesets.
### Notes

Maybe this script should be consolidated with ImportGraphics, but doing so would hide the change to people used to UMT. Cutting out the parts that import other assets might be best here?

### DIFF
--- a/UndertaleModTool/Scripts/Resource Repackers/ImportTileset.csx
+++ b/UndertaleModTool/Scripts/Resource Repackers/ImportTileset.csx
@@ -1,0 +1,728 @@
+// Texture packer by Samuel Roy
+// Uses code from https://github.com/mfascia/TexturePacker
+// Entirely based off of code from ImportGraphics.csx script
+// Extended to import tilesets with proper tile IDs
+
+using System;
+using System.IO;
+using System.Drawing;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using UndertaleModLib.Util;
+using System.ComponentModel;
+
+
+EnsureDataLoaded();
+
+bool importAsSprite = false;
+
+string importFolder = CheckValidity();
+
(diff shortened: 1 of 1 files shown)


UndertaleModTool/Scripts/Resource Repackers/ImportTileset.csx
```diff
@@ -0,0 +1,728 @@
...
+
+string importFolder = CheckValidity();
+
+string packDir = Path.Combine(ExePath, "Packager");
+Directory.CreateDirectory(packDir);
+
+string sourcePath = importFolder;
+string searchPattern = "*.png";
+string outName = Path.Combine(packDir, "atlas.txt");
+int textureSize = 2048;
```
AwfulNasty, line 29: Make this 16384 or some other larger number - the texture packer is fine with working down in size, but will just skip over tilesets larger than 2048x2048 pixels, which is a lot of tilesets.

H-A-M-G-E-R:
what if everyone has write access on a repo like a wiki

VladiStep:
You should undo [the "Room Editor Creation Code Support"](https://github.com/krzys-h/UndertaleModTool/pull/1409/commits/4957dfbfd5b7c3ff7b19afb19af9b08e1a083b94) commit and create a separate PR with that commit only.

Miepee:
@AwfulNasty can you please change the PR to remove the unrelated commit?

BenjaminUrquhart (changes requested):
Please remove this commit from the pull request as it is unrelated.


I am also unsure if this script requires updating to work with UTMT 0.8.2.0 but that can be managed later.

Issue #1466, opened 2023-08-27
Decompiler emits lots of redundant code


OliveIsAWord:
### Which component should be improved?

Decompiler

### Describe your feature suggestion in more detail

I found this tool when trying to extract assets from [Fazbear Entertainment: Storage](https://gamejolt.com/games/FazEntStorage/439239) and was impressed with the capabilities of the decompilation tool. However, the decompiled code was filled with dead code such as empty `if` blocks, redundant `with` blocks, and values assigned to a temporary variable before being immediately read once and then forgotten. This becomes a major problem for larger Codes, both in decompilation performance and readability of the decompiled code.

In the Code `gml_Object_Obj_tbmover_Alarm_1` (which controls Toy Bonnie's movement), the disassembly contains 1263 SLoC, but the decompilation takes multiple seconds and produces over 280,000 (!!) SLoC, many sections of which are indented so far that the lines wrap around the editor. I suspect that most of this indentation and perhaps a majority of these lines could be removed without major overhauls or major feature additions to decompilation.

As a shorter example, the Code `gml_Object_Obj_MenuBG_Alarm_1` currently decompiles to this:
```gml
var __b__;
with (Obj_NightSetter)
{
    __b__ = action_if_variable(SAVE, 0, 0)
    if __b__
    {
    }
    if __b__
        scr_loadgame()
    with (Obj_NightSetter)
    {
        __b__ = action_if_variable(SAVE, 1, 0)
        if __b__
        {
        }
        if __b__
            scr_savegame()
    }
}
```
If the dead code mentioned above could be algorithmically removed, this snippet might instead decompile to this much nicer code:
```gml
with (Obj_NightSetter)
{
    if action_if_variable(SAVE, 0, 0)
        scr_loadgame()
    if action_if_variable(SAVE, 1, 0)
        scr_savegame()
}
```
It's entirely possible that I'm missing obvious code refactors, or any of my suggested refactors are incorrect; this tool has been my near only exposure to GML.

Dobby233Liu:
FYI I think these stuff were generated automatically by gm from Drag and Drop code

Jacky720:
I'd like to mark this as won't fix on account of it being an accurate decompilation that is necessary for a 1:1 rewrite. Any attempt at optimizing out empty drag'n'drop should be made optional if possible, IMO.

Miepee:
would be cool to have, but definitely should be made optional.

Miepee:
[...]

PR #1467, opened 2023-08-27
Decompiler removes redundant `if`s and `with`s

OliveIsAWord:
## Description
A big step towards fixing https://github.com/krzys-h/UndertaleModTool/issues/1466.

### Caveats
This is unfinished! Also, I forgot to correctly format the code before i pushed it!

### Notes
Do tell me any changes I could make to bump this code up a letter grade. I do not have much experience at all in C#.


colinator27:
This will probably need to be updated to be part of (or using) the Underanalyzer decompiler, if this is still desired. The Underanalyzer decompiler does let you directly iterate over its tree output, which could be helpful. (Although I'm not sure if you can directly print a modified tree at the moment.)

PR #1481, opened 2023-09-15
Update ShowRoomName.csx to Work With Deltarune chapter 2.

EvaxTheFox:
## Description
The ShowRoomName script doesnt really work with deltarune. i have made it possible.
###Notes

Feel free to edit this (by maintainers)

### DIFF
--- a/UndertaleModTool/Scripts/Builtin Scripts/ShowRoomName.csx
+++ b/UndertaleModTool/Scripts/Builtin Scripts/ShowRoomName.csx
@@ -2,7 +2,7 @@
 // Adds room ID and name display under the debug mode timer display
 
 EnsureDataLoaded();
-ScriptMessage("Show room name and ID in debug mode\nby krzys_h, Kneesnap\n\n\nDeltarune Ch2 Patch By GattoDev");
+ScriptMessage("Show room name and ID in debug mode\nby krzys_h, Kneesnap\n");
 Data.Functions.EnsureDefined("room_get_name", Data.Strings); // required for Deltarune
 if (Data?.GeneralInfo?.DisplayName?.Content.ToLower() == "deltarune chapter 1 & 2")
 {
--- a/UndertaleModTool/Scripts/Builtin Scripts/ShowRoomName.csx
+++ b/UndertaleModTool/Scripts/Builtin Scripts/ShowRoomName.csx
@@ -1,25 +1,14 @@
+
 // Adds room ID and name display under the debug mode timer display
 
 EnsureDataLoaded();
-
+ScriptMessage("Show room name and ID in debug mode\nby krzys_h, Kneesnap\n\n\nDeltarune Ch2 Patch By GattoDev");
+Data.Functions.EnsureDefined("room_get_name", Data.Strings); // required for Deltarune
 if (Data?.GeneralInfo?.DisplayName?.Content.ToLower() == "deltarune chapter 1 & 2")
 {
-    ScriptError("Error 0: Incompatible with the new Deltarune Chapter 1 & 2 demo");
-    return;
-}
-else if (Data?.GeneralInfo?.DisplayName?.Content.ToLower() == "deltarune chapter 1&2")
-{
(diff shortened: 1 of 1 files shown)


UndertaleModTool/Scripts/Builtin Scripts/ShowRoomName.csx
```diff
@@ -1,30 +1,50 @@
+
 // Adds room ID and name display under the debug mode timer display
 
 EnsureDataLoaded();
-
+ScriptMessage("Show room name and ID in debug mode\nby krzys_h, Kneesnap\n\n\nDeltarune Ch2 Patch By GattoDev");
```
VladiStep, line 5: I think that the credits in this message are redundant

UndertaleModTool/Scripts/Builtin Scripts/ShowRoomName.csx
```diff
@@ -1,30 +1,50 @@
...
+Data.Functions.EnsureDefined("room_get_name", Data.Strings); // required for Deltarune
```
VladiStep, line 6: If this is required for Deltarune, then shouldn't you move that line below, to the block that checks for Deltarune?

VladiStep (changes requested):

UndertaleModTool/Scripts/Builtin Scripts/ShowRoomName.csx
VladiStep: Oh, nevermind, I see now that the message was there, and you just added yourself.

EvaxTheFox:
i will close this Pull request and try to make another hot piece of garbage next time

EvaxTheFox:
nevermind

VladiStep:
No, your PR seems fine, I think you only need to move that one line under the `if` block

UndertaleModTool/Scripts/Builtin Scripts/ShowRoomName.csx
EvaxTheFox: that was stupid and remoce that i believe so

EvaxTheFox:
> No, your PR seems fine, I think you only need to move that one line under the `if` block

its just so it doesnt get the naming wrong

EvaxTheFox:
checked and yes it does work with Undertale too so thats a win. hope me get accepte

UndertaleModTool/Scripts/Builtin Scripts/ShowRoomName.csx
Miepee: @EvaxTheFox is this only required for deltarune, or also required for undertale?
If the former, then it would be better if it'd be moved, if the latter, then don't mind me.

UndertaleModTool/Scripts/Builtin Scripts/ShowRoomName.csx
```diff
@@ -1,30 +1,50 @@
+
 // Adds room ID and name display under the debug mode timer display
 
 EnsureDataLoaded();
-
+ScriptMessage("Show room name and ID in debug mode\nby krzys_h, Kneesnap\n");
+Data.Functions.EnsureDefined("room_get_name", Data.Strings); // required for Deltarune
 if (Data?.GeneralInfo?.DisplayName?.Content.ToLower() == "deltarune chapter 1 & 2")
```
Miepee, line 7: 1) can you externalize the displayname, so that you're not lowering it everytime?
  something like 
  ```cs
  var displayName = Data?.GeneralInfo?.DisplayName?.Content.ToLower();
  if (displayName == "deltarune chapter 1 & 2") { ... }
  ```
  
2) The code for both for *all* blocks is the exact same. why have if conditions at all then?

Miepee:
@EvaxTheFox  any updates on the reviews?

Miepee (approved):

colinator27 (changes requested):
Needs to be changed back to use `global.debug` for Undertale again

colinator27 (approved):
Tested, and this seems to work!

colinator27:
Going to expedite getting this merged in, given this is a small script and I've now tested it.

PR #1485, opened 2023-09-19
A new "Use "Disassembly" code editor tab by default" setting.

### LINKED ISSUE #1475: Add a setting for default code editor tab/mode.

UI/UX

There should be a setting like "Open the "Disassembly" tab in the code editor by default".

VladiStep:
## Description
Closes #1475.

### Caveats
Now, if you had some opened code with "Disassembly" tab opened, it would switch to "Decompiled" tab on other code entry opening if the new setting is disabled.


VladiStep:
It's draft because it turns out that it causes the code editor to decompile the code twice in a row.

PR #1504, opened 2023-10-18
add relative path solution to script's #load

nhaar:
## Description
If attempting to use `#load` inside a csx script (which allows you to load another script, useful for file organization), and you supply a relative path, it will not work as one might believe, that is, it will use the relative path to the UndertaleModTool directory and not a path relative to the script. To fix this, the goal is so that whenever someone uses #load to a relative path, it will "hardcode" that with an absolute path so that it will work.

### Caveats
I can't see anything troublesome with this. I have tested this in a project and it works perfectly

### DIFF
--- a/UndertaleModTool/MainWindow.xaml.cs
+++ b/UndertaleModTool/MainWindow.xaml.cs
@@ -1,4 +1,4 @@
-﻿using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp.Scripting;
 using Microsoft.CodeAnalysis.Scripting;
 using Microsoft.CodeAnalysis.Scripting.Hosting;
@@ -2579,6 +2579,18 @@ namespace UndertaleModTool
         {
             string scriptText = $"#line 1 \"{path}\"\n" + File.ReadAllText(path);
             Debug.WriteLine(path);
+
+            // since attempting to load scripts with #load in a file will lead to it using the UTMT path,
+            // we can circumvent this by hardcoding the absolute path to the script directory
+            // in all instances of #load with a relative path
+            var scriptDir = Path.GetDirectoryName(path);
+            var matches = Regex.Matches(scriptText, @"(?<=^#load\s+"").*\.csx(?=""\s*$)", RegexOptions.Multiline);
+            foreach (Match match in matches)
+            {
+                if (Path.IsPathRooted(match.Value))
+                    continue;
+                scriptText = scriptText.Replace(match.Value, Path.Combine(scriptDir, match.Value));
+            }
 
             Dispatcher.Invoke(() => CommandBox.Text = "Running " + Path.GetFileName(path) + " ...");
             try


UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -2505,12 +2506,13 @@ public string ProcessException(in Exception exc, in string scriptText)
                 {
                     if (traceLine.TrimStart()[..13] == "at Submission") // only stack trace lines from the script
                     {
+                        // matches full path of the script file
+                        string sourceFile = Regex.Match(traceLine, @"(?<=in ).*\.csx(?=:line \d+)").Value;
```
VladiStep, line 2510:
```suggestion
                        string sourceFile = Regex.Match(traceLine, @"(?<=in ).*\.csx(?=:line \d+)").Value;
                        if (!File.Exists(sourceFile))
                            continue;
```

UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -2526,10 +2528,21 @@ public string ProcessException(in Exception exc, in string scriptText)
...
-                string[] scriptLines = scriptText.Split('\n');
-                string excLines = string.Join('\n', excLineNums.Select(n => $"Line {n}: {scriptLines[n].TrimStart(new char[] { '\t', ' ' })}"));
+                // read the code for the files to know what the code line associated with the stack trace is
+                Dictionary<string, List<string>> scriptsCode = new();
+                foreach (Tuple<string, int> scriptLineNum in loadedScriptLineNums)
+                {
+                    string scriptPath = scriptLineNum.Item1;
+                    if (!scriptsCode.ContainsKey(scriptPath))
+                    {
+                        string scriptCode = File.ReadAllText(scriptPath);
```
VladiStep, line 2540:
```suggestion
                        string scriptCode = null;
                        try
                        {
                            scriptCode = File.ReadAllText(scriptPath);
                        }
                        catch (Exception e)
                        {
                            return $"An error occurred while processing the exception text.\nError message - \"{e.Message}\"\nThe unprocessed text is below.\n\n" + excString;
                        }
```

UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -2526,10 +2528,21 @@ public string ProcessException(in Exception exc, in string scriptText)
...
+                        scriptsCode.Add(scriptPath, scriptCode.Split('\n').ToList());
+                    }
+                }
+
+                string excLines = string.Join('\n', loadedScriptLineNums.Select(pair => $"Line {pair.Item2} in script {pair.Item1}: {scriptsCode[pair.Item1][pair.Item2 - 1]}")); // - 1 because line numbers start from 1
```
VladiStep, line 2545:
```suggestion
                string excLines = string.Join('\n', loadedScriptLineNums.Select(pair => 
                {
                    string scriptName = pair.Item1.Split(Path.DirectorySeparatorChar)[^1];
                    return $"Line {pair.Item2} in script {scriptName}: {scriptsCode[pair.Item1][pair.Item2 - 1]}"; // - 1 because line numbers start from 1
                }));
```

VladiStep (changes requested):
I just made it safer, and also made it write only script names instead of full path.

nhaar:
I have accepted VladiStep's changes

As for Miepee's request, I ran a test using `RunUMTScript` to see if it would work the same, unfortunately it doesn't (besides the fact that it doesn't accept path relatives to the script itself). One of the things I have been using `#load` for is that it allows to pass variables/functions defined in the script:

```cs
#load "script.csx"

SomeFunction(); // from the script
```

As annoying as it is that it doesn't allow for properly declaring what you want to get from the script, it is still useful, and the alternative with the builtin method doesn't work unfortunately.

Let me know if it would still be the goal to have `RunUMTScript` take relative paths through this PR.

Miepee:
> As annoying as it is that it doesn't allow for properly declaring what you want to get from the script

Huh. Makes me wonder what would be needed to change that, but that dosesn't sound easy at first glance. Guess using #load it is then.
I'll try to get to reviewing the PR then soon:tm:.

Miepee (commented):
no real review but a) you seem to have compilation errors and b) since you only change the gui, this will mean that it will not work on CLi

nhaar:
Something must have gone wrong after I accepted the commits, I will properly review for that later and move what needs to be on the CLI

nhaar:
I fixed the compilation issue and added it to the CLI

I wasnt sure where to put the method that adds the "workaround", so I made a new file. You can probably organize it better than me, if you don't like that

colinator27:
As mentioned in the above PR I just linked, I looked further into this and found a solution that doesn't depend on regular expressions/replacing text in the script contents. That PR will probably supersede this one.

PR #1505, opened 2023-10-18
fix importing collision event files when the objects already exist

nhaar:
## Description
If you try using the `ImportGMLFile` for a collision event, for example, for a file named `gml_Object_obj_pipis_bullet_cone_Collision_obj_yheart_shot.gml`, and tried to link to the objects automatically, even if the objects in the event already existed, it would require you to type the object, despite it already knowing it. With this fix, this issue is removed.

### Caveats
In my tests, it worked perfectly. I'm not sure why the original code that caused this problem existed, so unless it had any special reason to be there, there should be no problems with this change.

### DIFF
--- a/UndertaleModTool/ImportCodeSystem.cs
+++ b/UndertaleModTool/ImportCodeSystem.cs
@@ -341,13 +341,6 @@ namespace UndertaleModTool
                                         Data.GameObjects.Add(gameObj);
                                     }
                                 }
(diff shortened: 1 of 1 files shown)


Miepee (approved):
not a fan of removing untested code, but since it seems to fix your issue, and doesnt have any known negatives right now, let's just go for it

colinator27:
Cannot recall the reason that code was there, so for now I'll merge this. If anything else is affected, we'll find out quickly enough.

PR #1511, opened 2023-11-05
Add more documentation params

Miepee:
## Description
Adds more documentation and some other small clean up my IDE screamed at me


colinator27:
Looks good overall! Only real concern would be the change of "FunctionType" to "CurveType", but that may be unimportant.

Miepee:
No script in our codebase uses it.

If another external script uses it, then they will get code breakage, but we currently don't even have a stable API anyway.

PR #1512, opened 2023-11-05
Memory and small runtime optimizations

Miepee:
## Description
Does some small code and memory optimizations

### Caveats
None that i'm aware of.

### Notes
- The custom "GetValueOrDefault" method was removed in favour of the same method in the standard library.
- The assert has been giving a conditional, so that it does not even execute in release builds
- The `Block[]` change has been done to remove memory overhead from enumerating over an IEnumerable
- The foreach -> for change has been done, as the comment suggests, to reduce memory issues. I did a small benchmark, decompiling ~2000 code entries in AM2R, which saved almost a full gigabyte.
  Before this foreach change:
 grafik  
  After the foreach change:
  grafik
 The mean time devation is ~1.1%, which could theoretically be margin of error.


colinator27:
Generally looks good, although unsure about the conditional on Assert here. It's used to give errors in file parsing, and could lead to other problems, potentially... Maybe that just means we need another error function to handle those.

Miepee:
From what I can tell, it should be fine.
Whether you get an exception in release mode because the assert failed, or because something critical that depends on the assert right after failed, shouldnt make any difference.

PR #1515, opened 2023-11-12
Add proper unit testing project

Miepee:
## Description
This PR creates another unit testing project, as well as a few small unit tests. Further PRs that add more unit tests are to follow.

### Notes
The existing "testing" project doesnt actually test anything. it's completely useless, and a future PR will completely delete it.
I also didn't want to completely refactor that project, as that would make the diffs too hard to read. So i chose to create a new project instead.

For the new project, xUnit was chosen as the testing framework, instead of nUnit, because xUnit is from my experience a little nicer to work with.

The current tests give us ~40% coverage in the embedded audio and the string models.

### DIFF
--- a/UndertaleModLib/Models/UndertaleString.cs
+++ b/UndertaleModLib/Models/UndertaleString.cs
@@ -62,6 +62,7 @@ public class UndertaleString : UndertaleResource, ISearchable, IDisposable
     /// <returns>A string that represents the current object.</returns>
     public string ToString(bool isGMS2)
     {
+        // TODO: someone clean this up please. this seems insane for a tostring method.
         if (Content == null)
             return "\"null\""; // NPE Fix.
 
@@ -102,6 +103,8 @@ public class UndertaleString : UndertaleResource, ISearchable, IDisposable
     /// <inheritdoc />
     public bool SearchMatches(string filter)
     {
+        // TODO: should this throw instead?
+        if (filter is null) return false;
         return Content?.ToLower().Contains(filter.ToLower()) ?? false;
     }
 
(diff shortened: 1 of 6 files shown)


UndertaleModLib/Models/UndertaleString.cs
```diff
@@ -62,6 +62,7 @@ public void Dispose()
     /// <returns>A string that represents the current object.</returns>
     public string ToString(bool isGMS2)
     {
+        // TODO: someone clean this up please. this seems insane for a tostring method.
```
colinator27, line 65: I think this occurs with corrupted or partially-loaded VM code, which otherwise could crash any code calling this... i.e., decompiling code and the code editors. It should probably stay here.

UndertaleModLib/Util/FileBinaryReader.cs
```diff
@@ -199,6 +199,7 @@ public string ReadGMString()
                 throw new IOException("Reading out of bounds");
 #endif
             string res;
+            // TODO: shouldn't these accomplish the same thing?
```
colinator27, line 202: I think this is written like this because the stack has limited memory, and thus is limited to only storing 1024 bytes.

Miepee: Do arrays of primitive types not land up on the stack?

Miepee: nvm they dont. will remove that then

UndertaleModLib/Models/UndertaleString.cs
Miepee: Well I mentioned "clean this up". not "outright remove".
There's a few things for example at a first glance, where one should use a stringbuilder instead of just concatonating over and over to an already existing string. Or using Trim methods instead.
To my knowledge, this method is called semi-frequently, so it should really get a second pass + some documentation. 

Although now i'm curious how "corrupted or partially loaded VM code" can even happen...

VladiStep (approved):

PR #1516, opened 2023-11-12
Apply some ide warnings

Miepee:
## Description
Applies quick fixed from the IDE for slightly cleaner code

### DIFF
--- a/UndertaleModLib/Models/UndertaleRoom.cs
+++ b/UndertaleModLib/Models/UndertaleRoom.cs
@@ -237,11 +237,9 @@ public class UndertaleRoom : UndertaleNamedResource, INotifyPropertyChanged, IDi
     {
         get
         {
-            return _layers?.Where(l => l.LayerType is LayerType.Background
-                                       && l.BackgroundData.Sprite is null
-                                       && l.BackgroundData.Color != 0)
-                           .OrderBy(l => l.LayerDepth)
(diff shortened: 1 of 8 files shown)


UndertaleModLib/Models/UndertaleRoom.cs
```diff
@@ -237,11 +237,9 @@ public Layer BGColorLayer
...
         get
         {
-            return _layers?.Where(l => l.LayerType is LayerType.Background
-                                       && l.BackgroundData.Sprite is null
-                                       && l.BackgroundData.Color != 0)
-                           .OrderBy(l => l.LayerDepth)
-                           .FirstOrDefault();
+            return (_layers?.Where(l => l.LayerType is LayerType.Background
+                                        && l.BackgroundData.Sprite is null
+                                        && l.BackgroundData.Color != 0)).MinBy(l => l.LayerDepth);
```
VladiStep, line 242: Nice suggestion, I just forgot about this method.
But what if:
```suggestion
            return (_layers?.Where(l => l.LayerType is LayerType.Background
                                        && l.BackgroundData.Sprite is null
                                        && l.BackgroundData.Color != 0))
                            .MinBy(l => l.LayerDepth);
```

Miepee: sure

VladiStep (approved):

Issue #1399, opened 2023-05-27
Cannot compile code that contains Functions defined through a Struct.


nekonesse:
### Describe the bug

When trying to save any code that calls a function through a struct say like "**inventory.add**", UMT refuses to compile the code and let you save.
![image](https://github.com/krzys-h/UndertaleModTool/assets/50628290/7e0704c8-28be-4922-a15d-21535efdaede)

![image](https://github.com/krzys-h/UndertaleModTool/assets/50628290/f489b9b6-1539-45ef-8446-9ef02be2428e)

### Reproducing steps

1. Open any sort of code that has accesses a function through a struct like "inventory.add". (i have my game linked in setup details, you can open up the "obj_global" draw gui event and try it there.)
2. Try to save the data.
3. The error happens.

### Setup Details

UMT Version: 0.5.1.0 Bleeding Edge (current release as of this post)
OS: Windows 10
Game: My own, MichaelTC
[MichealTC-Inventorytest.zip](https://github.com/krzys-h/UndertaleModTool/files/11582624/MichealTC-Inventorytest.zip)

Jacky720:
Duplicate of #1084, I believe.

rgc-exists:
This is just how UMT is for now.

In the Will You Snail community's case the developers are adding a function in the next update that returns an empty struct, as the next update relies heavily on structs.

We can also edit the "broken" scripts with GMML. (GMML is like UMT but you program what changes to make to the data.win, like enhanced UMLib.) GMML lets you add onto functions without having to compile them, which is good when compiling them causes an error. Unfortunately GMML does not have a UI like UMT and there is no documentation, so it's a really steep learning curve.

So if you want to use interact with structs, you might want to try to convince the devs of your game to add a function that returns an empty struct, and then use variable_struct_get() and variable_struct_set() to change the structs... And maybe using GMML.
I doubt Toby Fox will do this cause he's busy but other games might have a chance.

Ultimately... Please just add struct support to UMT XD

Jacky720:
I now have the power to act on my previous comment and close this as a duplicate (of #1084). If any further discussion is required, it can happen there.

PR #1518, opened 2023-11-12
Compiler struct support

### LINKED ISSUE #1084: GameMaker struct type compiler support

Compiler

image

I checked out that UMT decompiles struct very well, but it can't be compiled back.

Implementation-wise, I think there are two ways (not exclusive to each other) to handle structs. One way is to modify parser: [ParseBlock()](https://github.com/krzys-h/UndertaleModTool/blob/e2b8022c72d44c3af83078d7e168f958cd7007a2/UndertaleModLib/Compiler/Parser.cs#L540) should detect structs with `:` token. The other way is to modify lexer: opening/closing braces should be lexed as `OpenBlock`/`CloseBlock` or `OpenStruct`/`CloseStruct`(to be added) based on the context and then be parsed separately from blocks.

I'm not familiar with GameMaker nor with Compiler stuff so I might be wrong.

CST1229:
Resolves #1084

## Description
Adds support for GMS2.3+ structs in the compiler, cherrypicked from [UndertaleModTool Community Edition](https://github.com/XDOneDude/UndertaleModToolCE/) with some additional minor changes.

https://github.com/krzys-h/UndertaleModTool/assets/68464103/7ac27e6d-29be-47fa-a8d7-e9153e586fea

Also fixes a bug where an empty struct definition would cause all assignments after to also be struct-styled, causing syntax errors when recompiling.
a = {}; b: 123

### Caveats
`___struct___` can no longer be used at the start of a function definition name, as parts of the compiler and decompiler assume that the `___struct___` prefix means a struct.

### Notes
I've only tested this on one GMS2022.9 game (Pizza Tower), but it seems to work there. You may want to squash this PR if you don't like some of my ridiculous commit names. (Also, I'm not super experienced with C# or this project's code style so there probably are some things that need  to be looked at.)


UndertaleModLib/Compiler/Compiler.cs
```diff
@@ -54,6 +56,22 @@ public void OnSuccessfulFinish()
                 {
                     foreach (KeyValuePair<string, string> v in GlobalVars)
                         Data?.Variables?.EnsureDefined(v.Key, UndertaleInstruction.InstanceType.Global, false, Data.Strings, Data);
+                    if (Data != null) {
```
Miepee, line 59: Use `is not null` please. Not just here, but everywhere

UndertaleModLib/Compiler/Compiler.cs
```diff
@@ -54,6 +56,22 @@ public void OnSuccessfulFinish()
...
+                            UndertaleScript scriptObj = Data.Scripts.ByName(scriptName);
+                            if (scriptObj != null) Data.Scripts.Remove(scriptObj);
+                            UndertaleCode codeObj = Data.Code.ByName(scriptName);
+                            if (codeObj != null) {
+                                Data.Code.Remove(codeObj);
+                                OriginalCode.ChildEntries.Remove(codeObj);
+                            }
+                            UndertaleFunction functionObj = Data.Functions.ByName(scriptName);
+                            if (functionObj != null) Data.Functions.Remove(functionObj);
+                            Data.KnownSubFunctions.Remove(name);
```
Miepee, line 71: This thing looks a bit weird. Can a scriptName be both in Data.Scripts and Data.Code ? To my knowledge assets have to have unique names.

UndertaleModLib/Compiler/Parser.cs
```diff
@@ -668,6 +690,9 @@ private static Statement ParseFunction(CompileContext context)
                 {
                     expressionMode = false;
                     Statement s = remainingStageOne.Dequeue();
+                    if (s.Text.StartsWith("___struct___")) {
```
Miepee, line 693: Is it possible for any real GM games out in the wild to have functions that start with `__struct__`? If not, you may wanna slap a todo onto this.

UndertaleModLib/Compiler/Parser.cs
```diff
@@ -1553,6 +1579,132 @@ private static Statement ParseArrayLiteral(CompileContext context)
...
+                    varName = usableStructNames.Dequeue();
+                } else {
+                    // Create a new function
+                    int i = context.Data.Code.Count;
+                    do {
+                        varName = "___struct___" + context.OriginalCode.Name.Content +
+                            "__" + uuidCounter++.ToString();
+                        i++;
+                    } while (context.Data.KnownSubFunctions.ContainsKey(varName));
+                }
```
Miepee, line 1609: can you format this to comfort to c# style?

Miepee (commented):
i dont have any experience with that part of the codebase at all, so just a few cents

UndertaleModLib/Compiler/Compiler.cs
CST1229: Yes I think, each function definition is in both Scripts and Code (Data.Scripts.ByName and Data.Code.ByName both return those entries too):
image

UndertaleModLib/Compiler/Parser.cs
CST1229: GM does seem to let you compile functions that start with `___struct___`, so maybe (though unlikely).

UndertaleModLib/Compiler/Compiler.cs
CST1229: done

UndertaleModLib/Compiler/Compiler.cs
Miepee: huh, TIL.

Miepee:
Also can you add fixes #1399 to the inital PR description please??

CST1229:
> Also can you add fixes #1399 to the inital PR description please?

That issue seems more like it's about `struct.func()` calls, which this PR does not resolve.

UndertaleModLib/Compiler/Compiler.cs
```diff
@@ -54,6 +56,27 @@ public void OnSuccessfulFinish()
...
+                                Data.Scripts.Remove(scriptObj);
+                            UndertaleCode codeObj = Data.Code.ByName(scriptName);
+                            if (codeObj is not null)
+                            {
+                                Data.Code.Remove(codeObj);
+                                OriginalCode.ChildEntries.Remove(codeObj);
+                            }
+                            UndertaleFunction functionObj = Data.Functions.ByName(scriptName);
+                            if (functionObj is not null)
+                                Data.Functions.Remove(functionObj);
```
VladiStep, line 75: Are these instances of `Data.X.Remove()` safe?

CST1229: Should be if you're just normally adding and removing struct definitions, as each struct definition gets its own script, the compiler currently won't let you create `___struct___` functions and I'm fairly sure code, script and function references are all passed by reference instead of ID. But I'm not sure.
(If we don't remove unused struct functions, then adding then removing a struct definition would cause an `Unable to find function` error when launching the game.)

UndertaleModLib/Compiler/Parser.cs
```diff
@@ -439,8 +446,17 @@ public static Statement ParseTokens(CompileContext context, List<Lexer.Token> to
...
                 context.Enums.Clear();
+                context.FunctionsToObliterate.Clear();
+                uuidCounter = 0;
                 hasError = false;
 
+                usableStructNames.Clear();
+                foreach (UndertaleCode child in context.OriginalCode.ChildEntries)
+                {
+                    if (child.Name.Content.StartsWith("gml_Script____struct___"))
+                        usableStructNames.Enqueue(child.Name.Content["gml_Script_".Length..]);
```
colinator27, line 457: Is this entirely exhaustive? Depending on case, structs may not start with this, i.e. if embedded inside an anonymous function (`anon` part comes first in those cases). Although, I'm unsure if anonymous functions are supported by this compiler right now anyway.

CST1229: > Depending on case, structs may not start with this, i.e. if embedded inside an anonymous function (anon part comes first in those cases).

At least from my tests in GMS2022.9, that does not seem to be the case (although it might be different in earlier or later versions).
image


> Although, I'm unsure if anonymous functions are supported by this compiler right now anyway.

Sort of, but they don't work that well.
image
image

Miepee: AFAIK there's an open issue somewhere about anonymous functions, to my understanding their status essentially boils down to "unsupported/not working"

UndertaleModLib/Compiler/AssemblyWriter.cs
```diff
@@ -1332,7 +1375,10 @@ private static void AssembleExpression(CodeWriter cw, Parser.Statement e, Parser
                             });
                             cw.typeStack.Push(DataType.Variable);
                             cw.Emit(Opcode.Dup, DataType.Variable).Extra = 0;
-                            cw.Emit(Opcode.PushI, DataType.Int16).Value = (short)-1; // todo: -6 sometimes?
+                            if (isStructDef)
+                                cw.Emit(Opcode.PushI, DataType.Int16).Value = (short)-16;
+                            else
+                                cw.Emit(Opcode.PushI, DataType.Int16).Value = (short)-1; // todo: -6 sometimes?
```
Miepee, line 1381: can you capitalize TODO please?

Miepee: oh whoops. i only saw it now in the diff

Miepee (approved):

UndertaleModLib/Compiler/Parser.cs
colinator27: OK so I further investigated this, and I found out that the code entries I saw were just anonymous functions declared inside of structs, which aren't themselves structs, so this check is fine!

colinator27 (approved):

colinator27:
I temporarily ported these changes over to my new decompiler fork (almost verbatim), and it seems to work great there!

PR #1520, opened 2023-11-14
Revise for audio, texture, globalinit

Miepee:
## Description
Slightly improves documentation
Removes default values


UndertaleModLib/Models/UndertaleEmbeddedTexture.cs
```diff
@@ -68,13 +71,13 @@ public TexData TextureData
...
     /// </summary>
-    public bool TextureExternal { get; set; } = false;
+    internal bool TextureExternal { get; set; }
 
 
     /// <summary>
     /// Helper variable for whether or not a texture was loaded yet.
     /// </summary>
-    public bool TextureLoaded { get; set; } = false;
+    internal bool TextureLoaded { get; set; }
```
VladiStep, line 80: I think this will break the "Warning: Texture failed to load!" warning in the embedded texture editor.
See here - [UndertaleEmbeddedTextureEditor.xaml, line 47-54](https://github.com/krzys-h/UndertaleModTool/blob/fc29c73a55f9212f2cc9bd948c624cf57ddcbe6b/UndertaleModTool/Editors/UndertaleEmbeddedTextureEditor.xaml#L47-L54)

Miepee: yeah probably, mustve missed that in my search results :/

VladiStep: It's just within XAML, so it's not displayed in the references list

UndertaleModLib/Models/UndertaleGlobalInit.cs
```diff
@@ -29,7 +31,7 @@ public void Serialize(UndertaleWriter writer)
     public void Unserialize(UndertaleReader reader)
     {
         _code = new UndertaleResourceById<UndertaleCode, UndertaleChunkCODE>();
-        _code.Unserialize(reader); // TODO: reader.ReadUndertaleObject if one object starts with another one
```
VladiStep, line 32: Are you sure that it's safe?

Miepee: i have not tested it, but considering that `ReadUndertaleObject(T obj)` ends up also doing `obj.Unserialize()` it seemed fine to me.

colinator27: For now this should probably be reverted unless this actually fixes something, as this may mess up object counting.

UndertaleModLib/Models/UndertaleGlobalInit.cs
```diff
@@ -29,7 +31,7 @@ public void Serialize(UndertaleWriter writer)
...
+       _code.Unserialize(reader);   // Cannot use ReadUndertaleObject, as that messes things up.
```
colinator27, line 34: OK so this should be good to merge now, but I think you accidentally removed a space at the start of the line.

colinator27 (changes requested):

colinator27 (approved):

PR #1538, opened 2023-11-15
Refactor from System.Drawing.Common to Skiasharp.

### LINKED ISSUE #875: Migrate away from System.Drawing.Common
Since .NET 6 [System.Drawing.Common is now Windows only](https://docs.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/6.0/system-drawing-common-windows-only)
MS recommends switching to either ImageSharp or SkiaSharp. Depending on what exactly UTMT uses from there, an in-house solution might also be good enough.

Miepee:
## Description
This PR attempts to refactor from System.Drawing.Common to Skiasharp. Doing this successfully means, that we'll be finally able to bump the .NET version for the project and not make the project Windows only by doing so.
Fixes #875 

### Caveats
This is not finished yet

### Notes
TODO: 
- [x] Opening this, because I need help with the GUI part (CC @VladiStep ), as I'm unable to do meaningful changes there due to me using Linux.
- [ ] Someone should test this with QOI games (I don't have any on me).
- [x] Ensure that basic graphic functionality didn't get lost (i.e. viewing embedded texture pages, texture page entries, rooms etc.) once the GUI part has been refactored 
- [x] Make sure that all the graphic related scripts (importAllTilesets, importAllSprites, newTextureRepacker etc.) still work as they should
- [ ] Check if these behave properly on different display scales

### DIFF
--- a/UndertaleModLib/Models/UndertaleEmbeddedTexture.cs
+++ b/UndertaleModLib/Models/UndertaleEmbeddedTexture.cs
@@ -2,8 +2,6 @@
 using System;
 using System.Buffers.Binary;
 using System.ComponentModel;
-using System.Drawing;
-using System.Drawing.Imaging;
 using System.IO;
 using System.Linq;
 using System.Runtime.CompilerServices;
--- a/UndertaleModLib/Models/UndertaleEmbeddedTexture.cs
+++ b/UndertaleModLib/Models/UndertaleEmbeddedTexture.cs
@@ -7,6 +7,8 @@ using System.Drawing.Imaging;
 using System.IO;
 using System.Linq;
 using System.Runtime.CompilerServices;
+using System.Text;
+using SkiaSharp;
 using UndertaleModLib.Util;
 
 namespace UndertaleModLib.Models;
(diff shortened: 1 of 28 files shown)


GitMuslim:
> Opening this, because I need help with the GUI part (CC @VladiStep ), as I'm unable to do meaningful changes there due to me using Linux.

Can't you just use wine? I usually do my work on utmt with wine and it works just fine except for a *few* offs

Miepee:
That works for the GUI, but I don't want to make the core library require the use of wine.
Wine only runs on certain machines, while if we can get it to run natively, it will run anywhere where the dotnet runtime is able to run. Which is to say, a lot more machines.

GitMuslim:
Oh ok.

Miepee:
Opening this PR up for review. I did a very basic test via WINE and it worked as expected there.
But I would like to have someone on windows still test:
- QOI games
- the graphic exporter/importer scripts

colinator27:
So I tested this a few days ago and found that (for sure) SkiaSharp doesn't losslessly process PNG images, so unfortunately this probably won't be able to merge. (And it'll probably be superseded by #1870 and the PRs that follow it.)

However, this is still really good reference for what CSX scripts we'll need to update after #1870 is merged, which I'm going to start working on.

Miepee:
I will close this PR as its superseded as mentioned above.
For script reference, this PR existing is enough reference.

PR #1543, opened 2023-11-17
Fix NRE when trying to export rooms as PNG

### LINKED ISSUE #1430: ExportAllRoomsToPng.csx fails with a "Value cannot be null" error
### Describe the bug

When running the script 'ExportAllRoomsToPng.csx', the first 2ish rooms render successfully but after that the script throws an exception. This isn't a script issue either, as my custom script (which worked until I updated UMT) also stopped working.

### Reproducing steps

1. Click scripts at the top of the window
2. Run "Resource Unpackers/ExportAllRoomsToPng.csx"
3. Some rooms will render, but eventually the script throws System.ArgumentNullException.

### Setup Details

1. UndertaleModTool commit 32cd3ea (bleeding-edge)
2. Windows 11 22H2 Build 22621.1928
3. DELTARUNE Chapter 1&2 1.00 (itch.io)

Miepee:
## Description
Fixes a Null Reference Exception when trying to export rooms as PNG on GMS2 games.

### Notes
Fixes #1430

### DIFF
--- a/UndertaleModTool/Controls/UndertaleRoomRenderer.xaml.cs
+++ b/UndertaleModTool/Controls/UndertaleRoomRenderer.xaml.cs
@@ -64,7 +64,9 @@ namespace UndertaleModTool
             {
                 if (layer.LayerType == LayerType.Assets)
                 {
-                    var particleSystems = layer.AssetsData.ParticleSystems.Select(x => x.ParticleSystem);
+                    var particleSystems = Enumerable.Empty<UndertaleParticleSystem>();
+                    if (layer.AssetsData.ParticleSystems is not null)
+                        particleSystems = layer.AssetsData.ParticleSystems.Select(x => x.ParticleSystem);
                     ParticleSystemRectConverter.Initialize(particleSystems);
                 }
             }
--- a/UndertaleModTool/Controls/UndertaleRoomRenderer.xaml.cs
+++ b/UndertaleModTool/Controls/UndertaleRoomRenderer.xaml.cs
@@ -64,10 +64,13 @@ namespace UndertaleModTool
             {
                 if (layer.LayerType == LayerType.Assets)
                 {
-                    var particleSystems = Enumerable.Empty<UndertaleParticleSystem>();
                     if (layer.AssetsData.ParticleSystems is not null)
+                    {
+                        var particleSystems = Enumerable.Empty<UndertaleParticleSystem>();
                         particleSystems = layer.AssetsData.ParticleSystems.Select(x => x.ParticleSystem);
-                    ParticleSystemRectConverter.Initialize(particleSystems);
+                        ParticleSystemRectConverter.Initialize(particleSystems);    
+                    }
+                    
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


UndertaleModTool/Controls/UndertaleRoomRenderer.xaml.cs
```diff
@@ -64,7 +64,9 @@ private void RoomRenderer_DataContextChanged(object sender, DependencyPropertyCh
             {
                 if (layer.LayerType == LayerType.Assets)
                 {
-                    var particleSystems = layer.AssetsData.ParticleSystems.Select(x => x.ParticleSystem);
+                    var particleSystems = Enumerable.Empty<UndertaleParticleSystem>();
+                    if (layer.AssetsData.ParticleSystems is not null)
+                        particleSystems = layer.AssetsData.ParticleSystems.Select(x => x.ParticleSystem);
                     ParticleSystemRectConverter.Initialize(particleSystems);
```
VladiStep, line 70: I think it should be like this:
```suggestion
                    if (layer.AssetsData.ParticleSystems is not null)
                    {
                        var particleSystems = layer.AssetsData.ParticleSystems.Select(x => x.ParticleSystem);
                        ParticleSystemRectConverter.Initialize(particleSystems);
                    }
```

VladiStep: So this part is executed only if `layer.AssetsData.ParticleSystems` is not null

VladiStep: No, there is no point to do that - if the room doesn't contain any particle system instances, then then converter wouldn't be used for that room.

VladiStep: But I think we still should change this line just in case:
https://github.com/krzys-h/UndertaleModTool/blob/64b52a65ee4b0e8eecfa59ebf0a440eac52f55e3/UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs#L2694
(make it `if (!partSystemsDict?.TryGetValue(...`)

Miepee: changed these, not sure if it was what you wanted.

UndertaleModTool/Controls/UndertaleRoomRenderer.xaml.cs
```diff
@@ -64,8 +64,13 @@ private void RoomRenderer_DataContextChanged(object sender, DependencyPropertyCh
             {
                 if (layer.LayerType == LayerType.Assets)
                 {
-                    var particleSystems = layer.AssetsData.ParticleSystems.Select(x => x.ParticleSystem);
-                    ParticleSystemRectConverter.Initialize(particleSystems);
+                    if (layer.AssetsData.ParticleSystems is not null)
+                    {
+                        var particleSystems = Enumerable.Empty<UndertaleParticleSystem>();
+                        particleSystems = layer.AssetsData.ParticleSystems.Select(x => x.ParticleSystem);
+                        ParticleSystemRectConverter.Initialize(particleSystems);
```
VladiStep, line 71:
```suggestion
                        var particleSystems = layer.AssetsData.ParticleSystems.Select(x => x.ParticleSystem);
                        ParticleSystemRectConverter.Initialize(particleSystems);
```

UndertaleModTool/Controls/UndertaleRoomRenderer.xaml.cs
VladiStep: Almost, see the latest suggestion

VladiStep (approved):

PR #1553, opened 2023-11-27
better profile mode with no dogcheck

Romualdo666:
## Description
Improved profile mode with different profiles system as if it was a project with automatic code loading and all!

### Caveats
My code's a mess, appreciated if someone tries to fix it.

### Notes
I may or may not have grabbed stuff from other branches and modified them to improve. The only one I remember currently is that one pull request that would let you move on rooms with right click. I changed that and made it be able to move by holding space (like in Gamemaker Studio)


Romualdo666:
@Miepee hopefully last ping, sorry.

UndertaleModTool/MainWindow.xaml
```diff
@@ -105,6 +106,7 @@
                     </MenuItem>
                     <Separator/>
                     <MenuItem Header="Generate o_ffset map" Click="MenuItem_OffsetMap_Click"/>
+                    <MenuItem Header="Export to GameMaker project" Click="MenuItem_GameMaker_Click"/>
```
Miepee, line 109: build error

Miepee (changes requested):

Miepee:
If you still want to merge this in, you should probably rebase this correctly.

PR #1585, opened 2023-12-28
"UndertaleCode" improvements.

VladiStep:
## Description
Added new methods: `UndertaleCode.Insert(instructions, index)`, `UndertaleCode.Insert(instruction, index)` and `UndertaleInstruction.Clone()`.

### Caveats
It would require a little bit more work to add some new properties to `UndertaleInstruction`.

### Notes
Initial implementation by @neprim.


VladiStep:
Oops

VladiStep:
I think it's way too complicated to insert instructions, and I am not sure if the `Clone()` method is necessary.

PR #1588, opened 2023-12-31
Fix GMS2 runner discovery

Dobby233Liu:
## Description
Current 2.3 runtimes seems to only provide x64 runners and under the `x64` subfolder, this makes the picker account for that subfolder.

Testcase:
- Install 2023.4 runtime through GM
- Open (for example) UTY v1.1.0
- Select File -> Run game with other runner
- Say "no" to saving

Before: UMT simply starts the game's EXE

After: You get to choose
image

The %systemdrive%\ProgramData (which looks weird) -> %ProgramData% is simply a by-the-way thing. Windows does have a environment variable for it so why?

### Notes
In accordance to https://discord.com/channels/566861759210586112/586954521880690690/1191016299829133342

### DIFF
--- a/UndertaleModTool/Settings.cs
+++ b/UndertaleModTool/Settings.cs
@@ -17,7 +17,7 @@ namespace UndertaleModTool
 
         public string Version { get; set; } = MainWindow.Version;
         public string GameMakerStudioPath { get; set; } = "%appdata%\\GameMaker-Studio";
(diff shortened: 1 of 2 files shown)


Miepee (approved):

colinator27:
There is probably some caveat to checking the running system's environment for 64-bit (namely, the game could still be 32-bit itself), but it's probably not too big of an issue so we can go forward with this for now. Detecting a 32-bit game may require reading an INI file we don't always have access to, anyway.

Miepee:
If the game is 32bit, then it will still be availble as an option. Theres a loop where it iterates through a bunch of runners

PR #1592, opened 2023-12-31
Fix images in the input text box.

### LINKED ISSUE #1590: The input text box (e.g. the "Enter your search" window) accepts images.
### Describe the bug

It's possible to insert images into the `TextInput` window (e.g. when you use "Search.csx" script), though it has no use.
image

### Reproducing steps

1. Open any script that requires some text input.
2. Copy an image to a clipboard.
3. Insert it to the text box.

### Setup Details

1. UTMT version - https://github.com/krzys-h/UndertaleModTool/commit/979db25e44c99ec41641f6b6b50b5949af6a679c

VladiStep:
## Description
Fixes #1590, also made changes in "TextInput.Designer.cs" that Visual Studio automatically applied (just some code optimizations like `this.property` -> `property`).

### DIFF
--- a/UndertaleModTool/Windows/TextInput.cs
+++ b/UndertaleModTool/Windows/TextInput.cs
@@ -125,7 +125,7 @@ namespace UndertaleModTool.Windows
             // Based on this - https://stackoverflow.com/questions/13703083/prevent-images-in-rich-text-box
             // TODO: What if you use a different key combination when you run the app from Wine on Linux?
             if (e.Control && e.KeyCode == Keys.V
-                || e.Shift & e.KeyCode == Keys.Insert)
+                || e.Shift && e.KeyCode == Keys.Insert)
             {
                 if (Clipboard.ContainsText())
                     richTextBox1.Paste(DataFormats.GetFormat(DataFormats.Text));
--- a/UndertaleModTool/Windows/TextInput.cs
+++ b/UndertaleModTool/Windows/TextInput.cs
@@ -122,10 +122,13 @@ namespace UndertaleModTool.Windows
         private void richTextBox1_KeyDown(object sender, KeyEventArgs e)
         {
             // Prevent image pasting
-            if (e.Control && e.KeyCode == Keys.V)
+            // Based on this - https://stackoverflow.com/questions/13703083/prevent-images-in-rich-text-box
+            if (e.Control && e.KeyCode == Keys.V
+                || e.Shift & e.KeyCode == Keys.Insert)
             {
                 if (Clipboard.ContainsText())
                     richTextBox1.Paste(DataFormats.GetFormat(DataFormats.Text));
+
                 e.Handled = true;
             }
         }
--- a/UndertaleModTool/Windows/TextInput.cs
+++ b/UndertaleModTool/Windows/TextInput.cs
@@ -123,6 +123,7 @@ namespace UndertaleModTool.Windows
         {
             // Prevent image pasting
             // Based on this - https://stackoverflow.com/questions/13703083/prevent-images-in-rich-text-box
+            // TODO: What if you use a different key combination when you run the app from Wine on Linux?
             if (e.Control && e.KeyCode == Keys.V
                 || e.Shift & e.KeyCode == Keys.Insert)
             {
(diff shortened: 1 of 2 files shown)


UndertaleModTool/Windows/TextInput.cs
```diff
@@ -80,7 +80,7 @@ public TextInput(string message, string title, string defaultValue, bool allowMu
             label1.Text = message;
             richTextBox1.Multiline = AllowMultiline;
             richTextBox1.DetectUrls = false;
-            richTextBox1.LanguageOption = RichTextBoxLanguageOptions.UIFonts; //prevents the bug with Japanese characters
+            richTextBox1.LanguageOption = RichTextBoxLanguageOptions.UIFonts; // prevents the bug with Japanese characters
```
Miepee: can you also uppercase the "Prevents" please? :)

UndertaleModTool/Windows/TextInput.cs
```diff
@@ -100,7 +100,7 @@ public TextInput(string message, string title, string defaultValue, bool allowMu
 
         private void TextInput_Load(object sender, EventArgs e)
         {
-            richTextBox1.Clear(); //remove "Input text here"
+            richTextBox1.Clear(); // remove "Input text here"
```
Miepee: ditto uppercase

UndertaleModTool/Windows/TextInput.cs
```diff
@@ -111,14 +111,24 @@ private void TextInput_Load(object sender, EventArgs e)
 
             if (richTextBox1.ReadOnly)
             {
-                richTextBox1.BackColor = TextBoxBGColor; //restore color to default one.
+                richTextBox1.BackColor = TextBoxBGColor; // restore color to default one.
```
Miepee: diito uppercase

UndertaleModTool/Windows/TextInput.cs
```diff
@@ -111,14 +111,24 @@ private void TextInput_Load(object sender, EventArgs e)
...
             }
         }
 
         private void richTextBox1_TextChanged(object sender, EventArgs e)
         {
         }
+        private void richTextBox1_KeyDown(object sender, KeyEventArgs e)
+        {
+            // Prevent image pasting
+            if (e.Control && e.KeyCode == Keys.V)
```
Miepee, line 125: this seems bad. is there no way to check whether the user has input a "paste" combination? because this might not be necessarily CTRL+V if you change the combo in windows, or have a different combo in linux and run via WINE.

UndertaleModTool/Windows/TextInput.cs
Miepee: nvm too early. can close this and the other 3

UndertaleModTool/Windows/TextInput.cs
VladiStep: Hmm, I didn't know that you can change that

VladiStep: Wait, it seems you can't change that in Windows - https://answers.microsoft.com/en-us/windows/forum/all/need-to-change-the-shortcut-keys-for-copypaste/87583aa7-68c4-4b22-9cc0-d298b537a094.
Though, there is an alternative insert combination - `Shift+Insert`.

Also, are you sure that Wine doesn't emulate `Ctrl+V`?

VladiStep: Anyway, I added the second combination, and I don't know a better way to detect the paste action there.

Miepee: > it seems you can't change that in Windows 

You can: 
grafik
https://learn.microsoft.com/en-us/windows/powertoys/keyboard-manager

> Also, are you sure that Wine doesn't emulate Ctrl+V?

WINE passes the paste signal through. And linux DE's let you customize anything you want as paste. from META+V to F1 to CTRL+SHIFT+INSERT+PAGEUP+F12+F1+Z+SPACE

Miepee: cant we somehow intercept the paste event, and delete image contents instead? Am not a fan of hardcoding key combinations.

Miepee: can you please unresolve this? @VladiStep

VladiStep: I just checked, PowerToys sends `Ctrl+V` to the app in the end anyway, so it doesn't count (the fix still works).

Miepee: Meh. Still very much not a fan of harcoding key combinations. Can you at least add a small comment there?

Miepee (approved):

PR #1595, opened 2024-01-02
An additional check for GM 2023.8+.

VladiStep:
## Description
Adds an additional check for GM 2023.8+ - this version introduced a new instruction `pushref`.

### DIFF
--- a/UndertaleModLib/UndertaleChunks.cs
+++ b/UndertaleModLib/UndertaleChunks.cs
@@ -902,7 +902,7 @@ namespace UndertaleModLib
 
         private void CheckForGM2023_8(UndertaleReader reader)
         {
-            if (!reader.undertaleData.IsVersionAtLeast(2, 3) || reader.undertaleData.IsVersionAtLeast(2023, 8))
+            if (!reader.undertaleData.IsVersionAtLeast(2023, 2) || reader.undertaleData.IsVersionAtLeast(2023, 8))
                 return;
 
             // Check for a `pushref` instruction
--- a/UndertaleModLib/UndertaleChunks.cs
+++ b/UndertaleModLib/UndertaleChunks.cs
@@ -900,6 +900,19 @@ namespace UndertaleModLib
     {
         public override string Name => "CODE";
 
+        private void CheckForGM2023_8(UndertaleReader reader)
+        {
+            if (!reader.undertaleData.IsVersionAtLeast(2, 3) || reader.undertaleData.IsVersionAtLeast(2023, 8))
+                return;
+
+            // Check for a `pushref` instruction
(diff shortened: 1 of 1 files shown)


UndertaleModLib/UndertaleChunks.cs
```diff
@@ -900,6 +900,19 @@ public class UndertaleChunkCODE : UndertaleListChunk<UndertaleCode>
     {
         public override string Name => "CODE";
 
+        private void CheckForGM2023_8(UndertaleReader reader)
+        {
+            if (!reader.undertaleData.IsVersionAtLeast(2, 3) || reader.undertaleData.IsVersionAtLeast(2023, 8))
```
Jacky720, line 905: The minimum version check should be 2023.2 rather than 2.3 because of the PSEM chunk.

UndertaleModLib/UndertaleChunks.cs
```diff
@@ -900,6 +900,19 @@ public class UndertaleChunkCODE : UndertaleListChunk<UndertaleCode>
...
+        private void CheckForGM2023_8(UndertaleReader reader)
+        {
+            if (!reader.undertaleData.IsVersionAtLeast(2023, 2) || reader.undertaleData.IsVersionAtLeast(2023, 8))
+                return;
+
+            // Check for a `pushref` instruction
+            if (List.Any(c => c.Instructions.Any(x => x.Kind == UndertaleInstruction.Opcode.Break
+                                                      && x.Value is short val && val == -11)))
+            {
+                reader.undertaleData.SetGMS2Version(2023, 8);
```
Miepee, line 912: i assume its not possible for a chunk to overwrite this later?

Miepee (commented):
i assume nothing is able to overwrite the setGM

UndertaleModLib/UndertaleChunks.cs
VladiStep: WDYM?
After this line, no other GM check should be able to downgrade it if you meant that

Miepee: Ah ok. cool

suphamster:
I think there are still problems with GM 2023.8 data loading.
image
https://workupload.com/file/MGVgqNJkze2 https://pixeldrain.com/u/pL9Xhwyy (it's a demo game from GM examples)

Miepee:
This PR isn't supposed to fix 2023.8 data loading, it's supposed to be more of a guide for UTMT to know if something's supposed to be 2023.8

Miepee:
This PR is redundant. during the deserialization of the code chunk, it deserializes the code entries, where it already sets that version:
https://github.com/UnderminersTeam/UndertaleModTool/blob/e8f20627a84ee5b20472326f05f5a08386dffb51/UndertaleModLib/Models/UndertaleCode.cs#L761

PR #1636, opened 2024-01-12
Add ToString overload for CodeInstructions that accepts a StringBuilder

Miepee:
## Description
This adds a ToString overload for CodeInstruction, that accepts a StringBuilder and an index, to insert a string representation of this object at that specified index.

The disassembler of a code entry was frequently calling the normal ToString method, which led to a new stringBuilder instance getting created over and over and over again. To not do this everytime, and just reuse the existing stringbuilder that the disassembler is already using, this method was created.

Here's a benchmark comparing calling the old dissassembler method and this new disassembler method:
grafik
The allocated memory for this is somewhere around half than usual, with performance being *very slightly* faster. The benchmark was performed by disassembling all of AM2R's code entries.

### Caveats
None known.

### Notes
1. I am very unsure whether "StringBuilderHelper" is a good enough name. Also on whether its documentation is clear enough
2. Am Unsure whether the `ToString(StringBuilder, Code, List<uint>, int)` should really be still named "ToString" or if maybe a "StringAtStringBuilder" would maybe be clearer. Reading code like `inst.ToString(sb, code, blocks);` isn't immediately obvious that `sb` is getting modified unless you already know the documentation of that method.

### DIFF
--- a/UndertaleModLib/Decompiler/Disassembler.cs
+++ b/UndertaleModLib/Decompiler/Disassembler.cs
@@ -1,8 +1,8 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Text;
-using System.Threading.Tasks;
 using UndertaleModLib.Models;
 
 namespace UndertaleModLib.Decompiler
@@ -34,18 +34,24 @@ namespace UndertaleModLib.Decompiler
 
             return sb.ToString();
         }
-
+        
         public static string Disassemble(this UndertaleCode code, IList<UndertaleVariable> vars, UndertaleCodeLocals locals)
         {
-            StringBuilder sb = new StringBuilder();
+            // This StringBuilder is shared with the ToString method of the code instructions.
+            // Experimentation has shown that 200 is a good enough starting value for it. 
+            // 300 seemed too high and 100 too low. This may change in the future.
+            StringBuilder sb = new StringBuilder(200);
             if (locals == null && !code.WeirdLocalFlag)
                 sb.Append("; WARNING: Missing code locals, possibly due to unsupported bytecode version or a brand new code entry.\n");
             else
                 sb.Append(code.GenerateLocalVarDefinitions(vars, locals));
 
-            Dictionary<uint, string> fragments = new Dictionary<uint, string>();
+            Dictionary<uint, string> fragments = new(code.ChildEntries.Count);
             foreach (var dup in code.ChildEntries)
+            {
                 fragments.Add(dup.Offset / 4, (dup.Name?.Content ?? "<null>") + $" (locals={dup.LocalsCount}, argc={dup.ArgumentsCount})");
+            }
+
             List<uint> blocks = FindBlockAddresses(code);
 
             foreach (var inst in code.Instructions)
@@ -66,7 +72,8 @@ namespace UndertaleModLib.Decompiler
                     sb.AppendLine($":[{ind}]");
                 }
 
-                sb.AppendLine(inst.ToString(code, blocks));
+                inst.ToString(sb, code, blocks);
+                sb.Append(Environment.NewLine);
             }
 
             sb.AppendLine();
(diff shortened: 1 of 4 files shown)


UndertaleModLib/Decompiler/Disassembler.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
```
VladiStep, line 3: Is it intentional?

UndertaleModLib/Decompiler/Disassembler.cs
```diff
@@ -34,18 +35,24 @@ public static string GenerateLocalVarDefinitions(this UndertaleCode code, IList<
 
             return sb.ToString();
         }
-
+
```
VladiStep, line 38: And this one

UndertaleModLib/Decompiler/Disassembler.cs
Miepee: nope, removed

UndertaleModLib/Decompiler/Disassembler.cs
Miepee: Also removed, no clue how that got in there.

UndertaleModLib/Decompiler/Disassembler.cs
```diff
@@ -34,18 +34,24 @@ public static string GenerateLocalVarDefinitions(this UndertaleCode code, IList<
 
             return sb.ToString();
         }
-
+        
         public static string Disassemble(this UndertaleCode code, IList<UndertaleVariable> vars, UndertaleCodeLocals locals)
         {
-            StringBuilder sb = new StringBuilder();
+            // This stringbuilder is shared with the code instruction's ToString methods.
```
VladiStep, line 40:
```suggestion
            // This `StringBuilder` is shared with the `UndertaleInstruction.ToString()` methods.
```

UndertaleModLib/Decompiler/Disassembler.cs
```diff
@@ -34,18 +34,24 @@ public static string GenerateLocalVarDefinitions(this UndertaleCode code, IList<
...
+            // Experimentation has shown that 200 is a good enough starting value for it. 
+            // 300 seemed too high and 100 too low. This may change in the future.
+            StringBuilder sb = new StringBuilder(200);
             if (locals == null && !code.WeirdLocalFlag)
                 sb.Append("; WARNING: Missing code locals, possibly due to unsupported bytecode version or a brand new code entry.\n");
             else
                 sb.Append(code.GenerateLocalVarDefinitions(vars, locals));
 
-            Dictionary<uint, string> fragments = new Dictionary<uint, string>();
+            Dictionary<uint, string> fragments = new();
```
VladiStep, line 49: You can improve performance by setting an initial capacity:
```suggestion
            Dictionary<uint, string> fragments = new(code.ChildEntries.Count);
```

VladiStep (changes requested):

VladiStep:
Sorry, I shouldn't have chosen the "Add single comment" option.

UndertaleModLib/Decompiler/Disassembler.cs
Miepee: Ah cool, wasn't aware of that

UndertaleModLib/Decompiler/Disassembler.cs
VladiStep: Then:
```suggestion
            // This StringBuilder is shared with the ToString() methods of the code instruction.
```

Because you should use "'s" only for animated objects

Miepee: Ah whoops, this wasn't supposed to be possessive, it was supposed to be plural. Changing this around does make it more obvious, yeah.
Still gonna veto the brackets, method name + it being literally called "method" is enough.

PR #1647, opened 2024-01-18
Optimize disassembler slightly more

Miepee:
## Description
Optimize disassembler slightly more

### Caveats
Anything that relied on calling `GenerateLocalVarDefinitions` will not be able to use it anymore. However, I don't think anyone *should've* relied upon it either, as its a part of the disassembling process.

### DIFF
--- a/UndertaleModLib/Decompiler/Disassembler.cs
+++ b/UndertaleModLib/Decompiler/Disassembler.cs
@@ -9,30 +9,36 @@ namespace UndertaleModLib.Decompiler
 {
     public static class Disassembler
     {
-        public static string GenerateLocalVarDefinitions(this UndertaleCode code, IList<UndertaleVariable> vars, UndertaleCodeLocals locals)
+        private static void AppendLocalVarDefinitionsToStringBuilder(StringBuilder sb, UndertaleCode code, IList<UndertaleVariable> vars, UndertaleCodeLocals locals)
         {
             if (code.WeirdLocalFlag)
-                return "";
-            if (locals == null)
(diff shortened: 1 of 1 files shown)


UndertaleModLib/Decompiler/Disassembler.cs
```diff
@@ -9,30 +9,36 @@ namespace UndertaleModLib.Decompiler
...
+                throw new Exception("Name of the locals block does not match name of the code block!");
             foreach (var arg in locals.Locals)
             {
-                sb.Append(".localvar " + arg.Index + " " + arg.Name.Content);
-                var refvar = referenced.Where((x) => x.Name == arg.Name && x.VarID == arg.Index).FirstOrDefault();
-                if (refvar != null)
+                sb.Append(".localvar ");
+                sb.Append(arg.Index);
+                sb.Append(' ');
+                sb.Append(arg.Name.Content);
```
Jacky720, line 33: I take it that interpolated strings are parsed the same as `+` concatenation, meaning this separation is the only way that improves performance?

Miepee: Actually... not 100% sure. I know interpolation is faster than concatenation. But I haven't actually explicitly tested whether interpolation would be much slower here than seperate calls.
I assumed that seperate calls were faster than interpolation, because 
a) interpolation does some formatting still. While here, it can choose the most optimal path for the constant/int/char/string.
b) interpolation more or less still string-ifies all the individual elements.

BenjaminUrquhart (approved):

PR #1663, opened 2024-01-26
Move most of .dll files into "Libraries" subdirectory ("NetCoreBeauty")

VladiStep:
## Description
1) Removed the "UndertaleModTool.xml" file from output directory.
2) Moved most of .dll files into "Libraries" subdirectory using the ["NetCoreBeauty"](https://github.com/nulastudio/NetBeauty2/tree/v1) library.
image

If everything will be good, then we even can get rid of the single file version.

### Caveats
I didn't test the workflow actions changes.


Miepee:
Going to close this, as I don't think relying on an external library that ships its own CLR is a good solution.

PR #1675, opened 2024-02-08
Decompiler parenthesis cleanup

### LINKED ISSUE #1032: Malformed variables caused by automated simplification of expressions
### Describe the bug

Parentheses removal caused by the expression simplification procedure performed by UMT during code entry edit/compiling can lead to malformed variables, especially when expressions include **compare** operators. 

###  Original code from AM2R before editing:
image

### Code entry after changing the `1` to a `0` in the first line:
image

### Reproducing steps

1. Edit a code entry containing an expression with a compare operator followed by a math operation.
_example_ `(image_angle + ((image_xscale == -1) * 180))`
2. Edit and compile the code entry
3. Observe the parentheses are removed from around the compare operation
_result_ `(image_angle + (image_xscale == -1 * 180))`
No errors when effected code is run.
4. Edit and compile the code entry again.
5. Notice the compare operator's number is now combined with the adjacent math operation. 
_result_ `(image_angle + image_xscale == -180)`
Attempting to run the effected code will now result in a "Malformed Variable" error.

### Setup Details

**UndertaleModTool** v:85430c5 (master)
**OS:** Windows 10 21H1
**Game:** [Another Metroid 2 Remake](https://github.com/AM2R-Community-Developers/AM2RLauncher/releases)
**Code entry from AM2R used in example:** gml_Object_oSpaceJumpTrail_Draw_0

Jacky720:
## Description
Attempt to severely reduce the number of parentheses produced by the decompiler, especially when handling math and comparison expressions.

### Caveats
Each aspect has been tested, but I could be missing complex order-of-operations edge cases where the removed parentheses are required. I think that in that event, it's still preferable to handle those cases as they arise rather than continue with the current state.

### Notes
I may re-add decompiler rounding (numbers like 0.2 currently decompile as 0.20000000000000001 due to floating-point restrictions) to this branch later.

Fixes #1032


UndertaleModLib/Decompiler/Instructions/Decompiler.AssignmentStatement.cs
```diff
@@ -117,7 +117,16 @@ bool checkEqual(ExpressionVar a, ExpressionVar b)
                     }
                 }
             }
-            return String.Format("{0}{1}{2} {3}", varPrefix, varName, context.DecompilingStruct ? ":" : " =", Value.ToString(context));
+
+            // Since parenthesized is now the default output of ExpressionTwo (mainly math operations)
+            // It is specifically safe in variable assignment to not parenthesize this (e.g. "foo = a + b")
+            string stringOfValue;
+            if (Value is ExpressionTwo expressionTwoVal)
```
Miepee, line 124: So what if you have `foo = (a + b) * 4`? Would that be properly shown like that, or shown as `foo = a + b * 4` and then cause further errors down the line?

UndertaleModLib/Decompiler/Instructions/Decompiler.ExpressionTwo.cs
```diff
@@ -34,15 +34,69 @@ public override Statement CleanStatement(DecompileContext context, BlockHLStatem
...
+                argument is not ExpressionConstant &&
+                arg.Contains(' ', StringComparison.InvariantCulture))
+                needsParens = true;
+            else
+                needsParens = false;
+
+
+            if (argument is ExpressionTwo argumentAsBinaryExpression)
+            {
+                int myPriorityLevel = Opcode switch
```
Miepee, line 51: "myprioritylevel" is a bit of a weird variable name. These sound more like they should be "firstArgumentPriorityLevel" and "secondPriorityLevel"?

UndertaleModLib/Decompiler/Instructions/Decompiler.AssignmentStatement.cs
Jacky720: That would evaluate as an ExpressionTwo where its first argument is an ExpressionTwo of lower precedence. This only removes the outermost parentheses from `foo = ((a + b) * 4)`.

Miepee: ah right.

Miepee (approved):

UndertaleModLib/Decompiler/Instructions/Decompiler.ExpressionTwo.cs
```diff
@@ -34,15 +34,69 @@ public override Statement CleanStatement(DecompileContext context, BlockHLStatem
             return this;
         }
 
-        public override string ToString(DecompileContext context)
+        private string AddParensIfNeeded(Expression argument, DecompileContext context)
+        {
+            string arg = argument.ToString(context);
+            bool needsParens;
+            if (arg.Substring(0, 1) != "(" &&
```
Miepee, line 41: Just noticed, why use substring, instead of `arg[0]`

Jacky720: Oh, I forgot that was an option.

CST1229:
This seems to ignore that subtraction and diviison aren't commutative; `a = b - (c - d)` gets decompiled as `a = b - c - d`, which is different.

UndertaleModLib/Decompiler/Instructions/Decompiler.ExpressionTwo.cs
```diff
@@ -34,15 +34,69 @@ public override Statement CleanStatement(DecompileContext context, BlockHLStatem
...
+
+                // Suppose we have "(arg1a argOp arg1b) opcode argument2", and are wondering whether the depicted parentheses are needed
+                // If the argument's opcode is more highly-prioritized than our own, such as it being multiplication
+                // while we use addition, then no parentheses are required.
+                // If the argument's opcode doesn't fall into typical math rules (that is, I don't know my full order of operations)
+                // Assume it has lower priority and needs parentheses to clarify.
+                // Parentheses are also not needed for operations of the same level, especially string concatenation.
+                needsParens = (outerPriorityLevel > argPriorityLevel);
+                if (outerPriorityLevel == 0)
+                    needsParens = true; // Better safe than sorry
```
CST1229, line 77: This is how I'd fix the above comment (non-commutative operators):
```suggestion
                    needsParens = true; // Better safe than sorry
                
                // Non-commutative operators may still need parentheses, e.g `a - (b - c)`
                bool nonCommutative = Opcode == UndertaleInstruction.Opcode.Sub || Opcode == UndertaleInstruction.Opcode.Div;
                if (isSecond && nonCommutative && Opcode == argumentAsBinaryExpression.Opcode)
                    needsParens = true;
```

CST1229: Wait whoops I forgot that I added an additional argument to the function in my fork which I forgot to put here oops sorry.

Miepee:
compiling fails

Jacky720:
Wow, you both got here quick. Does the new commit fix it?

UndertaleModLib/Decompiler/Instructions/Decompiler.ExpressionTwo.cs
```diff
@@ -34,15 +34,74 @@ public override Statement CleanStatement(DecompileContext context, BlockHLStatem
...
+                
+                // Non-commutative operators may still need parentheses, e.g `a - (b - c)`
+                bool nonCommutative = Opcode == UndertaleInstruction.Opcode.Sub || Opcode == UndertaleInstruction.Opcode.Div;
```
Miepee, line 80: can you put this into parantheses please? `= (Opcode == ...Sub)  || (Opcode == ...Div)`?
Could alternatively also use pattern matching here.

CST1229:
> Wow, you both got here quick. Does the new commit fix it?

Yep.

BenjaminUrquhart (approved):

PR #1677, opened 2024-02-11
Fixed spine sprites in >= 2023.1

XdotCore:
Submachine Legacy did not work with before changing this.

## Description
In gamemaker version past 2023.4, the image data for a spine sprite is held within UndertaleSprite.TextureEntry rather than the UndertaleSpineTextureEntry.

This fixes that by checking the version and populating UndertaleSprite.Textures but excluding UndertaleSpineTextureEntry.TexBlob.

### Caveats
* There are 4 bytes in the UndertaleSpineTextureEntry part of the binary that I do not know the purpose of, but it doesn't seem to affect anything.
* I have not fully tested every version, but it works on at least 2023.4 and 2023.8, and it is possible this change is needed for before 2023.4.

### Notes
I was able to add this smiley face to the spined flame with this change:
spine change example

### DIFF
--- a/UndertaleModLib/Models/UndertaleSprite.cs
+++ b/UndertaleModLib/Models/UndertaleSprite.cs
@@ -42,8 +42,14 @@ public class UndertaleSpineTextureEntry : UndertaleObject, IDisposable
     {
         writer.Write(PageWidth);
         writer.Write(PageHeight);
-        writer.Write(TexBlob.Length);
-        writer.Write(TexBlob);
+        // TODO: check if this is also the case on earlier versions
+        if (writer.undertaleData.IsVersionAtLeast(2023, 4)) {
+            // TODO: figure out what these bytes mean
+            writer.Write(0);
+        } else {
+            writer.Write(TexBlob.Length);
+            writer.Write(TexBlob);
+        }
     }
 
     /// <inheritdoc />
@@ -51,14 +57,24 @@ public class UndertaleSpineTextureEntry : UndertaleObject, IDisposable
     {
         PageWidth = reader.ReadInt32();
         PageHeight = reader.ReadInt32();
-        TexBlob = reader.ReadBytes(reader.ReadInt32());
+        // TODO: check if this is also the case on earlier versions
+        if (reader.undertaleData.IsVersionAtLeast(2023, 4)) {
+            // TODO: figure out what these bytes mean
+            reader.ReadInt32();
+        } else
+            TexBlob = reader.ReadBytes(reader.ReadInt32());
     }
 
     /// <inheritdoc cref="UndertaleObject.UnserializeChildObjectCount(UndertaleReader)"/>
     public static uint UnserializeChildObjectCount(UndertaleReader reader)
     {
         reader.Position += 8;                        // Size
-        reader.Position += (uint)reader.ReadInt32(); // "TexBlob"
+        // TODO: check if this is also the case on earlier versions
+        if (reader.undertaleData.IsVersionAtLeast(2023, 4)) {
+            // TODO: figure out what these bytes mean
+            reader.Position += 4; // unknown
+        } else
+            reader.Position += (uint)reader.ReadInt32(); // "TexBlob"
 
         return 0;
     }
@@ -214,6 +230,7 @@ public class UndertaleSprite : UndertaleNamedResource, PrePaddedObject, INotifyP
     public UndertaleSimpleList<UndertaleSpineTextureEntry> SpineTextures { get; set; }
 
     public bool IsSpineSprite { get => SpineJSON != null && SpineAtlas != null && SpineTextures != null; }
+    public bool SpineHasTextureData { get; set; } = true;
     public bool IsYYSWFSprite { get => YYSWF != null; }
 
     private int _SWFVersion;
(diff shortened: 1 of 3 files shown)


UndertaleModLib/Models/UndertaleSprite.cs
```diff
@@ -42,23 +42,39 @@ public void Serialize(UndertaleWriter writer)
...
     /// <inheritdoc />
     public void Unserialize(UndertaleReader reader)
     {
         PageWidth = reader.ReadInt32();
         PageHeight = reader.ReadInt32();
-        TexBlob = reader.ReadBytes(reader.ReadInt32());
+        // TODO: check if this is also the case on earlier versions
+        if (reader.undertaleData.IsVersionAtLeast(2023, 4)) {
+            // TODO: figure out what these bytes mean
+            reader.ReadInt32();
```
BenjaminUrquhart, line 63: Usually, if there's unknown data, it's still good to store it so it can be written back on save (call it `Unknown` or something). That way, the tool is able to write the data file back byte for byte.

UndertaleModLib/Models/UndertaleSprite.cs
```diff
@@ -42,23 +42,39 @@ public void Serialize(UndertaleWriter writer)
     {
         writer.Write(PageWidth);
         writer.Write(PageHeight);
-        writer.Write(TexBlob.Length);
-        writer.Write(TexBlob);
+        // TODO: check if this is also the case on earlier versions
+        if (writer.undertaleData.IsVersionAtLeast(2023, 4)) {
+            // TODO: figure out what these bytes mean
+            writer.Write(0);
```
BenjaminUrquhart, line 48: See other comment

BenjaminUrquhart (changes requested):

UndertaleModLib/Models/UndertaleSprite.cs
XdotCore: Sure, will do

UndertaleModLib/Models/UndertaleSprite.cs
```diff
@@ -42,23 +48,36 @@ public void Serialize(UndertaleWriter writer)
     {
         writer.Write(PageWidth);
         writer.Write(PageHeight);
-        writer.Write(TexBlob.Length);
-        writer.Write(TexBlob);
+        // TODO: check if this is also the case on earlier versions
+        if (writer.undertaleData.IsVersionAtLeast(2023, 4)) {
+            writer.Write(Unknown);
+        } else {
```
Miepee, line 54: please use the Allman bracket style

UndertaleModLib/Models/UndertaleSprite.cs
```diff
@@ -42,23 +48,36 @@ public void Serialize(UndertaleWriter writer)
...
     /// <inheritdoc />
     public void Unserialize(UndertaleReader reader)
     {
         PageWidth = reader.ReadInt32();
         PageHeight = reader.ReadInt32();
-        TexBlob = reader.ReadBytes(reader.ReadInt32());
+        // TODO: check if this is also the case on earlier versions
+        if (reader.undertaleData.IsVersionAtLeast(2023, 4)) {
+            Unknown = reader.ReadInt32();
+        } else
```
Miepee, line 68: ditto

UndertaleModLib/Models/UndertaleSprite.cs
```diff
@@ -42,23 +48,36 @@ public void Serialize(UndertaleWriter writer)
...
 
     /// <inheritdoc cref="UndertaleObject.UnserializeChildObjectCount(UndertaleReader)"/>
     public static uint UnserializeChildObjectCount(UndertaleReader reader)
     {
         reader.Position += 8;                        // Size
-        reader.Position += (uint)reader.ReadInt32(); // "TexBlob"
+        // TODO: check if this is also the case on earlier versions
+        if (reader.undertaleData.IsVersionAtLeast(2023, 4)) {
+            reader.Position += 4; // unknown
+        } else
```
Miepee, line 79: ditto

UndertaleModTool/Editors/UndertaleSpriteEditor.xaml.cs
```diff
@@ -51,15 +51,13 @@ private void ExportAllSpine(SaveFileDialog dlg, UndertaleSprite sprite)
...
-                            catch (Exception ex)
-                            {
-                                mainWindow.ShowError("Failed to export file: " + ex.Message, "Failed to export file");
+                        if (sprite.SpineHasTextureData) {
+                            foreach (var tex in sprite.SpineTextures.Select((tex, id) => new { id, tex })) {
+                                try {
+                                    File.WriteAllBytes(Path.Combine(path, tex.id + ext), tex.tex.TexBlob);
+                                } catch (Exception ex) {
+                                    mainWindow.ShowError("Failed to export file: " + ex.Message, "Failed to export file");
+                                }
```
Miepee, line 60: ditto

XdotCore:
I fixed the bracket style to Allman

Miepee (approved):

Jacky720 (commented):
I'm not a huge fan of just calling it Unknown. Seems to me like it's equivalent to the existing TexBlob length, suggesting the actual location of the TexBlob has moved, and I'd really like to track that down. Did you say that the data.win reserializes 1:1 with these changes?

XdotCore:
Last I checked with Submachine: Legacy it was 1 to 1

BenjaminUrquhart (changes requested):
The change was made between 2022.11 and 2023.1 (most likely in 2023.1 since I don't think there's a .12 release).  The blob is indeed moved to the texture entry.

The field is the length of the texture blob. I would rename the variable to reflect this and change the version checks (would be also good to find games that use spine sprites made with 2022.11 and 2023.1 to confirm it works correctly).

Miepee:
bump @XdotCore

XdotCore:
Since we have determined the unknown value to be the length of the texture entry, I think it would be best to find a game where there is multiple texture entries in the spine sprite >= 2023.1 to test if each UndertaleSpineTextureEntry corresponds 1:1 with the textures in the sprite and in the same order.

If that is determined to be true, then it would be best to save the length of the corresponding texture entry by somehow passing that into the UndertaleSpineTextureEntry when the sprite is first unserialized.

In response to @BenjaminUrquhart .

BenjaminUrquhart (approved):
Must have missed the changes, sorry about that!

Issue #1456, opened 2023-08-07
Return statements get removed


kenan238:
### Describe the bug

I have a function with an if-guard
![image](https://github.com/krzys-h/UndertaleModTool/assets/73741117/4235819c-56f8-45d9-98bc-0dbd3266380d)
I compile, the guard is gone
![image](https://github.com/krzys-h/UndertaleModTool/assets/73741117/82227f9e-1c12-4d57-b474-df02640f551d)
I know that I can just do `return 0;` but it would be great to see this fixed, especially when I open a game's script that has these types of guards and it removes them.

### Reproducing steps

1. Make a function
2. Write an if-guard
3. Write the rest
4. Compile
5. It's gone

### Setup Details

1. I am using UTMT v:0.5.1.0
2. I am on Windows 10
3. I am editing the game Pizza Tower

Jacky720:
Looks like I'm making #1223 the central hub for duplicates of this, so, closing as duplicate.

PR #1680, opened 2024-02-13
Fixed disappearing returns

XdotCore:
## Description
This fixes issue #1456 caused by pr #1191. This issue occurred because the pr would remove any `exit` instruction that wasn't followed by an operation starting with `push int` (for some reason), which removed simple `return;`s that didn't meet that req because they compile to `exit` (while `return value;` compiles to `ret`).

### Caveats
N/A

### Notes
N/A


XdotCore:
found a better way

Issue #564, opened 2021-09-21
Add support for member method calls to the compiler


krzys-h:
After #561 is merged, calling global scripts should still work, but calling member functions (e.g. `something.method()`) is not supported yet.

For a code sample, see `gml_Object_obj_darkfountain_Draw_0` in DRCh2.

colinator27:
Implemented in https://github.com/UnderminersTeam/UndertaleModTool/pull/2056, which should be merged at some point soon after testing.

PR #1683, opened 2024-02-17
Compiler self call support

### LINKED ISSUE #1133: Using 'self' Throws Error "Expected token kind ProcVariable, got ProcFunction"
### Describe the bug

When using self it throws an error when it should work (I tested in actual gamemaker and it works there)
Image:
image

### Reproducing steps

1. write ```self.<insert any local function here>```
2. attempt to compile
3. viola, error

### Setup Details

UndertaleModTool Version: bleeding-edge c897860
OS: Windows 10
Game: Benbo Quest 2: BenboWars https://thumbtack.itch.io/benbowars

XdotCore:
Resolves #1133 

## Description
1. Adds support for GMS2.3+ self calls in the compiler, cherrypicked from [UndertaleModTool Community Edition](https://github.com/XDOneDude/UndertaleModToolCE/).

2. Fixed an issue where sub functions in an object should be stored in a self variable without the object function name suffix.
Example: the utmt compiler would make `gml_Script_act_gml_Object_menu_chapter_btn_Create_0` be stored in `self.act_gml_Object_menu_chapter_btn_Create_0` when the original game code had `self.act`.

### Caveats
None

### Notes
I've tested this with Submachine Legacy (2023.8), and the one from #1133 (2022.5)

How to Reproduce:
1. Create a function in a GlobalScript or Object code, then call that function from another function
image

2. In gamemaker (I've only tested this for 2023.8), make a function in an object, then call that function from another one in the same object
image

### DIFF
--- a/UndertaleModLib/Compiler/AssemblyWriter.cs
+++ b/UndertaleModLib/Compiler/AssemblyWriter.cs
@@ -1064,6 +1064,17 @@ namespace UndertaleModLib.Compiler
                     case Parser.Statement.StatementKind.Exit:
                         AssembleExit(cw);
                         break;
+                    case Parser.Statement.StatementKind.ExprVariableRef:
+                        for (int i = s.Children.Count - 1; i >= 1; i--)
+                        {
+                            if (s.Children[i].Kind == Parser.Statement.StatementKind.ExprFunctionCall)
+                            {
+                                PushFunctionArgs(cw, s.Children[i]);
+                            }
+                        }
+                        AssembleVariablePush(cw, s);
+                        cw.Emit(Opcode.Popz, cw.typeStack.Pop());
+                        break;
                     default:
                         AssemblyWriterError(cw, "Expected a statement, none found", s.Token);
                         break;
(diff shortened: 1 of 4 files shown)


Jacky720:
See also #564.

colinator27:
As a heads up, we'll probably be looking at merging this functionality in once the Underanalyzer decompiler is merged in. We do have a few things to make decisions about, though:
- Do we consult the builtin function list to determine self calls, or do we make a setting on the decompiler to force `self.` notation?
- Do we account for more complex chains of calls, like `a.b().c().d()`, or do we hold off until we (possibly) make a new compiler on the Underanalyzer side?

Curious for thoughts!

CST1229:
Just a note, UTMTCE's self calls have received further improvements since this PR (https://github.com/XDOneDude/UndertaleModToolCE/commit/19a4f44b8b4a81e6e22448b3f1c469e0b48b73a2 https://github.com/XDOneDude/UndertaleModToolCE/commit/aeb662421929adb0dd815e2f907b3090ff3cd9cf), and they now do support chains of calls and also local variables (although still using `self`):
```
var a = function() {
    return  {
        b: function() {
            return 123;
        }
    };
}
show_message(self.a().b()) // shows "123"
```

colinator27:
This has been obsoleted by the new compiler, now.

PR #1687, opened 2024-02-21
Increase flexibility in last shader length

### LINKED ISSUE #391: Test project with passthrough shader compiled for vita doesn't open in UndertaleModTool
This game throws an exception regarding the length of shaders. If this exception is ignored, it can resave the file, with a one byte difference. It is a very odd thing, and would like some help investigating it. 

[game.zip](https://github.com/krzys-h/UndertaleModTool/files/6160075/game.zip)

Jacky720:
## Description
Per comment on #391, the last piece of raw shader data can sometimes be misaligned with the padding at the end of the chunk. This PR adds flexibility to the raw shader length checks to permit a greater (padded) size than expected when the raw shader data being parsed is the last one in the shader. Closes #391.

### Caveats
None. Any external use of UndertaleRawShaderData.ReadData should function the same, as the isLast parameter was given a default value.

### Notes
N/A

### DIFF
--- a/UndertaleModLib/Models/UndertaleShader.cs
+++ b/UndertaleModLib/Models/UndertaleShader.cs
@@ -336,7 +336,7 @@ public class UndertaleShader : UndertaleNamedResource, IDisposable
             else
                 next = EntryEnd;
             int length = (int)(next - reader.AbsPosition);
-            HLSL11_VertexData.ReadData(reader, length, HLSL11_VertexData.IsNull);
+            HLSL11_VertexData.ReadData(reader, length, HLSL11_PixelData.IsNull);
         }
         if (!HLSL11_PixelData.IsNull)
         {
@@ -349,7 +349,7 @@ public class UndertaleShader : UndertaleNamedResource, IDisposable
             else
                 next = EntryEnd;
             int length = (int)(next - reader.AbsPosition);
-            HLSL11_PixelData.ReadData(reader, length, HLSL11_PixelData.IsNull);
+            HLSL11_PixelData.ReadData(reader, length, PSSL_VertexData.IsNull);
         }
 
         if (!PSSL_VertexData.IsNull)
--- a/UndertaleModLib/Models/UndertaleShader.cs
+++ b/UndertaleModLib/Models/UndertaleShader.cs
@@ -530,16 +530,16 @@ public class UndertaleShader : UndertaleNamedResource, IDisposable
 
         public void ReadData(UndertaleReader reader, int length, bool isLast = false)
         {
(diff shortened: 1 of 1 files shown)


UndertaleModLib/Models/UndertaleShader.cs
```diff
@@ -528,10 +528,20 @@ public void WriteData(UndertaleWriter writer)
...
         {
-            if (_Length != 0 && _Length != length)
-                throw new UndertaleSerializationException("Failed to compute length of shader data");
+            if (_Length != 0) // _Length is the "expected" length here
+            {
+                if (_Length > length)
+                    throw new UndertaleSerializationException("Failed to compute length of shader data: instructed to read less data than expected.");
+                else if (_Length < length)
+                {
+                    if (isLast) // Normal for the last element due to chunk padding, just trust the system
```
Miepee, line 539: Is there no way to detect how much padding we should have? would be nicer.

Jacky720: I haven't done extensive testing with this. I think the maximum padding would be 15 bytes but I don't know everything about padding and kind of figured we try to work with whatever padding the file comes with.

Miepee: Well, if you know how things are aligned on, you can determine that by checking the current reader position.
i.e. if things are always 16-aligned, and your current position is on 30, you'd have 2 padding. If your position is on 20, you'd have 12 padding etc.

Miepee (approved):

UndertaleModLib/Models/UndertaleShader.cs
```diff
@@ -528,10 +528,21 @@ public void WriteData(UndertaleWriter writer)
...
+                if (_Length > length)
+                    throw new UndertaleSerializationException("Failed to compute length of shader data: instructed to read less data than expected.");
+                else if (_Length < length)
+                {
+                    if (isLast && ((reader.AbsPosition + length) % 16 == 0)) // Normal for the last element due to chunk padding, just trust the system
+                        length = (int)_Length;
+                    else
+                        throw new UndertaleSerializationException("Failed to compute length of shader data: instructed to read more data than expected." +
+                            (isLast ? " Shader data was the last in the shader, but given length was incorrectly padded." : ""));
+                }
```
Miepee, line 544: just noticed: can you make this slightly easier to read by using curly brackets please?

colinator27 (approved):
There may be cases where the "isLast" check isn't actually correct, but considering this should only affect cases where an exception was originally thrown, this should be fine.

PR #1694, opened 2024-02-24
Verify changes were made in ReplaceTextInGML

### LINKED ISSUE #1644: Overzealous recompiling in FindAndReplace.csx
### Describe the bug

When attempting to run FindAndReplace.csx, it will recompile even code entries that do not have the text being searched for, but instead contain unsupported features. This usually terminates the script, even though there are very few scripts containing these features and they rarely overlap with the search text.

It (and possibly the ReplaceTextInGML function it uses) should instead decompile the text, but check if the text replace had any matches before recompiling.

### Reproducing steps

1. Open Undertale Yellow.
2. Attempt to use FindAndReplace.csx. (For example, replacing "Clover" to "Kris")
3. Script crashes on a code entry that doesn't actually contain "Clover".

### Setup Details

1. UndertaleModTool v0.5.1.0
2. Windows 10
3. [Undertale Yellow](https://gamejolt.com/games/UndertaleYellow/136925)

Jacky720:
## Description
Some general cleanup of both the CLI and Tool implementations of the ReplaceTextInGML scripting function. More importantly, adds a check to return early if the text replacement caused no changes, not attempting to recompile, reducing the number of scripts which error. Closes #1644, but while replacing Clover in Undertale Yellow still doesn't work, at least the script causing the error is one that actually contains "Clover". (It's the error with `string()` having an overload the tool doesn't support yet.)

### Caveats
Would cause the CLI version to give an immediate error if the UndertaleData has profile mode enabled. This shouldn't be possible because the CLI doesn't actually have a profile mode save directory. Also, as stated, the example replacement in Undertale Yellow still doesn't work; that would require a change to FindAndReplace to continue after an error in a single code replacement.

### Notes
It's possible the original layout of these implementations (defining a decompile context variable outside the profile mode if-block, then ignoring it and effectively duplicating its definition in one of the blocks, among other oddities) are important and the reworks in this PR could break it. It doesn't seem likely, but it's possible.

### DIFF
--- a/UndertaleModCli/Program.UMTLibInherited.cs
+++ b/UndertaleModCli/Program.UMTLibInherited.cs
@@ -657,11 +657,13 @@ public partial class Program : IScriptInterface
         {
             try
             {
-                if (code is null) // It would just be recompiling an empty string and messing with null entries seems bad
+                // It would just be recompiling an empty string and messing with null entries seems bad
+                if (code is null)
                     return;
-                string original = Decompiler.Decompile(code, decompileContext);
-                passBack = GetPassBack(original, keyword, replacement, caseSensitive, isRegex);
-                if (passBack == original) // No need to compile something unchanged
+                string originalCode = Decompiler.Decompile(code, decompileContext);
+                passBack = GetPassBack(originalCode, keyword, replacement, caseSensitive, isRegex);
+                // No need to compile something unchanged
+                if (passBack == originalCode)
                     return;
                 code.ReplaceGML(passBack, Data);
             }
--- a/UndertaleModCli/Program.UMTLibInherited.cs
+++ b/UndertaleModCli/Program.UMTLibInherited.cs
@@ -657,7 +657,12 @@ public partial class Program : IScriptInterface
         {
             try
             {
-                passBack = GetPassBack(Decompiler.Decompile(code, decompileContext), keyword, replacement, caseSensitive, isRegex);
+                if (code is null) // It would just be recompiling an empty string and messing with null entries seems bad
+                    return;
+                string original = Decompiler.Decompile(code, decompileContext);
+                passBack = GetPassBack(original, keyword, replacement, caseSensitive, isRegex);
+                if (passBack == original) // No need to compile something unchanged
+                    return;
                 code.ReplaceGML(passBack, Data);
             }
             catch (Exception exc)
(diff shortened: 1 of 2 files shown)


UndertaleModCli/Program.UMTLibInherited.cs
```diff
@@ -657,35 +657,22 @@ public void ReplaceTextInGML(UndertaleCode code, string keyword, string replacem
         {
             try
             {
-                passBack = GetPassBack(Decompiler.Decompile(code, decompileContext), keyword, replacement, caseSensitive, isRegex);
+                if (code is null) // It would just be recompiling an empty string and messing with null entries seems bad
```
Miepee, line 660: can you put the comment above the if statment please

UndertaleModCli/Program.UMTLibInherited.cs
```diff
@@ -657,35 +657,22 @@ public void ReplaceTextInGML(UndertaleCode code, string keyword, string replacem
...
+                    return;
+                string original = Decompiler.Decompile(code, decompileContext);
```
Miepee, line 662: *orignalCode

UndertaleModCli/Program.UMTLibInherited.cs
```diff
@@ -657,35 +657,22 @@ public void ReplaceTextInGML(UndertaleCode code, string keyword, string replacem
...
+                passBack = GetPassBack(original, keyword, replacement, caseSensitive, isRegex);
+                if (passBack == original) // No need to compile something unchanged
```
Miepee, line 664: please put comments above if statements

UndertaleModTool/ImportCodeSystem.cs
```diff
@@ -120,7 +120,12 @@ public void ReplaceTextInGML(UndertaleCode code, string keyword, string replacem
             {
                 try
                 {
-                    passBack = GetPassBack((code != null ? Decompiler.Decompile(code, DECOMPILE_CONTEXT ) : ""), keyword, replacement, caseSensitive, isRegex);
+                    if (code is null) // It would just be recompiling an empty string and messing with null entries seems bad
+                        return;
+                    string original = Decompiler.Decompile(code, DECOMPILE_CONTEXT);
+                    passBack = GetPassBack(original, keyword, replacement, caseSensitive, isRegex);
+                    if (passBack == original) // No need to compile something unchanged
```
Miepee, line 127: see cli implementation

UndertaleModTool/ImportCodeSystem.cs
```diff
@@ -135,18 +140,23 @@ public void ReplaceTextInGML(UndertaleCode code, string keyword, string replacem
                     string path = Path.Combine(ProfilesFolder, Data.ToolInfo.CurrentMD5, "Temp", codeName + ".gml");
                     if (File.Exists(path))
                     {
-                        passBack = GetPassBack(File.ReadAllText(path), keyword, replacement, caseSensitive, isRegex);
+                        string original = File.ReadAllText(path);
+                        passBack = GetPassBack(original, keyword, replacement, caseSensitive, isRegex);
+                        if (passBack == original) // No need to compile something unchanged
```
Miepee, line 145: ditto

UndertaleModTool/ImportCodeSystem.cs
```diff
@@ -135,18 +140,23 @@ public void ReplaceTextInGML(UndertaleCode code, string keyword, string replacem
...
                         {
-                            if (context is null)
-                                passBack = GetPassBack((code != null ? Decompiler.Decompile(code, new GlobalDecompileContext(Data, false)) : ""), keyword, replacement, caseSensitive, isRegex);
-                            else
-                                passBack = GetPassBack((code != null ? Decompiler.Decompile(code, context) : ""), keyword, replacement, caseSensitive, isRegex);
+                            if (code is null) // It would just be recompiling an empty string and messing with null entries seems bad
+                                return;
+                            string original = Decompiler.Decompile(code, DECOMPILE_CONTEXT);
+                            passBack = GetPassBack(original, keyword, replacement, caseSensitive, isRegex);
+                            if (passBack == original) // No need to compile something unchanged
```
Miepee, line 158: ditto

Miepee (commented):
Code style aside, seems fine for me.

However, we should probably at one point not have that part of the IScriptInterface and instead have it part of the core lib so that we don't duplicate this code over and over

Miepee (approved):

VladiStep (approved):

PR #1695, opened 2024-02-25
Rudimentary "line-eating" patch

### LINKED ISSUE #1645: Decompiler Error using Undertale With Jsons (Bleeding Edge)
### Describe the bug

This happens after opening an Undertale data.win that has been modified by the UndertaleWithJSONs.csx script
When decompiling some code entries UndertaleModTool behaves weirdly adding comments where it shouldn't and making invalid code.
It doesn't happen on vanilla Undertale itself, but running UndertaleWithJSONs.csx on a vanilla data.win, closing and reopening triggers the bug

image
image
image
image

There are two things i noticed about this bug, 1st it's that the comments are the missing correct code, 2nd it's that this happens only when scr_gettext is used within the code.

### Reproducing steps

1. Open an Undertale data.win
2. Run UndertaleWithJSONs.csx if it hasn't been run already, close and reopen the data.win
3. Open a code entry containing scr_gettext in it

### Setup Details

1. UndertaleModTool Version: Bleeding Edge -  January 13 2024
2. Operating System: Windows 11
3. Game Used: Undertale 1.08

Jacky720:
## Description
This fixes a bug where the tool would start a localization comment in Undertale (possibly also Deltarune), but would be unable to fetch the localized string and forget the newline, resulting in the following line of code being absorbed by the comment and disappearing on recompile. Closes #1645.

### Caveats
N/A

### Notes
The current solution is to include the text "\<localization fetch error\>", but I'm amenable to other fixes, such as having no text but moving the newline append out of the loop over matches, or constructing a separate list of successful fetches and only including the comment at all if that has any contents.

### DIFF
--- a/UndertaleModTool/Editors/UndertaleCodeEditor.xaml.cs
+++ b/UndertaleModTool/Editors/UndertaleCodeEditor.xaml.cs
@@ -715,7 +715,7 @@ namespace UndertaleModTool
                                     {
                                         Match match = matches[i];
                                         if (!currDict.TryGetValue(match.Groups[1].Value, out string text))
-                                            continue;
+                                            text = "<localization fetch error>";
 
                                         if (i != matches.Length - 1) // If not the last
                                             decompLinesBuilder.Append($"{text}; ");


Miepee (approved):
Feels fine. Better to let them know that something went wrong than silently surpressing.

VladiStep (approved):

PR #1716, opened 2024-03-23
Use R instead of G17 for double-to-string round trip

CST1229:
## Description
Use the R Double.ToString format of G17 in DoubleToString (which is used for numbers in decompiler output). This prevents things like this from happening:
image

This was happening since #1628. The description of said PR says:
> We use G17 instead of R because MSDN recommends it and it's supposedly faster.

Though that causes decimals to look pretty ugly, and i've had at least 2 people report this as a bug in UTMTCE, a fork I work on (which is on the latest UTMT commit), even though it does not actually affect anything.

### Caveats
I don't think there's any precision issues, as the old version of RoundTrip [used the R format](https://github.com/Miepee/UndertaleModTool/blob/65260c4bf3eaf3fd151d2805fa882b423b7a6110/UndertaleModLib/Decompiler/DoubleToString.cs#L50). Though it _might_ be slightly slower, and it's not recommended.

### DIFF
--- a/UndertaleModLib/Decompiler/DoubleToString.cs
+++ b/UndertaleModLib/Decompiler/DoubleToString.cs
@@ -45,6 +45,7 @@ namespace UndertaleModLib.Decompiler
             // As of time of writing this comment, C# does not offer a way to print fixed point notation while preserving precision.
             // You may ask "But why not use F17?". And the answer is, that for everything but R and G, the precision is hard-capped to 15 (according to MSDN: Double.ToString()).
             // Thus we use R to keep our precision, and then manually convert this to fixed point notation.
+            // R is used instead of G17 to prevent things like 0.1 turning into 0.10000000000000001.
             // For anyone unaware of the general algorithm: you get the exponent 'n', then move the decimal point n times to the right if it's positive / left if it's negative.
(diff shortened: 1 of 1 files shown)


Miepee:
Theoretically fine with the change. But if possible, i'd prefer to use G16 (or G15) instead.

CST1229:
> Theoretically fine with the change. But if possible, i'd prefer to use G16 (or G15) instead.

Wouldn't that cause precision issues, so that numbers at the very edge of the decimal limit would get rounded and cause recompilation to be different?

Miepee:
AKA the exact same behaviour as R.

CST1229:
> AKA the exact same behaviour as R.

I thought it didn't?
https://learn.microsoft.com/en-us/dotnet/api/system.double.tostring?view=net-8.0
image
https://learn.microsoft.com/en-us/dotnet/standard/base-types/standard-numeric-format-strings
image

Miepee:
R doesn't guarantee that it round trips (the reason why MS recommends G17 over R is precisely because of that) and IIRC it already does some rounding.

Jacky720 (approved):

Miepee:
Closing in favour of #1675

PR #1717, opened 2024-03-24
Add support for the GMS2.3.7 bool type

CST1229:
## Description
Adds both compiler and decompiler support for the "true" Boolean type introduced in GMS 2.3.7<sup>[[src](https://gms.yoyogames.com/release-notes-runtime.html#2.3.7.476)]</sup>, backported from UTMT Community Edition (which I did very recently).
In GMS2.3.7+ games, Booleans are no longer part of the asset type resolver and are now a native type (somewhat like the asset refs introduced much later in 2023.8).
The way they work is kinda interesting; Booleans cannot be immediately detected, as they start off as regular `0`/`1` ints. Their Boolean-ness can only be detected once they are popped from the stack for use in an instruction, in which case the Boolean datatype will be used in the assembly, e.g `pop.v.b self.bool_variable`.

Also adds some logic to detect 2.3.7 based on `chknullish` instructions (as nullish comparison also got added in this version), as otherwise this would only work on 2022.1+ games.

### Caveats
Nothing major, as far as I know. There might be some additional place where integers should be able to be casted into booleans in the decompiler that I forgot, but I couldn't find any. Also the 2.3.7 check won't work if the game has no nullish comparisons, but that is probably okay since there are 2022.1+ checks anyways, and if it gets a false negative it just doesn't use the bool type like it did before.

### Notes
This alone fixes probably like 50% of decompiler/compiler inaccuracies in modern GM games.


Miepee (approved):
fine by me, assuming this was tested with previous versions

Jacky720 (approved):
LGTM

colinator27 (approved):

PR #1718, opened 2024-03-30
Add Newtonsoft.Json to the list of assemblies available in CLI scripts

Marioalexsan:
## Description
Scripts that use `Newtonsoft.Json` for parsing JSON files fail to execute when using UndertaleModCli. Unlike the GUI tool, the CLI does not add the assembly to the references available to scripts.

This PR fixes that by adding said reference.

### Caveats
Adds `Newtonsoft.Json` as a dependency to the CLI.

### DIFF
--- a/UndertaleModCli/Program.cs
+++ b/UndertaleModCli/Program.cs
@@ -17,6 +17,7 @@ using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using UndertaleModLib.Models;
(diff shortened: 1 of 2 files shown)


Miepee (approved):

Miepee:
Ideally would be nice to use system.json in the future, but that would require first the GUI to not rely on it.

BenjaminUrquhart (approved):

PR #1722, opened 2024-04-03
Add support for 2024.2 tile layer compression

CST1229:
## Description
GM 2024.2 added compression to tile data in rooms, and this PR adds support for it by just decompressing it upon load and recompressing it on save (to preserve compatibility as much as possible). Also allows setting a 2024 GM version in SetGMS2Version.

The compression is a basic RLE similar to the one the IDE has used for a while, except with bytes:
Basically, the tile data is a sequence of what I call "runs", that each consist of 1 byte (which acts as a length/"opcode", I assume this is a signed byte but I'm gonna just treat it as unsigned) followed by some number of uint32 tile IDs.
If the opcode is between 0-127, then the next `opcode` tile IDs are just put directly into the tile data; a "verbatim run". There can be a 0-length verbatim run, though this normally never happens.
If it's between 128-255, then the next tile ID is repeated `opcode - 127` times; a "repeat run" (yes, not minus 128, so there cannot be a 0-length repeat run).
If the length of a run needs to go above 127/128, they are just split (so e.g for 177 tiles in a row, there'd be a repeat run of length 128, then another repeat run of length 49).
GM encodes singular tiles as verbatim runs.

### Caveats
The 2024.2 detection might throw a false negative if the size of the compressed tile data *somehow* happens to be `TilesX * TilesY * 4` bytes long. I'm not sure if that can even happen due to compressed tile data not being 4-byte-aligned due to the opcodes.

### Notes
Haven't tested this much.


UndertaleModLib/Models/UndertaleRoom.cs
```diff
@@ -1674,11 +1682,178 @@ public static uint UnserializeChildObjectCount(UndertaleReader reader)
...
+                    uint tileCount = tilesX * tilesY;
+                    int tiles = 0;
+                    while (tiles < tileCount)
+                    {
+                        byte opcode = reader.ReadByte();
+                        if (opcode >= 128)
+                        {
+                            // Repeat run
+                            int length = opcode - 127;
+                            reader.ReadUInt32();
```
VladiStep, line 1696: You should:
```suggestion
                            reader.Position += 4;
```

UndertaleModLib/Models/UndertaleRoom.cs
```diff
@@ -1674,11 +1682,178 @@ public static uint UnserializeChildObjectCount(UndertaleReader reader)
...
+                            tiles += length;
+                        }
+                        else
+                        {
+                            // Verbatim run
+                            int length = opcode;
+                            for (int i = 0; i < length; i++)
+                                reader.ReadUInt32();
```
VladiStep, line 1704: See the previous comment.

UndertaleModLib/Models/UndertaleRoom.cs
```diff
@@ -1674,11 +1682,178 @@ public static uint UnserializeChildObjectCount(UndertaleReader reader)
...
+                            int length = opcode - 127;
+                            reader.Position += 4;
+                            tiles += length;
+                        }
+                        else
+                        {
+                            // Verbatim run
+                            int length = opcode;
+                            for (int i = 0; i < length; i++)
+                                reader.Position += 4;
```
BenjaminUrquhart, line 1704: I know this is the result of a previous comment, but this block should be 
```cs
int length = opcode;
reader.Position += length * 4;
```

Jacky720 (approved):

BenjaminUrquhart (changes requested):
Now that I've had a moment to test this, it seems to fail if all rooms have their first layer as not a tile layer (see: [TetherGeist](https://store.steampowered.com/app/2474430/TetherGeist/)). Gonna poke around and see if I can get something working.

BenjaminUrquhart:
I've fixed the detection locally, but now it fails to load the tileset (10th room of TetherGeist, layer 28)
```
 ---> System.IndexOutOfRangeException: Index was outside the bounds of the array.
   at UndertaleModLib.Models.UndertaleRoom.Layer.LayerTilesData.ReadCompressedTileData(UndertaleReader reader) in C:\Users\benja\source\repos\UndertaleModTool\UndertaleModLib\Models\UndertaleRoom.cs:line 1749
   at UndertaleModLib.Models.UndertaleRoom.Layer.LayerTilesData.Unserialize(UndertaleReader reader) in C:\Users\benja\source\repos\UndertaleModTool\UndertaleModLib\Models\UndertaleRoom.cs:line 1664
   at UndertaleModLib.UndertaleReader.ReadUndertaleObject[T](T obj) in C:\Users\benja\source\repos\UndertaleModTool\UndertaleModLib\UndertaleIO.cs:line 580
   ```
   
   This corresponds to this line
   ```cs
                    byte length = reader.ReadByte();
                    if (length >= 128)
                    {
                        // Repeat run
                        int runLength = length - 128 + 1;
                        uint tile = reader.ReadUInt32();
                        for (int i = 0; i < runLength; i++)
                        {
                            TileData[y][x] = tile; // <----  right here
                            if (NextTile())
                                break;
                        }
                    }

                    ```

**Comment by BenjaminUrquhart** [comment]
It appears if the last 2 tiles in the layer are different, 2 blank tiles are inserted to work around a GMAC bug, and that's why there are misalignments. There is still an index out of bounds in TetherGeist even when accounting for these.

**Comment by BenjaminUrquhart** [comment]
I pushed my changes as a PR to this branch https://github.com/CST1229/UndertaleModTool/pull/1

File: UndertaleModTool/MainWindow.xaml.cs [hunk 4]
```diff
@@ -1,7 +1,8 @@
-﻿using Microsoft.CodeAnalysis;
+﻿#define WARNING_POPUP
```
Miepee, line 1: ?

BenjaminUrquhart: I accidentally committed some of the code I wrote to help debug

BenjaminUrquhart (approved):
I went ahead and made the changes myself, turns out my test game has other loading problems, but rooms themselves seem to be fine.

PR #1748, opened 2024-05-01
Fix 2024.2 detection (makes Undertale Yellow load again)

BenjaminUrquhart:
## Description

Rooms with effects on their tile layers would trigger the 2024.2 check. This PR fixes that.


Miepee (approved):

colinator27 (approved):
As per discussions on Discord, this check isn't necessarily exhaustive, but due to how unlikely it is for any real game to hit that check, it should be fine for now.

PR #1760, opened 2024-05-10
Fail-safe in DecompileContext.BooleanTypeEnabled

Jacky720:
## Description
This PR adds a null check to the recently-added `DecompileContext.BooleanTypeEnabled`. Fixes a bug where the first code entry being decompiled would hang indefinitely, and a bug where Undertale textdata localization would always fail.

### Caveats
May cause false negatives on that version match; to be frank, I'm not sure how `GlobalContext.Data` ends up being null myself.

### Notes
We should get this in ASAP.

### DIFF
--- a/UndertaleModLib/Decompiler/DecompileContext.cs
+++ b/UndertaleModLib/Decompiler/DecompileContext.cs
@@ -14,7 +14,7 @@ public class DecompileContext
     public UndertaleCode TargetCode;
     public UndertaleGameObject Object;
     public static bool GMS2_3;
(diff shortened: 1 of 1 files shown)


colinator27 (approved):
Apparently the data can be null due to (at least) this: https://github.com/UnderminersTeam/UndertaleModTool/blob/master/UndertaleModTool/Editors/UndertaleCodeEditor.xaml.cs#L521. Not sure why that's the case, but this fix should be fine for now.

Miepee:
I'd personally like to fix the underlying root cause than use bandaids

Miepee (approved):
good enough for now

PR #1770, opened 2024-05-25
Fix short-circuit or turning into ternaries in 2.3.7+ games

CST1229:
Fixes a regression caused by #1717

## Description
This PR should fix || turning into ternaries like `left ? true : right` in 2.3.7 games. This as just caused by the detection for them only checking for the *numbers* 1/0, but now that they can be casted into booleans they must also check for true/false too.

### Caveats
idk

### DIFF
--- a/UndertaleModLib/Decompiler/Decompiler.cs
+++ b/UndertaleModLib/Decompiler/Decompiler.cs
@@ -1329,10 +1329,17 @@ namespace UndertaleModLib.Decompiler
             return statement;
         }
 
-        private static bool TestNumber(Statement statement, int number, DecompileContext context = null)
+        private static bool TestNumber(Statement statement, int number)
         {
             statement = UnCast(statement);
             return (statement is ExpressionConstant constant) && constant.EqualsNumber(number);
+        }
+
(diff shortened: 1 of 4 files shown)


Underanalyzer
Miepee, line 1: you should remove this submodule.

UndertaleModLib/Decompiler/Decompiler.cs
```diff
@@ -1335,6 +1335,13 @@ private static bool TestNumber(Statement statement, int number, DecompileContext
             return (statement is ExpressionConstant constant) && constant.EqualsNumber(number);
         }
 
+        // A bool-like can be either 1, 0, true or false.
+        private static bool TestBoolLike(Statement statement, bool boolean, DecompileContext context = null)
```
Miepee, line 1339: `context` is never used.

Miepee (changes requested):

Underanalyzer
CST1229: whoops (i was testing collinator's underanalyzer branch before and i guess git did a git thing)

Miepee (approved):

Jacky720 (approved):

PR #1781, opened 2024-06-01
Fix font size being saved as a uint on gms2.3+

### LINKED ISSUE #1654: decimal in float EmSize discarded in conversion to uint
### Describe the bug

Loading a game with a font with decimal EM sizes can cause a different data.win output since the float is converted to a uint.

### Reproducing steps

1. Open this game [fontissue.zip](https://github.com/krzys-h/UndertaleModTool/files/13998694/fontissue.zip) (empty game with a single font that should have 12.5 size)
2. Font size is just 12
3. After saving the result is different to the original file

### Setup Details

1. Commit 9ba5d85
2. Windows 11
3. Chicory: A Colorful Tale (originally), attached data.win

Miepee:
## Description
Fixes #1654 

### Notes
I did not test this.

### DIFF
--- a/UndertaleModLib/Models/UndertaleFont.cs
+++ b/UndertaleModLib/Models/UndertaleFont.cs
@@ -24,9 +24,9 @@ public class UndertaleFont : UndertaleNamedResource, IDisposable
     public bool EmSizeIsFloat { get; set; }
 
     /// <summary>
-    /// The font size in Ems. In Game Maker: Studio 2.3 and above, this is a float instead.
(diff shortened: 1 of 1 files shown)


colinator27:
Hmm... are we safe from precision errors on games that use the `uint` version of the field?

Miepee:
We should be. At least if i read MSDN correctly, it states that every integral type is fine to and from a float.

colinator27:
With 32-bit floating point at least, I think there's a limit around 16 million or so before we start losing precision, but I don't think that field will ever get that high, so it's probably fine.

colinator27 (approved):

Jacky720 (approved):

PR #1807, opened 2024-06-10
Memory Optimization to not resize a dictionary

Miepee:
## Description
See the comment in the diff. Saved for me ~40MB in the Large Object Heap.

### Caveats
Slightly more unreadable code.

### Notes
N/A

### DIFF
--- a/UndertaleModLib/Compiler/Compiler.cs
+++ b/UndertaleModLib/Compiler/Compiler.cs
@@ -106,7 +106,25 @@ namespace UndertaleModLib.Compiler
 
         private void MakeAssetDictionary()
         {
+            // Clear the dictionary first and set the worst case max size so that we don't resize it over and over
             assetIds.Clear();
+            int maxSize = 0;
+            maxSize += Data?.GameObjects?.Count ?? 0;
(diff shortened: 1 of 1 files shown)


VladiStep:
I think you should add a null check in the beginning so it only clears `scripts` and `assetIds`  if `Data is null`.

UndertaleModLib/Compiler/Compiler.cs
```diff
@@ -106,7 +106,25 @@ public void Setup(bool redoAssets = false)
 
         private void MakeAssetDictionary()
         {
+            // Clear the dictionary first and set the worst case max size so that we don't resize it over and over
             assetIds.Clear();
+            int maxSize = 0;
```
VladiStep, line 111: I think there should be a `Data` null check, so if it's null, then this method should only clear `assetIds` and `scripts`.
This will allow us to change `Data?.` to `Data.`.

VladiStep: Nevermind, I am blind, I've already wrote that above...

colinator27 (approved):

Jacky720 (approved):

PR #1813, opened 2024-06-24
Add parentheses to nested comparison expressions

Jacky720:
## Description
Minor fix for a bug reported on Discord where the code `x -= ((1 + statetime) < (2 + statetime)) < 4` would decompile as `x -= (1 + statetime) < (2 + statetime) < 4`.

### Caveats
Only the possible exceptions where parentheses might be helpful to other, yet rarer, expressions, which would have already been decompiling incorrectly.

### Notes
N/A

### DIFF
--- a/UndertaleModLib/Decompiler/Instructions/Decompiler.ExpressionCompare.cs
+++ b/UndertaleModLib/Decompiler/Instructions/Decompiler.ExpressionCompare.cs
@@ -28,7 +28,12 @@ public static partial class Decompiler
         public override string ToString(DecompileContext context)
         {
             string arg1 = Argument1.ToString(context);
+            if (Argument1 is ExpressionCompare compare1)
+                arg1 = compare1.ToStringWithParen(context);
             string arg2 = Argument2.ToString(context);
+            if (Argument2 is ExpressionCompare compare2)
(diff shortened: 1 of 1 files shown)


Miepee (approved):

CST1229:
`a < b < c` _is_ valid GML syntax (it's accepted by GameMaker's compiler), so I think it might be better (saves parentheses) to instead allow multiple comparison operators next to eachother in the compiler (like it already is possible with math), which I already did for UTMTCE. Maybe I should PR that.

Jacky720:
What does GameMaker's compiler produce in this case? If it's like Python, it would be `a < b && b < c`, but if it's actually producing `(a < b) < c` as in turning the first comparison from boolean to number, we should decompile that.

Jacky720:
Yeah, I just spun up a dummy project, that's exactly what it's doing.
image
"a < b" is false, 0, and 0 < c. `(a < b) < c` is the opposite of the code's intent, of course, but consistent with the new decompilation (and the mod tool compiler can handle it).

colinator27 (approved):

Miepee:
Even though it is a compiler bug, let's just always force parattheses for now on. Makes that code way more readable and prevents user bugs.

PR #1828, opened 2024-07-10
Restore 2.3.1 version check (part 2)

colinator27:
## Description
See https://github.com/UnderminersTeam/UndertaleModTool/pull/1696; this just moves the check to the object count unserialization stage, preventing the warning that can occur otherwise. Closes https://github.com/UnderminersTeam/UndertaleModTool/issues/1414.

## Notes
I think it's OK that this can potentially yield false positives/negatives, as the original PR says, because newer GM versions bypass this check altogether.

### DIFF
--- a/UndertaleModLib/Models/UndertaleAnimationCurve.cs
+++ b/UndertaleModLib/Models/UndertaleAnimationCurve.cs
@@ -133,6 +133,29 @@ public class UndertaleAnimationCurve : UndertaleNamedResource, IDisposable
             Name = reader.ReadUndertaleString();
             Function = (FunctionType)reader.ReadUInt32();
             Iterations = reader.ReadUInt32();
+            // This check is partly duplicated from UndertaleChunks.cs, but it's necessary to handle embedded curves
+            // (for example, those in SEQN in the TS!Underswap v1.0 demo; see issue #1414)
+            if (!reader.undertaleData.IsVersionAtLeast(2, 3, 1))
+            {
+                long returnPosition = reader.AbsPosition;
+                uint numPoints = reader.ReadUInt32();
+                if (numPoints > 0)
+                {
+                    reader.AbsPosition += 8;
+                    if (reader.ReadUInt32() != 0) // in 2.3 a int with the value of 0 would be set here,
+                    {                             // it cannot be version 2.3 if this value isn't 0
+                        reader.undertaleData.SetGMS2Version(2, 3, 1);
+                    }
+                    else
+                    {
+                        if (reader.ReadUInt32() == 0)                      // At all points (besides the first one)
+                            reader.undertaleData.SetGMS2Version(2, 3, 1);  // if BezierX0 equals to 0 (the above check)
+                                                                           // then BezierY0 equals to 0 as well (the current check)
+                    }
+                }
+
+                reader.AbsPosition = returnPosition;
+            }
             Points = reader.ReadUndertaleObject<UndertaleSimpleList<Point>>();
         }
 
--- a/UndertaleModLib/Models/UndertaleAnimationCurve.cs
+++ b/UndertaleModLib/Models/UndertaleAnimationCurve.cs
@@ -133,6 +133,7 @@ public class UndertaleAnimationCurve : UndertaleNamedResource, IDisposable
             Name = reader.ReadUndertaleString();
             Function = (FunctionType)reader.ReadUInt32();
             Iterations = reader.ReadUInt32();
+
             // This check is partly duplicated from UndertaleChunks.cs, but it's necessary to handle embedded curves
             // (for example, those in SEQN in the TS!Underswap v1.0 demo; see issue #1414)
             if (!reader.undertaleData.IsVersionAtLeast(2, 3, 1))
@@ -156,6 +157,7 @@ public class UndertaleAnimationCurve : UndertaleNamedResource, IDisposable
 
                 reader.AbsPosition = returnPosition;
             }
+
             Points = reader.ReadUndertaleObject<UndertaleSimpleList<Point>>();
         }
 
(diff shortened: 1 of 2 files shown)


UndertaleModLib/UndertaleChunks.cs
```diff
@@ -1522,10 +1524,10 @@ private void CheckForGMS2_3_1(UndertaleReader reader)
                 return;
             }
 
-            reader.AbsPosition = reader.ReadUInt32(); // go to the first "Point"
+            reader.AbsPosition = reader.ReadUInt32(); // Go to the first "Point"
             reader.Position += 8;
 
-            if (reader.ReadUInt32() != 0) // in 2.3 a int with the value of 0 would be set here,
+            if (reader.ReadUInt32() != 0) // In 2.3 a int with the value of 0 would be set here,
```
Miepee, line 1530:
```suggestion
            if (reader.ReadUInt32() != 0) // In 2.3 an int with the value of 0 would be set here,
```

UndertaleModLib/UndertaleChunks.cs
```diff
@@ -1522,10 +1524,10 @@ private void CheckForGMS2_3_1(UndertaleReader reader)
...
             {                             // it cannot be version 2.3 if this value isn't 0
```
Miepee, line 1531:
```suggestion
            {                             // It cannot be version 2.3 if this value isn't 0
```

UndertaleModLib/Models/UndertaleAnimationCurve.cs
```diff
@@ -141,6 +142,30 @@ public static uint UnserializeChildObjectCount(UndertaleReader reader)
         {
             reader.Position += 12;
 
+            // This check is partly duplicated from UndertaleChunks.cs, but it's necessary to handle embedded curves
+            // (For example, those in SEQN in the TS!Underswap v1.0 demo; see issue #1414)
+            if (!reader.undertaleData.IsVersionAtLeast(2, 3, 1))
+            {
+                long returnPosition = reader.AbsPosition;
+                uint numPoints = reader.ReadUInt32();
```
Miepee, line 150: what is a point here in this context?

UndertaleModLib/Models/UndertaleAnimationCurve.cs
```diff
@@ -141,6 +142,30 @@ public static uint UnserializeChildObjectCount(UndertaleReader reader)
...
+                if (numPoints > 0)
+                {
+                    reader.AbsPosition += 8;
+                    if (reader.ReadUInt32() != 0) // In 2.3 a int with the value of 0 would be set here,
+                    {                             // it cannot be version 2.3 if this value isn't 0
```
Miepee, line 155:
```suggestion
                    {                             // It cannot be version 2.3 if this value isn't 0
```

UndertaleModLib/Models/UndertaleAnimationCurve.cs
```diff
@@ -141,6 +142,30 @@ public static uint UnserializeChildObjectCount(UndertaleReader reader)
...
+            // This check is partly duplicated from UndertaleChunks.cs, but it's necessary to handle embedded curves
+            // (For example, those in SEQN in the TS!Underswap v1.0 demo; see issue #1414)
+            if (!reader.undertaleData.IsVersionAtLeast(2, 3, 1))
+            {
+                long returnPosition = reader.AbsPosition;
+                uint numPoints = reader.ReadUInt32();
+                if (numPoints > 0)
+                {
+                    reader.AbsPosition += 8;
+                    if (reader.ReadUInt32() != 0) // In 2.3 a int with the value of 0 would be set here,
```
Miepee, line 154:
```suggestion
                    if (reader.ReadUInt32() != 0) // In 2.3 an int with the value of 0 would be set here,
```

UndertaleModLib/Models/UndertaleAnimationCurve.cs
```diff
@@ -141,6 +142,30 @@ public static uint UnserializeChildObjectCount(UndertaleReader reader)
...
+                    {                             // it cannot be version 2.3 if this value isn't 0
+                        reader.undertaleData.SetGMS2Version(2, 3, 1);
+                    }
+                    else
+                    {
+                        if (reader.ReadUInt32() == 0)                      // At all points (besides the first one)
+                            reader.undertaleData.SetGMS2Version(2, 3, 1);  // if BezierX0 equals to 0 (the above check)
```
Miepee, line 161:
```suggestion
                            reader.undertaleData.SetGMS2Version(2, 3, 1);  // If BezierX0 equals to 0 (the above check)
```

UndertaleModLib/Models/UndertaleAnimationCurve.cs
```diff
@@ -141,6 +142,30 @@ public static uint UnserializeChildObjectCount(UndertaleReader reader)
...
+                                                                           // then BezierY0 equals to 0 as well (the current check)
```
Miepee, line 162:
```suggestion
                                                                           // Then BezierY0 equals to 0 as well (the current check)
```

UndertaleModLib/Models/UndertaleAnimationCurve.cs
Miepee, line 161:
```suggestion
                            reader.undertaleData.SetGMS2Version(2, 3, 1);  // If BezierX0 equals to 0 (the above check)
```

Miepee:
I thought there were more issues floating around for this one?

UndertaleModLib/Models/UndertaleAnimationCurve.cs
colinator27: It's the list that's called "Points", AKA points on an animation curve asset

Miepee: I feel like them being named something like `animPoints` would be clearer. Or `animPointCount`.

Jacky720 (approved):

UndertaleModLib/Models/UndertaleAnimationCurve.cs
Miepee: IMO itd be nice if this would be addressed

colinator27: Just refactored this a little bit - should be more clear now (and also technically with one less line of duplicated code)

Miepee (approved):

PR #1838, opened 2024-07-20
Fix erroneous 2022.1 detection

colinator27:
## Description
This fixes erroneous detection in 2.3 games prior to 2.3.2, where a field was removed from room asset layers. Fixes https://github.com/UnderminersTeam/UndertaleModTool/issues/1837.

### Caveats
The only possible caveat is that, in the worst case, some 2022.1 games specifically may no longer get detected. However, I very much doubt that will happen, and I tested multiple games to make sure they still load.

### Notes
Also includes some changes to `BufferBinaryReader`, getting rid of debug build checks. While investigating this issue, an out of bounds exception was hit. I think such exceptions should also occur in release builds, to help us track down issues where they first happen. This has a minimal performance hit in release builds, especially as most of the checks were already occurring (but just didn't throw an exception in release builds).

### DIFF
--- a/UndertaleModLib/Util/BufferBinaryReader.cs
+++ b/UndertaleModLib/Util/BufferBinaryReader.cs
@@ -32,11 +32,7 @@ namespace UndertaleModLib.Util
                     n = count;
                 if (n <= 0)
                 {
-#if DEBUG
                     throw new IOException("Reading out of chunk bounds");
-#else
-                    return 0;
-#endif
                 }
 
                 if (n <= 8)
@@ -56,11 +52,7 @@ namespace UndertaleModLib.Util
                 int n = Math.Min(_length - _position, buffer.Length);
                 if (n <= 0)
                 {
-#if DEBUG
                     throw new IOException("Reading out of chunk bounds");
-#else
-                    return 0;
-#endif
                 }
 
                 new Span<byte>(_buffer, _position, n).CopyTo(buffer);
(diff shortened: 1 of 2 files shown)


UndertaleModLib/Util/BufferBinaryReader.cs
```diff
@@ -296,13 +285,12 @@ public string ReadGMString()
...
                 res = encoding.GetString(buf);
             }
-            
-#if DEBUG
+
             if (ReadByte() != 0)
+            {
                 throw new IOException("String not null terminated!");
-#else
-            Position++;
```
Miepee, line 304: The diff seems to suggest that something important was removed here?

Miepee (commented):
Should we then also remove the Debug.Assert class?

UndertaleModLib/Util/BufferBinaryReader.cs
colinator27: That's removing the old release code, it now always goes through the ReadByte() path instead.

colinator27:
> Should we then also remove the Debug.Assert class?

I think that's still fine, it just throws an exception always too, I think.

UndertaleModLib/Util/BufferBinaryReader.cs
Miepee: ah right.

Miepee (approved):

PR #1839, opened 2024-07-20
Move 2.0.6 detection to object counting stage

colinator27:
## Description
Moves 2.0.6 version detection to the object counting stage of deserialization. This fixes a regression with object counting that started occurring in a few GMS2 games after https://github.com/UnderminersTeam/UndertaleModTool/pull/1830 was merged, because the version was not detected early enough.

### Caveats
Unlikely to have any side effects.

### Notes
Tested on two games that had the issue, amongst a few that did not. Also, technically speaking, the 2.0.6 detection not occurring in the object counting stage was a hidden bug that could not be triggered until the aforementioned PR was merged.

### DIFF
--- a/UndertaleModLib/UndertaleChunks.cs
+++ b/UndertaleModLib/UndertaleChunks.cs
@@ -1459,6 +1459,8 @@ namespace UndertaleModLib
         public override string Name => "TXTR";
 
         private static bool checkedFor2022_3;
+        private static bool checkedFor2_0_6;
+
         private void CheckFor2022_3And5(UndertaleReader reader)
         {
             // Detect GM2022.3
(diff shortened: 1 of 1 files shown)


UndertaleModLib/UndertaleChunks.cs
```diff
@@ -1576,11 +1576,15 @@ internal override void SerializeChunk(UndertaleWriter writer)
         // GMS 2.0.6.96 is the oldest available runtime version,
         // so this actually could be some other version between GMS 2.0 - 2.0.6.
         // (the oldest copy of "Zeus-Runtime.rss" on web.archive.org has this version as first one)
+        private static bool checkedFor2_0_6;
```
Miepee, line 1579: this bool should probably be located where all the other attributes are instead of right before t his function.

Miepee (approved):

PR #1840, opened 2024-07-21
Alternate right-hand side special parentheses handling

### LINKED ISSUE #1833: Decompiler removes parentheses from math equations when adjacent operations are from the same "level" of order
### Describe the bug

When writing math equations, parentheses are used to "override" the order of operations. This is very important when subtracting and dividing, as changing the order (by removal of parentheses) will result in a difference value. 
The decompiler removes some of these parentheses when variables are present, resulting in undesired values during runtime.

Unsure of how GML handles Order of Operations (Is addition is prioritized before subtraction, or calculated in sequential order?), i ran the following test:
```
    a = 5
    b = 3
    c = 2
    d = b + c
    debug_draw_var("a - (b + c) = ", (a - d))
    debug_draw_var("a - b + c = ", (a - b + c)) 
``` 
Here is the result of these equations drawn in-game:
image
As shown here, addition and subtraction are performed in sequential order (as it should be).

Here are some examples of how the Bleeding Edge decompiler currently handles parentheses in math equations. 
Before decompiler:
```
z = (a - b) + c
z = a - (b + c)
z = a * (b / c)
z = (a * b) / c
z = a + (b - c) * (d / e)
z = (a + b) - (c * d) / e
z = (a + b - c) * (d / e)
z = (a + b) - (c * d / e)
debug_draw_var("5 - 3 + 2 = ", (5 - 3 + 2))
debug_draw_var("5 - (3 + 2) = ", (5 - (3 + 2)))
```
After decompiler:
```
z = a - b + c
z = a - b + c
z = a * b / c
z = a * b / c
z = a + (b - c) * d / e
z = a + b - c * d / e
z = (a + b - c) * d / e
z = a + b - c * d / e
debug_draw_var("5 - 3 + 2 = ", 4)
debug_draw_var("5 - (3 + 2) = ", 0)
```
As can be seen in the `debug_draw_var` functions, the decompiler respects the parentheses when only real numbers are used, during simplification of the code. It is only when variables are introduced that the parentheses are removed.

### Reproducing steps

1. In the Decompiled code editor, write an equation containing math operators of the same "tier", at least one variable, and parentheses encapsulating a latter section of the equation. _e.g. `x = a - (b + c)`_
2. Apply the change. Notice the resulting code now lacks parentheses.
3. Run the game, and see that the desired outcome is not achieved when the equation is calculated.

### Setup Details

UndertaleModTool version: Bleeding Edge (most recent commit as of this posting: https://github.com/UnderminersTeam/UndertaleModTool/commit/c8e1b9babec8c3130e8211f0190b8e2a9c04de4a)
OS: Windows 10 22H2
Game: AM2R, but occurs with any, including new projects

Jacky720:
## Description
This replaces the ExpressionTwo parentheses special check for right-side parentheses with a simpler one that covers more cases. For example, the previous solution correctly adds parentheses to `a - (b - c)` but not `a - (b + c)` or (technically different though functionally the same) `a + (b + c)`. This shouldn't cause any problems with the left side; we aren't going back to `((a + b) + c) + d`. Closes #1833 (not all examples given there will decompile 1:1 with this change, but those that don't match had redundant parentheses to begin with).

### Caveats
N/A

### Notes
Nobody commented on whether this seemed like a good fix in the Discord. Also, Underanalyzer completely redoes this anyway.

### DIFF
--- a/UndertaleModLib/Decompiler/Instructions/Decompiler.ExpressionTwo.cs
+++ b/UndertaleModLib/Decompiler/Instructions/Decompiler.ExpressionTwo.cs
@@ -65,6 +65,10 @@ public static partial class Decompiler
                     _ => 0,
                 };
 
+                // The second argument needs parentheses even if the operation level is equal, not just less; e.g. a - (b + c)
+                // Even some cases that match mathematically will recompile differently without right-hand parentheses e.g. a + (b + c)
+                if (isSecond)
+                    argPriorityLevel--;
 
                 // Suppose we have "(arg1a argOp arg1b) opcode argument2", and are wondering whether the depicted parentheses are needed
                 // If the argument's opcode is more highly-prioritized than our own, such as it being multiplication
@@ -76,10 +80,6 @@ public static partial class Decompiler
                 if (outerPriorityLevel == 0)
                     needsParens = true; // Better safe than sorry
                 
-                // Non-commutative operators may still need parentheses, e.g `a - (b - c)`
-                bool nonCommutative = Opcode is UndertaleInstruction.Opcode.Sub or UndertaleInstruction.Opcode.Div;
-                if (isSecond && nonCommutative && Opcode == argumentAsBinaryExpression.Opcode)
-                    needsParens = true;
             }
 
             return (needsParens ? String.Format("({0})", arg) : arg);


Miepee (approved):
Sorry for not commenting earlier, seems ok.
We just need something not (as) broken until underanalyzer.

PR #1843, opened 2024-07-22
Fix local variable syntax highlighting getting reset when switching tabs

### LINKED ISSUE #1320: The code editor changes the variable highlight color.
### Describe the bug

When you switch to the "Disassembly" tab and return back, the variable colors changes.

### Reproducing steps

https://user-images.githubusercontent.com/45438625/232037419-e8c5233a-cc06-4e14-bb00-911ffea0ab90.mp4

### Setup Details

1. UTMT version - https://github.com/krzys-h/UndertaleModTool/commit/e76dfbfe4812bf5e9001f3cee78725be43239517

colinator27:
## Description
Fixes #1320. The locals list used for syntax highlighting is reset upon changing to the disassembly tab, and previously was only re-populated on the decompiled tab if the code was newly-decompiled. This now re-populates when switching to the decompiled tab, every time.

### Caveats
Sometimes the local variable list can get re-populated twice in a row, but this has no real impact since there's not that many local variables.

### Notes
Also implements a small optimization to re-use the same local list object always: clearing it, rather than allocating a new one. This should prevent some amount of allocations/re-allocations.

### DIFF
--- a/UndertaleModTool/Editors/UndertaleCodeEditor.xaml.cs
+++ b/UndertaleModTool/Editors/UndertaleCodeEditor.xaml.cs
@@ -48,7 +48,7 @@ namespace UndertaleModTool
 
         public UndertaleCode CurrentDisassembled = null;
         public UndertaleCode CurrentDecompiled = null;
-        public List<string> CurrentLocals = null;
+        public List<string> CurrentLocals = new();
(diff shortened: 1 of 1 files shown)


Miepee (approved):
Cant see it from the small snippet here, what causes the list to be populated twice in a row?

PR #1737, opened 2024-04-22
Add differentiation between LTS and non-LTS builds

Jacky720:
## Description
_After the end of October 2022, the GameMaker versions were split into two: LTS and non-LTS. This marked the beginning of the era called "UndertaleModTool breaking for the 400th time"._

This PR adds a Branch enum value to GeneralInfo which can be used to differentiate between these versions without causing confusion with `UndertaleData.IsVersionAtLeast()` (as actually setting the version to 2022.0 would indicate being much older than many of the features therein). It also adds some methods to interact with this, one being `UndertaleData.IsNonLTSVersionAtLeast()`, and implements its usage in regards to Particle Systems and Spine Sprites, both of which are not present on the current LTS despite other fixes. Closes #1703, closes #1729.

### Caveats
Using an enum is theoretically extensible for other branches, but that seems improbable. More likely we'll just get 2024.0 and have to adjust around that; mostly I just don't want to go back to version booleans. The `UndertaleData.SetLTS()` method is also weird (boolean input), and this all renders the estimated version in the window title completely off. Oh, and it can't be set through the GUI.

### Notes
Due to all the caveats (and wanting further input on how to resolve them), I'm marking this as draft.

Issue #1729, opened 2024-04-17
Misaligned Data and Load Errors when Opening Latest ANTONBLAST Demo


DogToon64:
### Describe the bug

Whenever I try to open ANTONBLAST's latest demo, Dynamite Demo 2.0, I get these error messages:
![ABDD2 0 - Error 1](https://github.com/UnderminersTeam/UndertaleModTool/assets/159098439/e97d2076-dc55-4b41-9158-742817bc53ad)
![ABDD2 0 - Error 2](https://github.com/UnderminersTeam/UndertaleModTool/assets/159098439/a99bdd4a-87ac-41f5-9f26-c4f88d4dcc87)
![ABDD2 0 - Error 3](https://github.com/UnderminersTeam/UndertaleModTool/assets/159098439/f4391453-81a3-4e7d-89bd-601289ba0c11)

### Reproducing steps

1. Open ANTONBLAST's data.win file with Undertale Mod Tool
2. Displays error messages

### Setup Details

1. [Undertale v.0.5.1.0 (Bleeding Edge build, GUI-windows-latest-Release-isBundled-true-isSingleFile-true)](https://github.com/UnderminersTeam/UndertaleModTool/releases/tag/bleeding-edge)
2. Windows 10 x64
3. [ANTONBLAST (Demo)](https://store.steampowered.com/app/1887400/ANTONBLAST/)

TrashBandatcoot:
I'll add that the Nintendo Switch version displays a different error message.

![afbeelding](https://github.com/UnderminersTeam/UndertaleModTool/assets/49070403/e79bba2c-9140-417c-9b5d-e9dc847e28b5)

mike309game:
> I'll add that the Nintendo Switch version displays a different error message.
> 
[...]

PR #1848, opened 2024-07-27
Fix 2022.0.2 detection

Jacky720:
## Description
This ACTUALLY fixes 2022.0.2 (and presumably 2022.0.3) detection. Our previous test case, the AntonBlast Demo (#1729), was erroneously detecting as "2023.2, but actually LTS 2022.0" instead of "2023.6, but actually LTS 2022.0", which loaded successfully but doesn't extend to other test cases (and loaded the SDFSpread value instead of LineHeight). This error has been present since merging #1737.

### Caveats
N/A

### Notes
A test LTS project which now loads, built by Colinator27, is attached.
[TestLTSProject.zip](https://github.com/user-attachments/files/16397539/TestLTSProject.zip)


colinator27:
Additionally verified that this works with extensions, and on a few other games (2023.4 and 2023.11) to make sure this doesn't break anything.

colinator27 (approved):

PR #1850, opened 2024-07-28
Supply script file paths, better script exceptions

colinator27:
## Description
Based on code from https://github.com/UnderminersTeam/UndertaleModTool/pull/1504. Modified to fix the root cause of the problem (file paths/encodings not being specified in ScriptOptions), rather than using regular expressions to work around it.

Also includes the improved script stack traces/exception handling, but further cleaned up and commented.

### Caveats
As mentioned in the above PR, this does make exceptions slightly more verbose (even for single scripts), but this is probably a welcome change.

### Notes
Tested to work with built-in scripts and [UMP](https://github.com/nhaar/UMP).

### DIFF
--- a/UndertaleModCli/Program.UMTLibInherited.cs
+++ b/UndertaleModCli/Program.UMTLibInherited.cs
@@ -4,6 +4,7 @@ using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Runtime.InteropServices;
+using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
@@ -586,7 +587,7 @@ public partial class Program : IScriptInterface
         {
             CancellationTokenSource source = new CancellationTokenSource(100);
             CancellationToken token = source.Token;
-            CSharpScript.EvaluateAsync(File.ReadAllText(path), CliScriptOptions, this, typeof(IScriptInterface), token);
+            CSharpScript.EvaluateAsync(File.ReadAllText(path, Encoding.UTF8), CliScriptOptions.WithFilePath(path).WithFileEncoding(Encoding.UTF8), this, typeof(IScriptInterface), token);
         }
         catch (CompilationErrorException exc)
         {
--- a/UndertaleModCli/Program.cs
+++ b/UndertaleModCli/Program.cs
@@ -712,7 +712,7 @@ public partial class Program : IScriptInterface
         string lines;
         try
         {
-            lines = File.ReadAllText(path);
+            lines = File.ReadAllText(path, Encoding.UTF8);
         }
         catch (Exception exc)
         {
@@ -739,7 +739,7 @@ public partial class Program : IScriptInterface
 
         try
         {
-            CSharpScript.EvaluateAsync(code, CliScriptOptions, this, typeof(IScriptInterface)).GetAwaiter().GetResult();
+            CSharpScript.EvaluateAsync(code, CliScriptOptions.WithFilePath(scriptFile ?? "").WithFileEncoding(Encoding.UTF8), this, typeof(IScriptInterface)).GetAwaiter().GetResult();
             ScriptExecutionSuccess = true;
             ScriptErrorMessage = "";
         }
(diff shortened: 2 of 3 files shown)


UndertaleModCli/Program.UMTLibInherited.cs
```diff
@@ -586,7 +587,7 @@ public bool LintUMTScript(string path)
         {
             CancellationTokenSource source = new CancellationTokenSource(100);
             CancellationToken token = source.Token;
-            CSharpScript.EvaluateAsync(File.ReadAllText(path), CliScriptOptions, this, typeof(IScriptInterface), token);
+            CSharpScript.EvaluateAsync(File.ReadAllText(path, Encoding.UTF8), CliScriptOptions.WithFilePath(path).WithFileEncoding(Encoding.UTF8), this, typeof(IScriptInterface), token);
```
Miepee, line 590: Why do we need to specify utf8 here? what if someone has a script file thats saved as ansi?

UndertaleModCli/Program.cs
```diff
@@ -712,7 +712,7 @@ private void RunCSharpFile(string path)
         string lines;
         try
         {
-            lines = File.ReadAllText(path);
+            lines = File.ReadAllText(path, Encoding.UTF8);
```
Miepee, line 715: ditto

UndertaleModCli/Program.cs
```diff
@@ -739,7 +739,7 @@ private void RunCSharpCode(string code, string scriptFile = null)
 
         try
         {
-            CSharpScript.EvaluateAsync(code, CliScriptOptions, this, typeof(IScriptInterface)).GetAwaiter().GetResult();
+            CSharpScript.EvaluateAsync(code, CliScriptOptions.WithFilePath(scriptFile ?? "").WithFileEncoding(Encoding.UTF8), this, typeof(IScriptInterface)).GetAwaiter().GetResult();
```
Miepee, line 742: .

UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -2586,8 +2636,7 @@ public async Task RunScript(string path)
 
         private async Task RunScriptNow(string path)
         {
-            string scriptText = $"#line 1 \"{path}\"\n" + File.ReadAllText(path);
-            Debug.WriteLine(path);
+            string scriptText = $"#line 1 \"{path}\"\n" + File.ReadAllText(path, Encoding.UTF8);
```
Miepee, line 2639: .

UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -2598,7 +2647,7 @@ private async Task RunScriptNow(string path)
                 ScriptPath = path;
 
                 string compatScriptText = Regex.Replace(scriptText, @"\bDecompileContext(?!\.)\b", "GlobalDecompileContext", RegexOptions.None);
-                object result = await CSharpScript.EvaluateAsync(compatScriptText, scriptOptions, this, typeof(IScriptInterface));
+                object result = await CSharpScript.EvaluateAsync(compatScriptText, scriptOptions.WithFilePath(path).WithFileEncoding(Encoding.UTF8), this, typeof(IScriptInterface));
```
Miepee, line 2650: .

Miepee (commented):
How does the CLI handle this?

UndertaleModCli/Program.UMTLibInherited.cs
colinator27: I wasn't sure of this at the original time of writing, but apparently `File.ReadAllText` defaults to UTF-8 in the first place, so this is just being explicit about it. Technically we could try to handle other encodings some day, but we haven't up until this point. Although, UTF-8 should be backwards-compatible with other encodings like ASCII and subsets of ANSI.

colinator27:
> How does the CLI handle this?

CLI just specifies the file path and encoding now, so `#load` will work properly there too. There's no updated exception handling for it like there is in the GUI, though (it'll be the same as before).

Miepee (approved):

PR #1856, opened 2024-08-08
Use buffer for reading JSON in UndertaleWithJSONs

Dobby233Liu:
## Description
This makes the loading of lang JSONs almost instantaneous

### Caveats
Unclear, it works in v1.08 and v1.11

### Notes
Code copied from DR Ch1&2


colinator27:
We may hold off on merging this until after the next major release, but it looks good to me! These changes will have to be merged with the additional rewrite of this script that occurred on the `underanalyzer` branch, though it should be easy I think.

Miepee (approved):
Nothing block this, so I'll merge.

PR #1862, opened 2024-08-12
Replace GML decompiler with Underanalyzer

colinator27:
## Description
Removes the current GML decompiler included with UndertaleModTool, replacing it with calls to the [Underanalyzer](https://github.com/UnderminersTeam/Underanalyzer) library. This is intended to drastically improve accuracy and stability on valid GML bytecode. This also provides decompiler settings, which can be tailored per user, and in theory, be used to prevent C# scripts that depend on exact decompiler output from breaking in the future.

C# scripts have been updated to account for the change in decompiler. Most should be fully functional, but could use some additional stability testing.

The built-in GML code corrections have all been removed, as the decompiler is accurate enough to supersede them. Support for code corrections has not been removed, so they will still function as normal if re-added.

Some small GML compiler upgrades have been made to complement the decompiler improvements, including function constructor support, default argument values, enum support, CSS hex color literals, named instance IDs, `throw`, `new`, amongst some other control flow fixes. Syntax highlighting in the code editor has been adjusted as well.

External JSON game-specific data is now supported through this PR, which currently is only used for Underanalyzer type definition files. Features related to these can be expanded on in the future, after this PR.

This PR also completes support for bytecode changes observed in GameMaker 2024.6, which also updates the assembly syntax, adding `[function]` and `[variable]`.

A few other miscellaneous fixes and changes were made, such as some optimization/cleanup in assembly parsing, and removing extra arguments to `EventHandlerFor` (simplifying its usage).

### Caveats
Automatic type inference support is not present in the Underanalyzer decompiler as of right now, but it's probably something we can live without, as external JSON type definitions are now supported through this PR.

Some partially corrupt code may no longer decompile properly, but this should not be an issue, as we are not going to target any obfuscation techniques, and any other cases (e.g. manual assembly editing) are rare.

Currently, Underanalyzer is included as a submodule. We may want to switch to a different method in the future (e.g. using NuGet).

### Notes
This code will be under review alongside Underanalyzer's library itself.

This branch has been under active testing by multiple users for a number of weeks, and stability seems good right now.


hallo1126:
Requesting this to be included asap as many games fail with main, which this branch solves

Miepee:
> Requesting this to be included asap as many games fail with main, which this branch solves

Unfortunately, that's not how reviewing works.

Voidymite:
Would enjoy if this happened as Underanalyzer has more support and is overall better at GML decompilation

PR #1864, opened 2024-08-12
Some serialization cleanup, optimization, fixes

colinator27:
## Description
- Removed more cases of `#if DEBUG`, to ensure consistent behavior on Release builds.
- Added more safety checks to binary readers, and to deserialization in general (preventing unintended negative lengths/jumps that could lead to infinite loops).
- ~~Fixed `EmSize` on fonts (when in float form) being accidentally always parsed as 0, due to a typo.~~
- Completely refactored code instruction deserialization, removing many calls and I/O operations.
- Commented and reformatted code instruction deserialization/serialization.
- Fixed QOI padding being added when saving, when none seems to have ever actually existed.

### Caveats
Has potential to break some instruction parsing if I typo'd somewhere, but I did test it on multiple games (including Undertale 1.00, a bytecode 14 game), and it seems to work fine. If it does break, it'll be pretty obvious, as re-saving a data.win file will show differences in the code.

### Notes
Probably too risky to merge into stable, ~~although in the event that we make a minor release, we may want to backport the `EmSize` typo fix, as that is pretty diabolical~~. It affects the Deltarune Chapter 1 & 2 demo, version 1.15 (which I tested this PR on).

### DIFF
--- a/UndertaleModLib/Models/UndertaleCode.cs
+++ b/UndertaleModLib/Models/UndertaleCode.cs
@@ -110,6 +110,8 @@ public class UndertaleInstruction : UndertaleObject
             _ => throw new IOException("Unknown opcode " + op.ToString().ToUpper(CultureInfo.InvariantCulture)),
         };
     }
+
+    // Converts from bytecode 14 instruction opcodes to modern opcodes
     private static byte ConvertInstructionKind(byte kind)
     {
         kind = kind switch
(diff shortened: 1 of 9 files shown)


BenjaminUrquhart:
The `EmSize` typo prevents the Xbox version of Undertale from saving 1 to 1. Might be worth its own small PR.

Miepee:
the emsize fix should be removed from here now

Miepee (commented):
Only reviewed the changes in UndertaleCode so you have something to work with.
Please split the PR up into the following
- #if debug changes
- code commenting in UTcode serialization
- code refactor in UTcode deserialization
- binary reader checks
- QOI padding fix.

PR #1870, opened 2024-08-14
Rewrite texture handling

colinator27:
## Description
Rewrites all of the core texture code to use a new immutable `GMImage` class, which can be one of four texture types: PNG, raw BGRA pixel data, QOI, or BZip2 + QOI. This class is used as a replacement to the existing `byte[]` containing compressed PNG data always.

The major benefit of this is that load/save times on games using non-PNG textures are now *much* faster (some games are at least 5 seconds faster on either end, on my machine). QOI and/or BZ2 textures are now decoded/decompressed on the fly, rather than during the (de)serialization process.

This also mostly replaces `System.Drawing.Common` calls with ones to `ImageMagick` instead. In UndertaleModLib, it's used only for PNG encoding/decoding, but I also rewrote a tiny bit of code on the UI side. This should mean the library code itself is cross-platform. Additionally, I performed some tests and found that exporting PNGs with the new UI code preserves colors with 0 alpha.

### Caveats
This breaks pretty much every script that interacts with texture data, but that will be resolved in an upcoming PR.

There's also one further memory optimization that can be made with BZ2 + QOI textures, if we can decompress the first 8 bytes of the BZ2 data only. Essentially, it involves finding the size of the inner QOI data, which can be used to initialize the capacity of the buffer used to decompress the BZ2 data later (when called to do so, on the fly). The performance impact is probably not extreme, though - it would prevent a few reallocations at most. It's something maybe worth revisiting in the future.

### Notes
Tested on a number of games, including ones with PNG textures, BZ2 + QOI textures, and pre/post GameMaker 2022.5. I haven't yet tested QOI-only games, but the code should theoretically work with them. Files have loaded/saved byte-for-byte, besides a bug in the FONT chunk that's fixed by the serialization cleanup PR (#1864).

I tested collision mask saving/loading, the room editor, the sprite editor, the embedded texture editor (saving/loading), and the texture page item editor (saving/loading).

On the UI front, instances of `BitmapSource` are cached for as long as the garbage collector doesn't get rid of them from memory, which can prevent unnecessary image decodes.

### DIFF
--- a/UndertaleModCli/Program.cs
+++ b/UndertaleModCli/Program.cs
@@ -658,7 +658,7 @@ public partial class Program : IScriptInterface
             if (Verbose)
                 Console.WriteLine($"Dumping {texture.Name}");
             using FileStream fs = new($"{directory}/{texture.Name.Content}.png", FileMode.Create);
-            fs.Write(texture.TextureData.Image.ConvertToPng().ToSpan());
+            texture.TextureData.Image.SavePng(fs);
         }
     }
 
--- a/UndertaleModCli/Program.cs
+++ b/UndertaleModCli/Program.cs
@@ -657,7 +657,8 @@ public partial class Program : IScriptInterface
         {
             if (Verbose)
                 Console.WriteLine($"Dumping {texture.Name}");
-            File.WriteAllBytes($"{directory}/{texture.Name.Content}.png", texture.TextureData.TextureBlob);
+            using FileStream fs = new($"{directory}/{texture.Name.Content}.png", FileMode.Create);
+            fs.Write(texture.TextureData.Image.ConvertToPng().ToSpan());
         }
     }
 
@@ -700,7 +701,7 @@ public partial class Program : IScriptInterface
         if (Verbose)
             Console.WriteLine("Replacing " + textureEntry);
 
-        texture.TextureData.TextureBlob = File.ReadAllBytes(fileToReplace.FullName);
+        texture.TextureData.Image = GMImage.FromPng(File.ReadAllBytes(fileToReplace.FullName));
     }
 
     /// <summary>
--- a/UndertaleModLib/Models/UndertaleEmbeddedTexture.cs
+++ b/UndertaleModLib/Models/UndertaleEmbeddedTexture.cs
@@ -334,9 +334,11 @@ public class UndertaleEmbeddedTexture : UndertaleNamedResource, IDisposable
         public bool FormatBZ2 => _image.Format is GMImage.ImageFormat.Bz2Qoi;
 
         /// <summary>
-        /// If located within a data file, this is the maximum end position of the image data (or start of the next texture blob).
-        /// All data between the actual end position and this maximum end position should be padding.
+        /// If located within a data file, this is the upper bound on the end position of the image data (or start of the next texture blob).
         /// </summary>
+        /// <remarks>
+        /// All data between the actual end position and this maximum end position should be 0x00 byte padding.
+        /// </remarks>
         private int _maxEndOfStreamPosition { get; set; } = -1;
 
         /// <inheritdoc />
--- a/UndertaleModLib/Models/UndertaleEmbeddedTexture.cs
+++ b/UndertaleModLib/Models/UndertaleEmbeddedTexture.cs
@@ -309,7 +309,7 @@ public class UndertaleEmbeddedTexture : UndertaleNamedResource, IDisposable
             set
             {
                 _image = value;
-                OnPropertyChanged(nameof(Image));
+                OnPropertyChanged();
             }
         }
 
--- a/UndertaleModLib/Models/UndertaleEmbeddedTexture.cs
+++ b/UndertaleModLib/Models/UndertaleEmbeddedTexture.cs
@@ -3,7 +3,6 @@ using System;
 using System.Buffers.Binary;
 using System.ComponentModel;
 using System.Drawing;
-using System.Drawing.Imaging;
 using System.IO;
 using System.Linq;
 using System.Runtime.CompilerServices;
@@ -57,7 +56,6 @@ public class UndertaleEmbeddedTexture : UndertaleNamedResource, IDisposable
         set => _textureData = value;
     }
     private TexData _textureData = new TexData();
-
 
     /// <summary>
     /// Helper variable for whether or not this texture is to be stored externally or not.
@@ -230,20 +228,9 @@ public class UndertaleEmbeddedTexture : UndertaleNamedResource, IDisposable
         }
     }
 
-    // 1x1 black pixel in PNG format
-    private static TexData _placeholderTexture = new()
-    {
-        TextureBlob = new byte[]
-        {
-            0x89, 0x50, 0x4E, 0x47, 0x0D, 0x0A, 0x1A, 0x0A, 0x00, 0x00, 0x00, 0x0D, 0x49, 0x48, 0x44, 0x52, 0x00, 0x00, 0x00, 0x01, 
-            0x00, 0x00, 0x00, 0x01, 0x08, 0x02, 0x00, 0x00, 0x00, 0x90, 0x77, 0x53, 0xDE, 0x00, 0x00, 0x00, 0x01, 0x73, 0x52, 0x47, 
-            0x42, 0x00, 0xAE, 0xCE, 0x1C, 0xE9, 0x00, 0x00, 0x00, 0x04, 0x67, 0x41, 0x4D, 0x41, 0x00, 0x00, 0xB1, 0x8F, 0x0B, 0xFC, 
-            0x61, 0x05, 0x00, 0x00, 0x00, 0x09, 0x70, 0x48, 0x59, 0x73, 0x00, 0x00, 0x0E, 0xC3, 0x00, 0x00, 0x0E, 0xC3, 0x01, 0xC7, 
-            0x6F, 0xA8, 0x64, 0x00, 0x00, 0x00, 0x0C, 0x49, 0x44, 0x41, 0x54, 0x18, 0x57, 0x63, 0x60, 0x60, 0x60, 0x00, 0x00, 0x00, 
-            0x04, 0x00, 0x01, 0x5C, 0xCD, 0xFF, 0x69, 0x00, 0x00, 0x00, 0x00, 0x49, 0x45, 0x4E, 0x44, 0xAE, 0x42, 0x60, 0x82
-        }
-    };
-    private static object _textureLoadLock = new();
+    // 1x1 blank image
+    private static readonly TexData _placeholderTexture = new() { Image = new GMImage(1, 1) };
+    private static readonly object _textureLoadLock = new();
 
     /// <summary>
     /// Attempts to load the corresponding external texture. Should only happen in 2022.9 and above.
@@ -311,62 +298,41 @@ public class UndertaleEmbeddedTexture : UndertaleNamedResource, IDisposable
     /// </summary>
     public class TexData : UndertaleObject, INotifyPropertyChanged, IDisposable
     {
-        private byte[] _textureBlob;
+        private GMImage _image;
         private static MemoryStream sharedStream;
 
         /// <summary>
-        /// The image data of the texture.
+        /// The underlying image of the texture.
         /// </summary>
-        public byte[] TextureBlob
-        {
-            get => _textureBlob;
+        public GMImage Image 
+        { 
+            get => _image;
             set
             {
-                _textureBlob = value;
-                OnPropertyChanged();
+                _image = value;
+                OnPropertyChanged(nameof(Image));
             }
         }
 
         /// <summary>
         /// The width of the texture.
-        /// In case of an invalid texture data, this will be <c>-1</c>.
         /// </summary>
-        public int Width
-        {
-            get
-            {
-                if (_textureBlob is null || _textureBlob.Length < 24)
-                    return -1;
+        public int Width => _image.Width;
 
-                ReadOnlySpan<byte> span = _textureBlob.AsSpan();
-                return BinaryPrimitives.ReadInt32BigEndian(span[16..20]);
-            }
-        }
         /// <summary>
         /// The height of the texture.
-        /// In case of an invalid texture data, this will be <c>-1</c>.
         /// </summary>
-        public int Height
-        {
-            get
-            {
-                if (_textureBlob is null || _textureBlob.Length < 24)
-                    return -1;
-
-                ReadOnlySpan<byte> span = _textureBlob.AsSpan();
-                return BinaryPrimitives.ReadInt32BigEndian(span[20..24]);
-            }
-        }
+        public int Height => _image.Height;
 
         /// <summary>
         /// Whether this texture uses the QOI format.
         /// </summary>
-        public bool FormatQOI { get; set; } = false;
+        public bool FormatQOI => _image.Format is GMImage.ImageFormat.Qoi or GMImage.ImageFormat.Bz2Qoi;
 
         /// <summary>
         /// Whether this texture uses the BZ2 format. (Always used in combination with QOI.)
         /// </summary>
-        public bool FormatBZ2 { get; set; } = false;
+        public bool FormatBZ2 => _image.Format is GMImage.ImageFormat.Bz2Qoi;
 
         /// <inheritdoc />
         public event PropertyChangedEventHandler PropertyChanged;
@@ -378,21 +344,6 @@ public class UndertaleEmbeddedTexture : UndertaleNamedResource, IDisposable
         {
             PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(name));
         }
-
-        /// <summary>
-        /// Header used for PNG files.
-        /// </summary>
-        public static readonly byte[] PNGHeader = { 137, 80, 78, 71, 13, 10, 26, 10 };
-
-        /// <summary>
-        /// Header used for GameMaker QOI + BZ2 files.
-        /// </summary>
-        public static readonly byte[] QOIAndBZip2Header = { 50, 122, 111, 113 };
-
-        /// <summary>
-        /// Header used for GameMaker QOI files.
-        /// </summary>
-        public static readonly byte[] QOIHeader = { 102, 105, 111, 113 };
 
         /// <summary>
         /// Frees up <see cref="sharedStream"/> from memory.
@@ -412,42 +363,20 @@ public class UndertaleEmbeddedTexture : UndertaleNamedResource, IDisposable
         /// <inheritdoc />
         public void Serialize(UndertaleWriter writer)
         {
-            Serialize(writer, writer.undertaleData.IsVersionAtLeast(2022, 3), writer.undertaleData.IsVersionAtLeast(2022, 5));
+            Serialize(writer, writer.undertaleData.IsVersionAtLeast(2022, 5));
         }
 
         /// <summary>
         /// Serializes the texture to any type of writer (can be any destination file).
         /// </summary>
-        public void Serialize(FileBinaryWriter writer, bool gm2022_3, bool gm2022_5)
+        public void Serialize(FileBinaryWriter writer, bool gm2022_5)
         {
-            if (FormatQOI)
+            if (Image.Format == GMImage.ImageFormat.RawBgra)
             {
-                if (FormatBZ2)
-                {
-                    writer.Write(QOIAndBZip2Header);
-
-                    // Encode the PNG data back to QOI+BZip2
-                    using Bitmap bmp = TextureWorker.GetImageFromByteArray(TextureBlob);
-                    writer.Write((short)bmp.Width);
-                    writer.Write((short)bmp.Height);
-                    byte[] qoiData = QoiConverter.GetArrayFromImage(bmp, gm2022_3 ? 0 : 4);
-                    using MemoryStream input = new MemoryStream(qoiData);
-                    if (sharedStream.Length != 0)
-                        sharedStream.Seek(0, SeekOrigin.Begin);
-                    BZip2.Compress(input, sharedStream, false, 9);
-                    if (gm2022_5)
-                        writer.Write((uint)qoiData.Length);
-                    writer.Write(sharedStream.GetBuffer().AsSpan()[..(int)sharedStream.Position]);
-                }
-                else
-                {
-                    // Encode the PNG data back to QOI
-                    using Bitmap bmp = TextureWorker.GetImageFromByteArray(TextureBlob);
-                    writer.Write(QoiConverter.GetSpanFromImage(bmp, gm2022_3 ? 0 : 4));
-                }
+                throw new Exception("Unexpected raw RGBA image");
             }
-            else
-                writer.Write(TextureBlob);
+
+            Image.WriteToBinaryWriter(writer, gm2022_5);
         }
 
         /// <inheritdoc />
@@ -462,69 +391,7 @@ public class UndertaleEmbeddedTexture : UndertaleNamedResource, IDisposable
         public void Unserialize(IBinaryReader reader, bool gm2022_5)
         {
             sharedStream ??= new();
-
-            long startAddress = reader.Position;
-
-            byte[] header = reader.ReadBytes(8);
-            if (!header.SequenceEqual(PNGHeader))
-            {
-                reader.Position = startAddress;
-
-                if (header.Take(4).SequenceEqual(QOIAndBZip2Header))
-                {
-                    FormatQOI = true;
-                    FormatBZ2 = true;
-
-                    // Don't really care about the width/height, so skip them, as well as header
-                    reader.Position += (uint)(gm2022_5 ? 12 : 8);
-
-                    // Need to fully decompress and convert the QOI data to PNG for compatibility purposes (at least for now)
-                    if (sharedStream.Length != 0)
-                        sharedStream.Seek(0, SeekOrigin.Begin);
-                    BZip2.Decompress(reader.Stream, sharedStream, false);
-                    ReadOnlySpan<byte> decompressed = sharedStream.GetBuffer().AsSpan()[..(int)sharedStream.Position];
-                    using Bitmap bmp = QoiConverter.GetImageFromSpan(decompressed);
-                    sharedStream.Seek(0, SeekOrigin.Begin);
-                    bmp.Save(sharedStream, ImageFormat.Png);
-                    TextureBlob = new byte[(int)sharedStream.Position];
-                    sharedStream.Seek(0, SeekOrigin.Begin);
-                    sharedStream.Read(TextureBlob, 0, TextureBlob.Length);
-                    return;
-                }
-                else if (header.Take(4).SequenceEqual(QOIHeader))
-                {
-                    FormatQOI = true;
-                    FormatBZ2 = false;
-
-                    // Need to convert the QOI data to PNG for compatibility purposes (at least for now)
-                    using Bitmap bmp = QoiConverter.GetImageFromStream(reader.Stream);
-                    if (sharedStream.Length != 0)
-                        sharedStream.Seek(0, SeekOrigin.Begin);
-                    bmp.Save(sharedStream, ImageFormat.Png);
-                    TextureBlob = new byte[(int)sharedStream.Position];
-                    sharedStream.Seek(0, SeekOrigin.Begin);
-                    sharedStream.Read(TextureBlob, 0, TextureBlob.Length);
-                    return;
-                }
-                else
-                    throw new IOException("Didn't find PNG or QOI+BZip2 header");
-            }
-
-            // There is no length for the PNG anywhere as far as I can see
-            // The only thing we can do is parse the image to find the end
-            while (true)
-            {
-                // PNG is big endian and BinaryRead can't handle that (damn)
-                uint len = (uint)reader.ReadByte() << 24 | (uint)reader.ReadByte() << 16 | (uint)reader.ReadByte() << 8 | (uint)reader.ReadByte();
-                uint type = reader.ReadUInt32();
-                reader.Position += len + 4;
-                if (type == 0x444e4549) // 0x444e4549 -> "IEND"
-                    break;
-            }
-
-            long length = reader.Position - startAddress;
-            reader.Position = startAddress;
-            TextureBlob = reader.ReadBytes((int)length);
+            Image = GMImage.FromBinaryReader(reader, sharedStream, gm2022_5);
         }
 
 
@@ -533,7 +400,7 @@ public class UndertaleEmbeddedTexture : UndertaleNamedResource, IDisposable
         {
             GC.SuppressFinalize(this);
 
-            _textureBlob = null;
+            _image = null;
             ClearSharedStream();
         }
     }
--- a/UndertaleModLib/Models/UndertaleEmbeddedTexture.cs
+++ b/UndertaleModLib/Models/UndertaleEmbeddedTexture.cs
@@ -335,8 +335,9 @@ public class UndertaleEmbeddedTexture : UndertaleNamedResource, IDisposable
 
         /// <summary>
         /// If located within a data file, this is the maximum end position of the image data (or start of the next texture blob).
+        /// All data between the actual end position and this maximum end position should be padding.
         /// </summary>
-        public int MaxEndOfStreamPosition { get; set; } = -1;
+        private int _maxEndOfStreamPosition { get; set; } = -1;
 
         /// <inheritdoc />
         public event PropertyChangedEventHandler PropertyChanged;
@@ -371,20 +372,34 @@ public class UndertaleEmbeddedTexture : UndertaleNamedResource, IDisposable
         /// <inheritdoc />
         public void Unserialize(UndertaleReader reader)
         {
-            if (MaxEndOfStreamPosition == -1)
-            {
-                throw new Exception("Expected max end of stream position to be set before unserializing");
-            }
-
-            Unserialize(reader, MaxEndOfStreamPosition, reader.undertaleData.IsVersionAtLeast(2022, 5));
+            Unserialize(reader, _maxEndOfStreamPosition, reader.undertaleData.IsVersionAtLeast(2022, 5));
         }
 
         /// <summary>
         /// Unserializes the texture from any type of reader (can be from any source).
         /// </summary>
+        /// <param name="reader"><see cref="IBinaryReader"/> to read the texture's image from.</param>
+        /// <param name="maxEndOfStreamPosition">Upper bound on the end of the texture's image data (e.g., for padding).</param>
+        /// <param name="gm2022_5">Whether to unserialize the image data using GameMaker 2022.5+ format.</param>
         public void Unserialize(IBinaryReader reader, long maxEndOfStreamPosition, bool gm2022_5)
         {
+            if (maxEndOfStreamPosition == -1)
+            {
+                throw new Exception("Expected max end of stream position to be set before unserializing");
+            }
+
             Image = GMImage.FromBinaryReader(reader, maxEndOfStreamPosition, gm2022_5);
+        }
+
+        /// <summary>
+        /// Sets the upper bound on the position of the end of the image stream, for use when loading a full data file.
+        /// </summary>
+        /// <remarks>
+        /// All data between the actual end position and this maximum end position should be padding (zero bytes).
+        /// </remarks>
+        public void SetMaxEndOfStreamPosition(int position)
+        {
+            _maxEndOfStreamPosition = position;
         }
 
 
--- a/UndertaleModLib/Models/UndertaleEmbeddedTexture.cs
+++ b/UndertaleModLib/Models/UndertaleEmbeddedTexture.cs
@@ -55,7 +55,7 @@ public class UndertaleEmbeddedTexture : UndertaleNamedResource, IDisposable
         get => _textureData ??= LoadExternalTexture();
         set => _textureData = value;
     }
-    private TexData _textureData = new TexData();
+    private TexData _textureData = new();
 
     /// <summary>
     /// Helper variable for whether or not this texture is to be stored externally or not.
@@ -260,7 +260,7 @@ public class UndertaleEmbeddedTexture : UndertaleNamedResource, IDisposable
                 using FileStream fs = new(path, FileMode.Open);
                 using FileBinaryReader fbr = new(fs);
                 texData = new TexData();
-                texData.Unserialize(fbr, true);
+                texData.Unserialize(fbr, fs.Length, true);
                 TextureLoaded = true;
             }
             catch (IOException)
@@ -299,7 +299,6 @@ public class UndertaleEmbeddedTexture : UndertaleNamedResource, IDisposable
     public class TexData : UndertaleObject, INotifyPropertyChanged, IDisposable
     {
         private GMImage _image;
-        private static MemoryStream sharedStream;
 
         /// <summary>
         /// The underlying image of the texture.
@@ -334,6 +333,11 @@ public class UndertaleEmbeddedTexture : UndertaleNamedResource, IDisposable
         /// </summary>
         public bool FormatBZ2 => _image.Format is GMImage.ImageFormat.Bz2Qoi;
 
+        /// <summary>
+        /// If located within a data file, this is the maximum end position of the image data (or start of the next texture blob).
+        /// </summary>
+        public int MaxEndOfStreamPosition { get; set; } = -1;
+
         /// <inheritdoc />
         public event PropertyChangedEventHandler PropertyChanged;
         
@@ -344,21 +348,6 @@ public class UndertaleEmbeddedTexture : UndertaleNamedResource, IDisposable
         {
             PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(name));
         }
-
-        /// <summary>
-        /// Frees up <see cref="sharedStream"/> from memory.
-        /// </summary>
-        public static void ClearSharedStream()
-        {
-            sharedStream?.Dispose();
-            sharedStream = null;
-        }
-
-        /// <summary>
-        /// Initializes <see cref="sharedStream"/> with a specified initial size.
-        /// </summary>
-        /// <param name="size">Initial size of <see cref="sharedStream"/> in bytes</param>
-        public static void InitSharedStream(int size) => sharedStream = new(size);
 
         /// <inheritdoc />
         public void Serialize(UndertaleWriter writer)
@@ -382,16 +371,20 @@ public class UndertaleEmbeddedTexture : UndertaleNamedResource, IDisposable
         /// <inheritdoc />
         public void Unserialize(UndertaleReader reader)
         {
-            Unserialize(reader, reader.undertaleData.IsVersionAtLeast(2022, 5));
+            if (MaxEndOfStreamPosition == -1)
+            {
+                throw new Exception("Expected max end of stream position to be set before unserializing");
+            }
+
+            Unserialize(reader, MaxEndOfStreamPosition, reader.undertaleData.IsVersionAtLeast(2022, 5));
         }
 
         /// <summary>
         /// Unserializes the texture from any type of reader (can be from any source).
         /// </summary>
-        public void Unserialize(IBinaryReader reader, bool gm2022_5)
+        public void Unserialize(IBinaryReader reader, long maxEndOfStreamPosition, bool gm2022_5)
         {
-            sharedStream ??= new();
-            Image = GMImage.FromBinaryReader(reader, sharedStream, gm2022_5);
+            Image = GMImage.FromBinaryReader(reader, maxEndOfStreamPosition, gm2022_5);
         }
 
 
@@ -401,7 +394,6 @@ public class UndertaleEmbeddedTexture : UndertaleNamedResource, IDisposable
             GC.SuppressFinalize(this);
 
             _image = null;
-            ClearSharedStream();
         }
     }
 }
--- a/UndertaleModLib/Models/UndertaleTexturePageItem.cs
+++ b/UndertaleModLib/Models/UndertaleTexturePageItem.cs
@@ -146,13 +146,12 @@ public class UndertaleTexturePageItem : UndertaleNamedResource, INotifyPropertyC
     /// Replaces the current image of this texture page item to hold a new image.
     /// </summary>
     /// <param name="replaceImage">The new image that shall be applied to this texture page item.</param>
-    /// <param name="disposeImage">Whether to dispose <paramref name="replaceImage"/> afterwards.</param>
-    public void ReplaceTexture(MagickImage replaceImage, bool disposeImage = true)
+    public void ReplaceTexture(MagickImage replaceImage)
     {
         // Resize image to bounds on texture page
-        IMagickImage<byte> finalImage = TextureWorker.ResizeImage(replaceImage, SourceWidth, SourceHeight);
+        using IMagickImage<byte> finalImage = TextureWorker.ResizeImage(replaceImage, SourceWidth, SourceHeight);
 
-        // Apply the image to the TexturePage.
+        // Apply the image to the texture page
         lock (TexturePage.TextureData)
         {
             using TextureWorker worker = new();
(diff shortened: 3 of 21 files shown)


UndertaleModCli/Program.cs
```diff
@@ -657,7 +657,8 @@ private void DumpAllTextures()
         {
             if (Verbose)
                 Console.WriteLine($"Dumping {texture.Name}");
-            File.WriteAllBytes($"{directory}/{texture.Name.Content}.png", texture.TextureData.TextureBlob);
+            using FileStream fs = new($"{directory}/{texture.Name.Content}.png", FileMode.Create);
+            fs.Write(texture.TextureData.Image.ConvertToPng().ToSpan());
```
VladiStep, line 661: Can't you just:
```suggestion
            using FileStream fs = new($"{directory}/{texture.Name.Content}.png", FileMode.Create);
            texture.TextureData.Image.SavePng(fs));
```

UndertaleModLib/Models/UndertaleEmbeddedTexture.cs
```diff
@@ -311,62 +298,45 @@ public void Dispose()
...
-            get => _textureBlob;
+        public GMImage Image 
+        { 
+            get => _image;
             set
             {
-                _textureBlob = value;
-                OnPropertyChanged();
+                _image = value;
+                OnPropertyChanged(nameof(Image));
```
VladiStep, line 312: Why is it necessary to provide the name?

UndertaleModLib/Models/UndertaleTexturePageItem.cs
```diff
@@ -146,24 +147,21 @@ public void Dispose()
...
-            g.CompositingMode = System.Drawing.Drawing2D.CompositingMode.SourceCopy;
-            g.DrawImage(finalImage, SourceX, SourceY);
-            g.Dispose();
+            embImage.Composite(finalImage, SourceX, SourceY, CompositeOperator.Copy);
 
-            TexturePage.TextureData.TextureBlob = TextureWorker.GetImageBytes(embImage);
-
-            worker.Cleanup();
+            // Replace original texture with the new version, in the original texture format
+            TexturePage.TextureData.Image = GMImage.FromMagickImage(embImage).ConvertToFormat(TexturePage.TextureData.Image.Format);
```
VladiStep, line 164: I would split this line like this:
```suggestion
TexturePage.TextureData.Image = GMImage.FromMagickImage(embImage)
									   .ConvertToFormat(TexturePage.TextureData.Image.Format);
```

VladiStep: image

UndertaleModTool/Editors/UndertaleEmbeddedTextureEditor.xaml.cs
```diff
@@ -154,29 +174,41 @@ private void Import_Click(object sender, RoutedEventArgs e)
             {
                 try
                 {
-                    Bitmap bmp;
-                    using (var ms = new MemoryStream(TextureWorker.ReadTextureBlob(dlg.FileName)))
+                    GMImage image;
+                    if (System.IO.Path.GetExtension(dlg.FileName).Equals(".png", StringComparison.InvariantCultureIgnoreCase))
```
VladiStep, line 178: Why it needs to be this complicated, can't you just:
```suggestion
                    if (System.IO.Path.GetExtension(dlg.FileName) == ".png")
```

UndertaleModTool/Editors/UndertaleEmbeddedTextureEditor.xaml.cs
```diff
@@ -154,29 +174,41 @@ private void Import_Click(object sender, RoutedEventArgs e)
...
                     {
-                        bmp = new Bitmap(ms);
+                        // Import PNG data verbatim, without attempting to modify it
+                        image = GMImage.FromPng(File.ReadAllBytes(dlg.FileName), true).ConvertToFormat(target.TextureData.Image.Format);
```
VladiStep, line 181: I would split this line like this:
```suggestion
image = GMImage.FromPng(File.ReadAllBytes(dlg.FileName), true)
               .ConvertToFormat(target.TextureData.Image.Format);
```

image

UndertaleModTool/Editors/UndertaleEmbeddedTextureEditor.xaml.cs
```diff
@@ -154,29 +174,41 @@ private void Import_Click(object sender, RoutedEventArgs e)
...
+                    else
+                    {
+                        // Import any file type
+                        using var magickImage = new MagickImage(dlg.FileName);
+                        magickImage.Format = MagickFormat.Rgba;
+                        magickImage.Alpha(AlphaOption.Set);
+                        magickImage.SetCompression(CompressionMethod.NoCompression);
+
+                        // Import image
+                        image = GMImage.FromMagickImage(magickImage).ConvertToFormat(target.TextureData.Image.Format);
```
VladiStep, line 192:
```suggestion
                        image = GMImage.FromMagickImage(magickImage)
									   .ConvertToFormat(target.TextureData.Image.Format);
```

image

UndertaleModTool/Editors/UndertaleEmbeddedTextureEditor.xaml.cs
```diff
@@ -198,7 +230,8 @@ private void Export_Click(object sender, RoutedEventArgs e)
             {
                 try
                 {
-                    File.WriteAllBytes(dlg.FileName, target.TextureData.TextureBlob);
+                    using FileStream fs = new(dlg.FileName, FileMode.Create);
+                    fs.Write(target.TextureData.Image.ConvertToPng().ToSpan());
```
VladiStep, line 234: Can't you just:
```suggestion
                    target.TextureData.Image.SavePng(fs);
```

UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -254,6 +258,45 @@ public MainWindow()
...
+                }
+
+                // If we found our source, return it
+                if (foundSource is not null)
+                {
+                    return foundSource;
+                }
+
+                // If no source was found, then create a new one
+                BitmapSource bitmap = BitmapSource.Create(image.Width, image.Height, 96, 96, PixelFormats.Bgra32, null, image.ConvertToRawBgra().ToSpan().ToArray(), image.Width * 4);
```
VladiStep, line 293: I would:
```suggestion
                byte[] imgDataArray = image.ConvertToRawBgra().ToSpan().ToArray();
                BitmapSource bitmap = BitmapSource.Create(image.Width, image.Height, 96, 96, PixelFormats.Bgra32, null, imgDataArray, image.Width * 4);
```

UndertaleModTool/UndertaleModTool.csproj
```diff
@@ -99,6 +99,7 @@
         <PackageReference Include="System.Diagnostics.StackTrace">
             <Version>4.3.0</Version>
         </PackageReference>
+        <PackageReference Include="System.Formats.Asn1" Version="8.0.1" />
```
VladiStep, line 102: Is it necessary?
Same for two lines below (115-116).

VladiStep (dismissed):

UndertaleModCli/Program.cs
colinator27: Good catch, I've now changed both occurrences of this.

UndertaleModLib/Models/UndertaleEmbeddedTexture.cs
colinator27: I forgot the `CallerMemberName` attribute was there, so I've now removed this.

UndertaleModTool/Editors/UndertaleEmbeddedTextureEditor.xaml.cs
colinator27: It needs to be case-insensitive (e.g., if the filename ends in `.PNG` instead of `.png`).

UndertaleModTool/UndertaleModTool.csproj
colinator27: I think I accidentally added those when testing the .NET upgrade, so I'm now removing them.

UndertaleModTool/Editors/UndertaleEmbeddedTextureEditor.xaml.cs
VladiStep: Is there a difference between them?

UndertaleModLib/Util/TextureWorker.cs
```diff
@@ -53,19 +73,19 @@ public IMagickImage<byte> GetTextureFor(UndertaleTexturePageItem texPageItem, st
...
                 catch (OutOfMemoryException)
                 {
-                    throw new OutOfMemoryException(imageName + "'s texture is abnormal. 'Source Position/Size' boxes 3 & 4 on texture page may be bigger than the sprite itself or it's set to '0'.");
+                    throw new OutOfMemoryException($"{imageName}'s entry is abnormal. 'Source Position/Size' boxes on the texture page item may be set incorrectly.");
                 }
             }
 
-            // Resize the image, if necessary.
-            if ((texPageItem.SourceWidth != texPageItem.TargetWidth) || (texPageItem.SourceHeight != texPageItem.TargetHeight))
+            // Resize the image, if necessar
```
VladiStep, line 80: // Resize the image, if necessar**y**

UndertaleModLib/Util/TextureWorker.cs
```diff
@@ -1,217 +1,250 @@
...
-                                 (int)(bm.VerticalResolution + 0.5f));
+                // We already have the correct dimensions, so just make a clone of the image.
+                // If we don't clone here, we have potential ownership/disposal issues.
+                return image.Clone();
             }
 
-            return bm;
+            // Clone and resize using bilinear interpolation
+            IMagickImage<byte> newImage = image.Clone();
+            newImage.InterpolativeResize(width, height, PixelInterpolateMethod.Bilinear);
```
Miepee, line 133: feels like its nicer to rewrite this as 
```cs
IMagickImage<byte> newImage = image.Clone();
if (dimensionsCheck)
  return newImage;
 
newImage.Interpolate();
return newImage;
```

UndertaleModLib/Util/TextureWorker.cs
```diff
@@ -1,217 +1,250 @@
...
         }
 
-        // This should perform a high quality resize.
-        // Grabbed from https://stackoverflow.com/questions/1922040/how-to-resize-an-image-c-sharp
-        public static Bitmap ResizeImage(Image image, int width, int height)
+        /// <summary>
+        /// Reads collision mask data from the given file path, and required width/height.
+        /// </summary>
+        /// <exception cref="Exception">If the loaded image dimensions do not match the required width/height</exception>
+        public static byte[] ReadMaskData(string filePath, int requiredWidth, int requiredHeight)
```
Miepee, line 141: missing parameter doc

UndertaleModLib/Util/TextureWorker.cs
```diff
@@ -1,217 +1,250 @@
...
             {
-                graphics.CompositingMode = CompositingMode.SourceCopy;
-                graphics.CompositingQuality = CompositingQuality.HighQuality;
-                graphics.InterpolationMode = InterpolationMode.HighQualityBicubic;
-                graphics.SmoothingMode = SmoothingMode.HighQuality;
-                graphics.PixelOffsetMode = PixelOffsetMode.HighQuality;
-
-                using (var wrapMode = new ImageAttributes())
+                // Verify width/height match required width/height
+                if (image.Width != requiredWidth || image.Height != requiredHeight)
```
Miepee, line 147: so, why do we need the custom width/height as parameters if we always end up using the image's one?

UndertaleModLib/Util/TextureWorker.cs
```diff
@@ -1,217 +1,250 @@
...
+                // Get image pixels, and allocate enough capacity for mask
+                IPixelCollection<byte> pixels = image.GetPixels();
+                bytes = new((requiredWidth + 7) / 8 * requiredHeight);
 
-        public static byte[] ReadMaskData(string filePath)
-        {
-            Bitmap image = ReadImageFromFile(filePath);
-            List<byte> bytes = new List<byte>();
+                // Get white color, used to represent bits that are set
+                IMagickColor<byte> white = MagickColor.FromRgba(255, 255, 255, 255);
```
Miepee, line 157: surely imagemagick has a "white" constant?

UndertaleModLib/Util/TextureWorker.cs
```diff
@@ -1,217 +1,250 @@
...
-        {
-            var blackImage = new Bitmap(width, height);
-            for (int x = 0; x < width; x++)
-                for (int y = 0; y < height; y++)
-                    blackImage.SetPixel(x, y, Color.Black);
-            SaveImageToFile(FullPath, blackImage);
-        }
+            // Get black/white colors to use for drawing
+            ReadOnlySpan<byte> black = MagickColor.FromRgba(0, 0, 0, 255).ToByteArray().AsSpan();
+            ReadOnlySpan<byte> white = MagickColor.FromRgba(255, 255, 255, 255).ToByteArray().AsSpan();
```
Miepee, line 196: see above

UndertaleModLib/Util/TextureWorker.cs
```diff
@@ -1,217 +1,250 @@
...
-            image.Dispose();
             return bytes.ToArray();
         }
 
-        public static byte[] ReadTextureBlob(string filePath)
+        /// <summary>
+        /// Generates and returns a black-and-white image representing the given sprite's specified collision mask,
+        /// and with the given width/height.
+        /// </summary>
+        public static IMagickImage<byte> GetCollisionMaskImage(UndertaleSprite sprite, UndertaleSprite.MaskEntry mask, int maskWidth, int maskHeight)
```
Miepee, line 188: missing parameters

UndertaleModLib/Util/TextureWorker.cs
```diff
@@ -1,217 +1,250 @@
...
         {
-            Image.FromFile(filePath).Dispose(); // Make sure the file is valid image.
-            return File.ReadAllBytes(filePath);
-        }
+            // Create image to draw on
+            MagickImage bitmap = new(MagickColor.FromRgba(0, 0, 0, 255), maskWidth, maskHeight);
```
Miepee, line 191: probably want to rename this variable

UndertaleModLib/Util/TextureWorker.cs
```diff
@@ -1,217 +1,250 @@
...
             }
 
             return bitmap;
         }
 
-        public static void ExportCollisionMaskPNG(UndertaleSprite sprite, UndertaleSprite.MaskEntry mask, string fullPath)
+        /// <summary>
+        /// Exports a collision mask entry from the given sprite, as a PNG file, at the specified path, and with the given width/height.
+        /// </summary>
+        public static void ExportCollisionMaskPNG(UndertaleSprite sprite, UndertaleSprite.MaskEntry mask, string path, int maskWidth, int maskHeight)
```
Miepee, line 217: missing params

UndertaleModLib/Util/TextureWorker.cs
```diff
@@ -1,217 +1,250 @@
...
+        private Dictionary<UndertaleEmbeddedTexture, MagickImage> embeddedDictionary = new();
+        private readonly object embeddedDictionaryLock = new();
+
+        /// <summary>
+        /// Retrieves an image representing the supplied texture page.
+        /// </summary>
+        /// <remarks>
+        /// The returned image will be cached for this <see cref="TextureWorker"/> instance.
+        /// </remarks>
+        public MagickImage GetEmbeddedTexture(UndertaleEmbeddedTexture embeddedTexture)
```
Miepee, line 23: missing return documentation

UndertaleModLib/Util/TextureWorker.cs
```diff
@@ -1,217 +1,250 @@
...
+
+                return newImage;
             }
         }
 
-        public void ExportAsPNG(UndertaleTexturePageItem texPageItem, string FullPath, string imageName = null, bool includePadding = false)
+        /// <summary>
+        /// Exports the given texture page item to disk, as a PNG, to the supplied path. (With or without padding.)
+        /// </summary>
+        public void ExportAsPNG(UndertaleTexturePageItem texPageItem, string fullPath, string imageName = null, bool includePadding = false)
```
Miepee, line 44: missing return documentation

UndertaleModLib/Util/TextureWorker.cs
```diff
@@ -1,217 +1,250 @@
...
-            SaveImageToFile(FullPath, GetTextureFor(texPageItem, imageName != null ? imageName : Path.GetFileNameWithoutExtension(FullPath), includePadding));
+            using var image = GetTextureFor(texPageItem, imageName ?? Path.GetFileNameWithoutExtension(fullPath), includePadding);
+            SaveImageToFile(image, fullPath);
         }
 
-        public Bitmap GetTextureFor(UndertaleTexturePageItem texPageItem, string imageName, bool includePadding = false)
+        /// <summary>
+        /// Creates an image representing the sole texture page item supplied, with or without padding.
+        /// </summary>
+        public IMagickImage<byte> GetTextureFor(UndertaleTexturePageItem texPageItem, string imageName, bool includePadding = false)
```
Miepee, line 53: missing return documentation

UndertaleModLib/Util/TextureWorker.cs
```diff
@@ -1,217 +1,250 @@
...
         }
 
-        public static Bitmap ReadImageFromFile(string filePath)
+        /// <summary>
+        /// Reads an image from the given file path (of arbitrary format, as supported by <see cref="MagickImage(string, IMagickReadSettings{byte})"/>).
+        /// </summary>
+        /// <remarks>
+        /// Image color format will always be converted to BGRA, with no compression.
+        /// </remarks>
+        public static MagickImage ReadBGRAImageFromFile(string filePath)
```
Miepee, line 106: missing return documentation

UndertaleModLib/Util/TextureWorker.cs
```diff
@@ -1,217 +1,250 @@
...
+            image.SetCompression(CompressionMethod.NoCompression);
+            return image;
         }
 
-        // Grabbed from https://stackoverflow.com/questions/3801275/how-to-convert-image-to-byte-array/16576471#16576471
-        public static Bitmap GetImageFromByteArray(byte[] byteArray)
+        /// <summary>
+        /// Performs a resize of the given image, if required, using bilinear interpolation. Always returns a new image.
+        /// </summary>
+        public static IMagickImage<byte> ResizeImage(IMagickImage<byte> image, int width, int height)
```
Miepee, line 122: missing return documentation

UndertaleModLib/Util/TextureWorker.cs
Miepee, line 141: missing return doc

UndertaleModLib/Util/TextureWorker.cs
Miepee, line 188: missing return doc

UndertaleModLib/Util/TextureWorker.cs
Miepee, line 217: missing return

UndertaleModLib/Util/TextureWorker.cs
```diff
@@ -1,217 +1,250 @@
...
-            SaveImageToFile(fullPath, GetCollisionMaskImage(sprite, mask));
+            using var image = GetCollisionMaskImage(sprite, mask, maskWidth, maskHeight);
+            SaveImageToFile(image, path);
         }
 
-        public static byte[] GetImageBytes(Image image, bool disposeImage = true)
+        /// <summary>
+        /// Saves the provided image as a PNG file, at the specified path.
+        /// </summary>
+        public static void SaveImageToFile(IMagickImage<byte> image, string path)
```
Miepee, line 226: param + return doc

UndertaleModLib/Util/TextureWorker.cs
```diff
@@ -1,217 +1,250 @@
...
-                    image.Dispose();
-                return result;
-            }
+            using var stream = new FileStream(path, FileMode.Create);
+            image.Write(stream, MagickFormat.Png32);
         }
 
-        public static void SaveImageToFile(string FullPath, Image image, Boolean disposeImage = true)
+        /// <summary>
+        /// This should be called when the TextureWorker is no longer going to be used; this frees image data.
```
Miepee, line 233: Surely we can just use inheritdoc for this, no?
If not:

```suggestion
        /// This should be called when the TextureWorker is no longer going to be used, as it frees cached image data.
```

UndertaleModTool/Converters/UndertaleCachedImageLoader.cs
```diff
@@ -160,30 +161,40 @@ public static void Reset()
...
+            var data = spriteBitmap.LockBits(new Rectangle(x, y, rect.Width, rect.Height), ImageLockMode.WriteOnly, System.Drawing.Imaging.PixelFormat.Format32bppArgb);
+            bitmapSource.CopyPixels(new Int32Rect(rect.X, rect.Y, rect.Width, rect.Height), data.Scan0, data.Height * data.Stride, data.Stride);
+            spriteBitmap.UnlockBits(data);
+
+            return spriteBitmap;
         }
         private ImageSource CreateSpriteSource(in Rectangle rect, in UndertaleTexturePageItem texture, int diffW = 0, int diffH = 0, bool isTile = false)
         {
-            Bitmap spriteBMP = CreateSpriteBitmap(rect, in texture, diffW, diffH, isTile);
+            Bitmap spriteBitmap = CreateSpriteBitmap(rect, in texture, diffW, diffH, isTile);
```
Miepee, line 192: why not use `using` instad?

UndertaleModTool/Editors/UndertaleTexturePageItemEditor.xaml
```diff
@@ -116,9 +116,9 @@
                     </DrawingBrush>
                 </Grid.Background>
 
-                <Image Grid.Row="0" Grid.Column="1" Source="{Binding TexturePage.TextureData.TextureBlob, Mode=OneWay}" Opacity="0.2"/>
+                <Image Grid.Row="0" Grid.Column="1" Name="TextureImageView1" Opacity="0.2"/>
 
-                <Image Grid.Row="0" Grid.Column="1" Source="{Binding TexturePage.TextureData.TextureBlob, Mode=OneWay}">
+                <Image Grid.Row="0" Grid.Column="1" Name="TextureImageView2">
```
Miepee, line 121: would be nice if these would get some better names

UndertaleModTool/Editors/UndertaleEmbeddedTextureEditor.xaml
```diff
@@ -89,7 +89,7 @@
                         </Border.Background>
 
                         <Grid Cursor="Hand" MouseDown="Grid_MouseDown" MouseMove="Grid_MouseMove" MouseLeave="Grid_MouseLeave">
-                            <Image Source="{Binding TextureData.TextureBlob, Mode=OneWay}"/>
+                            <Image Name="TextureImageView"/>
```
Miepee, line 92: same here

UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -101,6 +102,9 @@ public object Selected
 
         public List<Tab> ClosedTabsHistory { get; } = new();
 
+        private List<(GMImage, WeakReference<BitmapSource>)> _bitmapSourceLookup { get; } = new();
+        private object _bitmapSourceLookupLock = new();
```
Miepee, line 106: see lock

UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -101,6 +102,9 @@ public object Selected
 
         public List<Tab> ClosedTabsHistory { get; } = new();
 
+        private List<(GMImage, WeakReference<BitmapSource>)> _bitmapSourceLookup { get; } = new();
```
Miepee, line 105: is there a reason why this needs to be a list instead of a dictionary?

UndertaleModLib/Models/UndertaleTexturePageItem.cs
```diff
@@ -146,24 +147,22 @@ public void Dispose()
     /// </summary>
     /// <param name="replaceImage">The new image that shall be applied to this texture page item.</param>
     /// <param name="disposeImage">Whether to dispose <paramref name="replaceImage"/> afterwards.</param>
-    public void ReplaceTexture(Image replaceImage, bool disposeImage = true)
+    public void ReplaceTexture(MagickImage replaceImage, bool disposeImage = true)
     {
-        Image finalImage = TextureWorker.ResizeImage(replaceImage, SourceWidth, SourceHeight);
+        // Resize image to bounds on texture page
```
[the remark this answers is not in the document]
VladiStep: "texture page", "TexturePage" (line 155).
I also didn't notice that.

UndertaleModLib/UndertaleChunks.cs
```diff
@@ -1503,7 +1503,7 @@ private void CheckFor2022_3And5(UndertaleReader reader)
                     reader.AbsPosition = reader.ReadUInt32() + 12; // Go to texture, at an offset
                     reader.AbsPosition = reader.ReadUInt32(); // Go to texture data
                     byte[] header = reader.ReadBytes(4);
-                    if (!header.SequenceEqual(UndertaleEmbeddedTexture.TexData.QOIAndBZip2Header))
+                    if (header is not [50, 122, 111, 113]) // '2zoq' magic
```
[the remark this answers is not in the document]
VladiStep: I also had this idea, but I thought there's no need to create a property that's used only in one place.

UndertaleModTool/Converters/UndertaleCachedImageLoader.cs
VladiStep: But then wouldn't this line be too long?

UndertaleModLib/Util/GMImage.cs
```diff
@@ -0,0 +1,817 @@
...
+
+    /// <summary>
+    /// Width of this image, in pixels.
+    /// </summary>
+    public int Width { get; init; }
+
+    /// <summary>
+    /// Height of this image, in pixels.
+    /// </summary>
+    public int Height { get; init; }
```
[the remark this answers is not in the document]
VladiStep: Probably because "... a new **immutable** `GMImage` class ...".
Also, even if these were settable, how would it work?

UndertaleModLib/Util/GMImage.cs
```diff
@@ -0,0 +1,817 @@
...
+
+    /// <summary>
+    /// Initializes an image with raw format, of the desired width and height.
+    /// </summary>
+    /// <remarks>
+    /// Creates a completely blank image (black, 0 alpha).
+    /// </remarks>
+    public GMImage(int width, int height)
+    {
+        if (width < 0 || width > MaxImageDimension)
```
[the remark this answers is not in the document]
VladiStep: Could you write how it should look like?

UndertaleModLib/Util/GMImage.cs
```diff
@@ -0,0 +1,817 @@
...
+        Width = width;
+        Height = height;
+        _data = data;
+    }
+
+    // Searches for the BZ2 footer magic, when around the end of a BZ2 stream, and returns the exact end position of the stream
+    private static long FindEndOfBZ2Search(IBinaryReader reader, long endDataPosition)
+    {
+        // Magic bit sequence found near the end of a BZip2 stream (square root of pi)
+        ReadOnlySpan<byte> footerMagic = stackalloc byte[] { 0x17, 0x72, 0x45, 0x38, 0x50, 0x90 };
```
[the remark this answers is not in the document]
VladiStep: Seems like a good point.

UndertaleModLib/Util/GMImage.cs
```diff
@@ -0,0 +1,817 @@
...
+            }
+
+            // Calculate length, read entire image to byte array
+            long length = reader.Position - startAddress;
+            reader.Position = startAddress;
+            return FromPng(reader.ReadBytes((int)length));
+        }
+
+        // QOI + BZip2
+        if (header is [50, 122, 111, 113, ..])
```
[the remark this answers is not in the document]
VladiStep: Interesting.
I know about the new C# 12 collection expression, but I also see that syntax the first time.
Apparently, it's the same as `if (header.StartsWith([50, 122, 111, 113])`:
```csharp
if (a is [1, 2, ..])
{
    Console.WriteLine("a starts with 1, 2");
}
```
... translates into ...
```csharp
if (array2 != null && array2.Length >= 2 && array2[0] == 1 && array2[1] == 2)
{
    Console.WriteLine("a starts with 1,2");
}
```

UndertaleModLib/Models/UndertaleEmbeddedTexture.cs
```diff
@@ -311,62 +298,45 @@ public void Dispose()
...
+        public bool FormatQOI => _image.Format is GMImage.ImageFormat.Qoi or GMImage.ImageFormat.Bz2Qoi;
 
         /// <summary>
         /// Whether this texture uses the BZ2 format. (Always used in combination with QOI.)
         /// </summary>
-        public bool FormatBZ2 { get; set; } = false;
+        public bool FormatBZ2 => _image.Format is GMImage.ImageFormat.Bz2Qoi;
+
+        /// <summary>
+        /// If located within a data file, this is the maximum end position of the image data (or start of the next texture blob).
```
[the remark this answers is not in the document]
colinator27: So the "maximum end position" in this context _is_ actually an upper bound on the end position of the image data. Inside a data file, it's either set to the start address of the next texture blob, or the end of the TXTR chunk if none exists. When loading textures, this is then narrowed down to the actual end address (but this field isn't updated).

I think I'm going to actually make this a private field, and just make a setter method for this, since this isn't meant to be relied upon for anything other than deserialization.

UndertaleModLib/Models/UndertaleEmbeddedTexture.cs
```diff
@@ -379,152 +349,42 @@ protected void OnPropertyChanged([CallerMemberName] string name = null)
...
-                uint type = reader.ReadUInt32();
-                reader.Position += len + 4;
-                if (type == 0x444e4549) // 0x444e4549 -> "IEND"
-                    break;
-            }
-
-            long length = reader.Position - startAddress;
-            reader.Position = startAddress;
-            TextureBlob = reader.ReadBytes((int)length);
+            Image = GMImage.FromBinaryReader(reader, maxEndOfStreamPosition, gm2022_5);
```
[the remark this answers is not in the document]
colinator27: Not always... only if it's a BZ2+QOI type image, although it could also break on other image types in the future. I guess since it's public I'll add an exception for that, and probably document that behavior.

UndertaleModLib/Models/UndertaleTexturePageItem.cs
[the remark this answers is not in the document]
colinator27: Oh yeah, these are a bunch of ancient comments that I haven't changed, but I'll change them now.

UndertaleModLib/UndertaleModLib.csproj
```diff
@@ -28,10 +29,10 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
+    <PackageReference Include="Magick.NET-Q8-AnyCPU" Version="13.10.0" />
```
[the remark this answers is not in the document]
colinator27: Yeah, all the channels in GameMaker's textures (right now) are 8-bit. Also I think it should work cross-platform, although we may need to touch a few extra settings to copy the native libraries correctly.

UndertaleModLib/Util/GMImage.cs
```diff
@@ -0,0 +1,817 @@
...
+    private readonly byte[] _data = null;
+
+    // If this is a Bz2Qoi image in GameMaker 2022.5 and above, this is the size of the BZip2 data when entirely uncompressed
+    private int _bz2UncompressedSize { get; init; } = -1;
+
+    /// <summary>
+    /// Initializes an image with raw format, of the desired width and height.
+    /// </summary>
+    /// <remarks>
+    /// Creates a completely blank image (black, 0 alpha).
```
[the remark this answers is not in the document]
colinator27: Fully transparent, I'll update it to clarify.

UndertaleModLib/Util/GMImage.cs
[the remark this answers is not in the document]
colinator27: Yeah, I'll just make a private static (readonly) field for this I guess.

UndertaleModLib/Util/GMImage.cs
```diff
@@ -0,0 +1,817 @@
...
+        throw new IOException("Failed to find nonzero data");
+    }
+
+    /// <summary>
+    /// Creates a <see cref="GMImage"/> from the image contents stored at the current position of the provided <see cref="IBinaryReader"/>.
+    /// </summary>
+    /// <param name="reader">Binary reader to read the image data from.</param>
+    /// <param name="maxEndOfStreamPosition">
+    /// Location where the image stream ends, from within the <see cref="IBinaryReader"/>.
+    /// There should only be 0 (null) bytes between this position and the end of the image data.
```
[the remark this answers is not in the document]
colinator27: Left a comment above sort of explaining this, but there's basically padding between the end of the image and the next image blob (or the end of the chunk). Maybe I'll reword it to be "padding" or something...

UndertaleModLib/Util/GMImage.cs
```diff
@@ -0,0 +1,817 @@
...
+
+    /// <summary>
+    /// Creates a <see cref="GMImage"/> of BZ2 + QOI format, wrapping around the provided byte array containing BZ2-compressed data (no header).
+    /// </summary>
+    /// <param name="compressedData">Compressed BZ2 data, excluding the header.</param>
+    /// <param name="width">Width of the image, as provided in BZ2 + QOI header.</param>
+    /// <param name="height">Height of the image, as provideed in BZ2 + QOI header.</param>
+    /// <param name="uncompressedLength">Length of BZ2 data when fully uncompressed.</param>
+    /// <exception cref="InvalidDataException">Invalid BZ2 + QOI data, or image is too large</exception>
+    public static GMImage FromBz2Qoi(byte[] compressedData, int width, int height, int uncompressedLength)
```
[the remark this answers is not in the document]
VladiStep: But if the width and height are already known, why we should make extra work?
Though, it _may_ be useful to have an overload like this.

UndertaleModLib/Util/GMImage.cs
```diff
@@ -0,0 +1,817 @@
...
+        ArgumentNullException.ThrowIfNull(reader);
+
+        // Determine type of image by reading the first 8 bytes
+        long startAddress = reader.Position;
+        byte[] header = reader.ReadBytes(8);
+
+        // PNG
+        if (header is [137, 80, 78, 71, 13, 10, 26, 10])
+        {
+            // There is no length for the PNG anywhere as far as I can see
```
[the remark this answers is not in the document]
colinator27: This is actually an ancient UTMT comment I pasted in here, but yeah I can change it to clarify what it's doing.

UndertaleModLib/UndertaleChunks.cs
```diff
@@ -1564,8 +1560,7 @@ internal override void SerializeChunk(UndertaleWriter writer)
                 if (anythingUsesQoi)
                 {
                     // Calculate maximum size of QOI converter buffer
-                    maxSize = List.Select(x => x.TextureData.Width * x.TextureData.Height).Max()
-                              * QoiConverter.MaxChunkSize + QoiConverter.HeaderSize + (writer.undertaleData.IsVersionAtLeast(2022, 3) ? 0 : 4);
+                    int maxSize = (List.Select(x => x.TextureData.Width * x.TextureData.Height).Max() * QoiConverter.MaxChunkSize) + QoiConverter.HeaderSize;
```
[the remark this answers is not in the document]
colinator27: I'll try it out... Not sure if that's supported in this C# version or not.

UndertaleModLib/Util/GMImage.cs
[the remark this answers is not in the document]
colinator27: Not entirely sure if you can do pattern matching on non-constant expressions, but maybe it'll work here... I'll give it a shot.

UndertaleModLib/Util/GMImage.cs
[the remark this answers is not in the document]
colinator27: Yeah, that's only in the header, or after uncompressing the data (which we can't do for performance reasons).

UndertaleModLib/Util/GMImage.cs
```diff
@@ -0,0 +1,817 @@
...
+    /// <summary>
+    /// Returns the same or a new <see cref="GMImage"/>; the result of converting this image to the specified <see cref="ImageFormat"/>.
+    /// </summary>
+    /// <param name="format">Format to convert to</param>
+    /// <param name="sharedStream">Reusable shared <see cref="MemoryStream"/> to be used when compressing with BZ2, as required.</param>
+    public GMImage ConvertToFormat(ImageFormat format, MemoryStream sharedStream = null)
+    {
+        return format switch
+        {
+            ImageFormat.RawBgra => ConvertToRawBgra(),
```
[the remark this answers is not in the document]
colinator27: That's actually how half of the conversion operations are already implemented - they convert to raw BGRA, then to the target format.

UndertaleModLib/Util/QoiConverter.cs
```diff
@@ -63,19 +60,19 @@ public static Bitmap GetImageFromStream(Stream s)
...
-        public static Bitmap GetImageFromSpan(ReadOnlySpan<byte> bytes) => GetImageFromSpan(bytes, out _);
+        public static GMImage GetImageFromSpan(ReadOnlySpan<byte> bytes) => GetImageFromSpan(bytes, out _);
 
         /// <summary><inheritdoc cref="GetImageFromSpan(System.ReadOnlySpan{byte})"/></summary>
         /// <param name="bytes"><inheritdoc cref="GetImageFromSpan(System.ReadOnlySpan{byte})"/></param>
         /// <param name="length">The total amount of data read from the <see cref="Span{TKey}"/>.</param>
         /// <returns><inheritdoc cref="GetImageFromSpan(System.ReadOnlySpan{byte})"/></returns>
         /// <exception cref="Exception"><inheritdoc cref="GetImageFromSpan(System.ReadOnlySpan{byte})"/></exception>
-        public unsafe static Bitmap GetImageFromSpan(ReadOnlySpan<byte> bytes, out int length)
+        public unsafe static GMImage GetImageFromSpan(ReadOnlySpan<byte> bytes, out int length)
```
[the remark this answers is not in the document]
colinator27: I think it might not be that bad, actually. At this point it's just looping over data in an array, so I guess the only performance gain we get from this is no bounds checks. Not sure how optimized the JIT is at getting rid of those, though.

UndertaleModLib/Util/TextureWorker.cs
```diff
@@ -1,217 +1,250 @@
...
+            // Create an image cropped from the item's part of the texture page
+            IMagickImage<byte> resultImage = null;
             lock (embeddedImage)
             {
                 try
                 {
-                    resultImage = embeddedImage.Clone(new Rectangle(texPageItem.SourceX, texPageItem.SourceY, texPageItem.SourceWidth, texPageItem.SourceHeight), PixelFormat.DontCare);
+                    resultImage = embeddedImage.Clone(texPageItem.SourceX, texPageItem.SourceY, texPageItem.SourceWidth, texPageItem.SourceHeight);
                 }
                 catch (OutOfMemoryException)
```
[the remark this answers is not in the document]
colinator27: Honestly I have no idea at this point. I'm personally fine with removing this try..catch altogether, especially since the underlying code is different now. I'm assuming this was a hack added to account for some kind of overflow. What do you think?

UndertaleModLib/Util/TextureWorker.cs
```diff
@@ -1,217 +1,250 @@
...
+            image.Alpha(AlphaOption.Set);
+            image.Format = MagickFormat.Bgra;
+            image.SetCompression(CompressionMethod.NoCompression);
+            return image;
         }
 
-        // Grabbed from https://stackoverflow.com/questions/3801275/how-to-convert-image-to-byte-array/16576471#16576471
-        public static Bitmap GetImageFromByteArray(byte[] byteArray)
+        /// <summary>
+        /// Performs a resize of the given image, if required, using bilinear interpolation. Always returns a new image.
```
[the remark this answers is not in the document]
colinator27: Not sure, but for now it's probably fine to leave this as-is. We can add optional parameters later to configure the interpolation, if we want.

UndertaleModLib/Util/TextureWorker.cs
colinator27: Good catch, I just dug around the docs for a bit and finally found the color constants.

UndertaleModTool/Converters/UndertaleCachedImageLoader.cs
colinator27: I didn't really touch this code other than renaming the variable, but I think this is manually disposed a little bit before the function returns. We could switch to `using`, but it would need to be scoped in the same way.

UndertaleModTool/MainWindow.xaml.cs
colinator27: So this was originally a dictionary, but I changed it to a list because every lookup does a full sweep on the weak references anyway. Not sure if there's any better way to do it, though (we could possibly defer the sweep operation).

colinator27 (commented):
Left some responses, but I'll get to updating the code soon to resolve most of the comments.

UndertaleModLib/Util/GMImage.cs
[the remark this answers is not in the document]
Miepee: oh yeah whoops.

UndertaleModLib/Util/GMImage.cs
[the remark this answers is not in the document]
Miepee: having that more explicitly would be nice, but could just always say that that's for the future:tm:

UndertaleModLib/Util/TextureWorker.cs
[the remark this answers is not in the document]
Miepee: considering you have a few clones further down below, id just remove it yeah

UndertaleModLib/Util/TextureWorker.cs
Miepee: bump

UndertaleModTool/MainWindow.xaml.cs
Miepee: wdym with "every lookup does a full sweep on the references"?
The way I see it is that currently, it *always* iterates over the full list when you want to get just one item, cause it always invalidated everything.

With a dictionary, you would only check the image asked for, and only invalidate that.

UndertaleModLib/Models/UndertaleEmbeddedTexture.cs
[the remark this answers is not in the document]
Miepee: I'm still confused by the name. If this is the end position of the image, why not just call it "endOfImagePosition"?

UndertaleModTool/Editors/UndertaleTexturePageItemEditor.xaml.cs
```diff
@@ -21,30 +20,109 @@ public partial class UndertaleTexturePageItemEditor : DataUserControl
...
+            _textureDataContext.PropertyChanged += ReloadTextureImage;
+        }
+
+        private void ReloadTexturePage(object sender, PropertyChangedEventArgs e)
+        {
+            UndertaleTexturePageItem item = (DataContext as UndertaleTexturePageItem);
+            if (item is null)
+                return;
+
+            if (e.PropertyName == nameof(UndertaleTexturePageItem.TexturePage))
```
Miepee, line 81: early return

UndertaleModTool/Editors/UndertaleTexturePageItemEditor.xaml.cs
```diff
@@ -21,30 +20,109 @@ public partial class UndertaleTexturePageItemEditor : DataUserControl
...
+        }
+
+        private void ReloadTextureImage(object sender, PropertyChangedEventArgs e)
+        {
+            UndertaleTexturePageItem item = (DataContext as UndertaleTexturePageItem);
+            if (item is null)
+                return;
+
+            // If the texture's image was updated, reload it
+            if (e.PropertyName == nameof(UndertaleEmbeddedTexture.TexData.Image))
```
Miepee, line 102: early return

UndertaleModLib/Util/GMImage.cs
```diff
@@ -0,0 +1,817 @@
...
+                    searchStartBitPosition = 0;
+                    searchStartPosition--;
+                }
+            }
+        }
+
+        throw new IOException("Failed to find BZip2 footer magic");
+    }
+
+    // Finds the end position of a BZ2 stream exactly, given the start and end bounds of the data
```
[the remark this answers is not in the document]
Miepee: this wasnt addressed

UndertaleModLib/Util/GMImage.cs
```diff
@@ -0,0 +1,817 @@
...
+
+                    // We didn't find a full match yet, so we also need to progress our search position to the next bit
+                    bitPosition++;
+                    if (bitPosition >= 8)
+                    {
+                        bitPosition = 0;
+                        searchPosition--;
+                    }
+                }
+                else
```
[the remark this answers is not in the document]
Miepee: this also wasnt addressed

UndertaleModLib/Util/GMImage.cs
```diff
@@ -0,0 +1,863 @@
...
+                {
+                    // BZip2 footer started partway through a byte, and so it will end partway through the last byte.
+                    // By the BZip2 specification, the unused bits of the last byte are essentially padding.
+                    endOfBZ2StreamPosition++;
+                }
+
+                // Return position relative to the start of the data we read
+                return (endDataPosition - data.Length) + endOfBZ2StreamPosition;
+            }
+            else
```
Miepee, line 198: no need for an else, the if returns

UndertaleModLib/Util/GMImage.cs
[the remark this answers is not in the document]
Miepee: nvm i think i t was, as i cant see this piece of code anymore

UndertaleModLib/Util/GMImage.cs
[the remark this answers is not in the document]
Miepee: it was, i was shown an outdated diff >:(

UndertaleModTool/MainWindow.xaml.cs
colinator27: Yeah, so in this case it's weird, because we also want to clear the "dictionary" of the weak references that are no longer valid (because the GC freed their memory). If you don't do this, it could gradually leak memory, especially if the images are getting replaced by new ones. If we move this clearing process somewhere else, we don't need to do a linear lookup here, but then we have to find another good place to do that "sweep" operation anyway. As it stands, though, I don't think we ever have more than a small handful of textures loaded at once.

UndertaleModLib/Models/UndertaleEmbeddedTexture.cs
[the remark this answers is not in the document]
colinator27: I'm going to update this with a newer comment in just a moment, hopefully it'll be more clear - but this is the upper bound on the image end position, not the actual image end position.

UndertaleModTool/MainWindow.xaml.cs
Miepee: I see, then probably no harm in keeping it as a list and keeping the linearity.

Miepee (approved):
LGTM, I'll do a quick test on linux CLI soon:tm: to check whether it works and hten merge.

Miepee:
CLI was able to dump all embedded textures, so i'll call this a success!
Gonna be merging this now, fixing scripts is next on the table then :)

PR #1874, opened 2024-08-17
Fix Branch being overwritten less precisely

### LINKED ISSUE #1873: UndertaleTextureGroupInfo in chunk TGIN
### Describe the bug

obraz
this is a bug while loading data with UndertaleModTool v0.6.0.0. The bug is doesn't appear in UndertaleModTool v0.5.1.0. Why is that?

### Reproducing steps

this is a bug while loading data with UndertaleModTool v0.6.0.0. The bug is doesn't appear in UndertaleModTool v0.5.1.0. Why is that?

### Setup Details

1. UndertaleModTool v.0.6.0.0
2. OS Windows.
3. GM 2023.1
4. game: SuperAutoCoots, attached.
[bocary (itch) - superautocoots.zip](https://github.com/user-attachments/files/16643293/bocary.itch.-.superautocoots.zip)

Jacky720:
## Description
The method TestForCommonGMSVersions estimates the version of a WAD based on which chunks are present. This is invoked twice: once at the start of deserialization, and once while parsing the GEN8 (General Info) chunk. The second case is... actually, it's completely redundant, so this PR removes it. This fixes the identification of non-LTS 2023.1 in TGIN and therefore closes #1873.

### Caveats
N/A

### Notes
N/A


colinator27:
Tested on a wide range of GameMaker versions, seems to be pretty stable.

UndertaleModLib/Models/UndertaleGeneralInfo.cs
```diff
@@ -481,6 +481,14 @@ public void Unserialize(UndertaleReader reader)
...
             return;
 
+        // TestForCommonGMSVersions is already run during the object counting phase, so the previous general info is always more accurate.
+        var prevGenInfo = reader.undertaleData.GeneralInfo;
+        Major = prevGenInfo.Major;
+        Minor = prevGenInfo.Minor;
+        Release = prevGenInfo.Release;
+        Build = prevGenInfo.Build;
+        Branch = prevGenInfo.Branch;
+        /*
```
colinator27, line 491: We probably want to just remove this old code, rather than commenting it out.

colinator27 (changes requested):

UndertaleModLib/Models/UndertaleGeneralInfo.cs
Miepee: agree

Miepee (approved):

PR #1877, opened 2024-08-18
Changed quitting to show a yes/no/cancel dialog instead of 2 yes/no dialogs

luizzeroxis:
## Description
It was really annoying trying to close the program and it showing two dialog boxes back to back! So I changed it so it's a more standard dialog like other programs.

### Caveats
Should not be any

### Notes
Hopefully this will save several seconds of people's lives

### DIFF
--- a/UndertaleModTool/MainWindow.xaml.cs
+++ b/UndertaleModTool/MainWindow.xaml.cs
@@ -772,58 +772,59 @@ namespace UndertaleModTool
                 e.Cancel = true;
 
                 bool save = false;
+                bool quit = false;
 
                 if (SettingsWindow.WarnOnClose)
                 {
-                    if (this.ShowQuestion("Are you sure you want to quit?") == MessageBoxResult.Yes)
-                    {
-                        if (this.ShowQuestion("Save changes first?") == MessageBoxResult.Yes)
+                    MessageBoxResult result = this.ShowQuestionWithCancel("Save changes before quitting?");
+
+                    if (result == MessageBoxResult.Yes) {
+                        quit = true;
+
+                        if (scriptDialog is not null)
                         {
-                            if (scriptDialog is not null)
-                            {
-                                if (this.ShowQuestion("Script still runs. Save anyway?\nIt can corrupt the data file that you'll save.") == MessageBoxResult.Yes)
-                                    save = true;
-                            }
-                            else
+                            if (this.ShowQuestion("Script still runs. Save anyway?\nIt can corrupt the data file that you'll save.") == MessageBoxResult.Yes)
                                 save = true;
-
-                            if (save)
-                            {
-                                SaveResult saveRes = await SaveCodeChanges();
-
-                                if (saveRes == SaveResult.NotSaved)
-                                    _ = DoSaveDialog();
-                                else if (saveRes == SaveResult.Error)
-                                {
-                                    this.ShowError("The changes in code editor weren't saved due to some error in \"SaveCodeChanges()\".");
-                                    return;
-                                }
-                            }
                         }
                         else
-                            RevertProfile();
-
-                        DestroyUMTLastEdited();
+                            save = true;
+                    } else if (result == MessageBoxResult.No) {
+                        quit = true;
                     }
-                    else
-                        return;
                 }
                 else
                 {
-                    RevertProfile();
-                    DestroyUMTLastEdited();
+                    quit = true;
                 }
 
-                if (SettingsWindow.UseGMLCache && Data?.GMLCache?.Count > 0 && !Data.GMLCacheWasSaved && Data.GMLCacheIsReady)
-                    if (this.ShowQuestion("Save unedited code cache?") == MessageBoxResult.Yes)
-                        await SaveGMLCache(FilePath, save);
+                if (quit) {
+                    if (save) {
+                        SaveResult saveRes = await SaveCodeChanges();
 
-                CloseOtherWindows();
+                        if (saveRes == SaveResult.NotSaved)
+                            _ = DoSaveDialog();
+                        else if (saveRes == SaveResult.Error)
+                        {
+                            this.ShowError("The changes in code editor weren't saved due to some error in \"SaveCodeChanges()\".");
+                            return;
+                        }
+                    } else {
+                        RevertProfile();
+                    }
 
-                IsAppClosed = true;
+                    DestroyUMTLastEdited();
 
-                Closing -= DataWindow_Closing; //disable "on window closed" event handler (prevent recursion)
-                _ = Task.Run(() => Dispatcher.Invoke(Close));
+                    if (SettingsWindow.UseGMLCache && Data?.GMLCache?.Count > 0 && !Data.GMLCacheWasSaved && Data.GMLCacheIsReady)
+                        if (this.ShowQuestion("Save unedited code cache?") == MessageBoxResult.Yes)
+                            await SaveGMLCache(FilePath, save);
+
+                    CloseOtherWindows();
+
+                    IsAppClosed = true;
+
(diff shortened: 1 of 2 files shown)


UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -772,58 +772,59 @@ private async void DataWindow_Closing(object sender, CancelEventArgs e)
...
-                                    _ = DoSaveDialog();
-                                else if (saveRes == SaveResult.Error)
-                                {
-                                    this.ShowError("The changes in code editor weren't saved due to some error in \"SaveCodeChanges()\".");
-                                    return;
-                                }
-                            }
+                        if (scriptDialog is not null)
+                        {
+                            if (this.ShowQuestion("Script still runs. Save anyway?\nIt can corrupt the data file that you'll save.") == MessageBoxResult.Yes)
```
Miepee, line 786: please combine these two checks into one.

UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -772,58 +772,59 @@ private async void DataWindow_Closing(object sender, CancelEventArgs e)
...
-                    RevertProfile();
-                    DestroyUMTLastEdited();
+                    quit = true;
                 }
 
-                if (SettingsWindow.UseGMLCache && Data?.GMLCache?.Count > 0 && !Data.GMLCacheWasSaved && Data.GMLCacheIsReady)
-                    if (this.ShowQuestion("Save unedited code cache?") == MessageBoxResult.Yes)
-                        await SaveGMLCache(FilePath, save);
+                if (quit) {
+                    if (save) {
```
Miepee, line 801: brace style

UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -772,58 +772,59 @@ private async void DataWindow_Closing(object sender, CancelEventArgs e)
...
+                        SaveResult saveRes = await SaveCodeChanges();
 
-                CloseOtherWindows();
+                        if (saveRes == SaveResult.NotSaved)
+                            _ = DoSaveDialog();
```
Miepee, line 805: see dangling comment above

UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -772,58 +772,59 @@ private async void DataWindow_Closing(object sender, CancelEventArgs e)
...
+                        else if (saveRes == SaveResult.Error)
+                        {
+                            this.ShowError("The changes in code editor weren't saved due to some error in \"SaveCodeChanges()\".");
+                            return;
+                        }
+                    } else {
```
Miepee, line 811: brace style

UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -772,58 +772,59 @@ private async void DataWindow_Closing(object sender, CancelEventArgs e)
...
+                        RevertProfile();
+                    }
 
-                IsAppClosed = true;
+                    DestroyUMTLastEdited();
+
+                    if (SettingsWindow.UseGMLCache && Data?.GMLCache?.Count > 0 && !Data.GMLCacheWasSaved && Data.GMLCacheIsReady)
+                        if (this.ShowQuestion("Save unedited code cache?") == MessageBoxResult.Yes)
```
Miepee, line 818: combine these two checks into one.

UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -772,58 +772,59 @@ private async void DataWindow_Closing(object sender, CancelEventArgs e)
...
+                            await SaveGMLCache(FilePath, save);
+
+                    CloseOtherWindows();
+
+                    IsAppClosed = true;
 
-                Closing -= DataWindow_Closing; //disable "on window closed" event handler (prevent recursion)
-                _ = Task.Run(() => Dispatcher.Invoke(Close));
+                    Closing -= DataWindow_Closing; //disable "on window closed" event handler (prevent recursion)
```
Miepee, line 825:
```suggestion
                    Closing -= DataWindow_Closing; // Disable "on window closed" event handler (prevent recursion)
```

UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -772,58 +772,59 @@ private async void DataWindow_Closing(object sender, CancelEventArgs e)
...
-                        if (this.ShowQuestion("Save changes first?") == MessageBoxResult.Yes)
-                        {
-                            if (scriptDialog is not null)
-                            {
-                                if (this.ShowQuestion("Script still runs. Save anyway?\nIt can corrupt the data file that you'll save.") == MessageBoxResult.Yes)
-                                    save = true;
-                            }
-                            else
-                                save = true;
+                    MessageBoxResult result = this.ShowQuestionWithCancel("Save changes before quitting?");
```
Miepee, line 779: removing the "are you sure you want to quit" is probably fine to do for now, but if you could add a comment that mentions to only ask whether changes should be saved if the file was changed, and ask whether the user really wants to quit otherwise would be nice.

UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -772,58 +772,59 @@ private async void DataWindow_Closing(object sender, CancelEventArgs e)
...
                 {
-                    RevertProfile();
-                    DestroyUMTLastEdited();
+                    quit = true;
                 }
 
-                if (SettingsWindow.UseGMLCache && Data?.GMLCache?.Count > 0 && !Data.GMLCacheWasSaved && Data.GMLCacheIsReady)
-                    if (this.ShowQuestion("Save unedited code cache?") == MessageBoxResult.Yes)
-                        await SaveGMLCache(FilePath, save);
+                if (quit) {
```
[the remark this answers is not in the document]
luizzeroxis: I'm pretty sure that if you press cancel quit isn't true.

UndertaleModTool/MainWindow.xaml.cs
luizzeroxis: I don't think they can be combined since if you answer no it should not set the save variable

Miepee: the can. `if (scriptDialog is not null && this.ShowQuestion(blabla)) save = true;`

UndertaleModTool/MainWindow.xaml.cs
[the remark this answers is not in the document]
Miepee: you're right, I got confused by the nesting.
You should make cancel just early return in the function, and thus assume that every other branch just automatically quits.

UndertaleModTool/MainWindow.xaml.cs
luizzeroxis: I don't think that works but now I realized this could work:

`if (scriptDialog is null || this.ShowQuestion(blabla)) save = true;`

Then remove the else.

Honestly though, who cares about such a small thing lol

Miepee: I feel that it improves readability much more, so would appreciate that being implemented.

luizzeroxis: I think it's worse readability but sure thing boss

UndertaleModTool/MainWindow.xaml.cs
luizzeroxis: I don't know how to get if changes have been made; maybe that's for another PR.

Miepee: it absolutely is, was mostly just mentioning that adding a comment for that would be neat.

Miepee (approved):
Comment can be added later, good enough to be merged now.
Ty for the PR!

PR #1891, opened 2024-08-26
Update workflows to setup .NET 8

colinator27:
## Description
We already updated to .NET 8, but the GitHub Actions workflows still set up .NET 6. This updates them.

### Caveats
Something may break in the build process, but unlikely.

### DIFF
--- a/.github/workflows/publish_cli.yml
+++ b/.github/workflows/publish_cli.yml
@@ -31,7 +31,7 @@ jobs:
     - name: Setup .NET
       uses: actions/setup-dotnet@v4
       with:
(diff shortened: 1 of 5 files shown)


Miepee:
Why did this not fail CI before?

colinator27:
> Why did this not fail CI before?

I *think* .NET 8 is included by default on the GitHub-hosted runners, so it probably "just worked" by coincidence.

Miepee:
then we may want to completely remove the wirth in the future maybe

Miepee (approved):

PR #1893, opened 2024-08-27
Manually add DEBUG constant to nightly builds

colinator27:
## Description
It seems that the updater button has completely vanished from nightly/bleeding edge builds, presumably because `DEBUG` isn't being defined correctly. Not sure if we should apply this to other workflows, too.

### Caveats
Not sure if there's some other configuration thing I'm missing, or if this is how the compiler is meant to work.

### DIFF
--- a/.github/workflows/publish_gui_nightly.yml
+++ b/.github/workflows/publish_gui_nightly.yml
@@ -37,7 +37,7 @@ jobs:
         dotnet build UndertaleModTool --no-restore
         dotnet build UndertaleModToolUpdater --no-restore
     - name: Publish ${{ matrix.os }} GUI
-      run: |
+      run: | # FIXME: debug constant should automatically be applied, but for some reason it isn't
         dotnet publish UndertaleModTool -c ${{ matrix.configuration }} -r win-x64 -p:DefineConstants="SHOW_COMMIT_HASH;DEBUG" --self-contained ${{ matrix.bundled }} -p:PublishSingleFile=${{ matrix.singlefile }} --output ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


Miepee (changes requested):
This seems wrong, the -c configuration should already handle DEBUG being set.
Maybe it needs to be -c DEBUG instead of -c Debug?

colinator27:
> This seems wrong, the -c configuration should already handle DEBUG being set. Maybe it needs to be -c DEBUG instead of -c Debug?

I don't think so, the configuration name *should* be `Debug`. Not sure what's supposed to control `DEBUG` being defined, but it's definitely not in any project files right now. Some docs on https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives suggest that it should be defined automatically, but it's clearly not happening right now. Maybe it's just an IDE thing for some reason?

.github/workflows/publish_gui_nightly.yml
```diff
@@ -38,7 +38,7 @@ jobs:
         dotnet build UndertaleModToolUpdater --no-restore
     - name: Publish ${{ matrix.os }} GUI
       run: |
-        dotnet publish UndertaleModTool -c ${{ matrix.configuration }} -r win-x64 -p:DefineConstants="SHOW_COMMIT_HASH" --self-contained ${{ matrix.bundled }} -p:PublishSingleFile=${{ matrix.singlefile }} --output ${{ matrix.os }}
+        dotnet publish UndertaleModTool -c ${{ matrix.configuration }} -r win-x64 -p:DefineConstants="SHOW_COMMIT_HASH;DEBUG" --self-contained ${{ matrix.bundled }} -p:PublishSingleFile=${{ matrix.singlefile }} --output ${{ matrix.os }}
```
Miepee, line 41: Can you add a comment above this with `# FIXME: debug constant should automatically be applied, but for some reason it isn't`

Miepee (changes requested):

Miepee (approved):

PR #1894, opened 2024-08-27
Clean up some warnings and errors

colinator27:
## Description
PR checks have been plagued with tons of warning messages at the end (at least with GitHub's beta "Unchanged files with check annotations"), so this PR aims to get rid of a bunch of them.

### Caveats
`WebClient` was replaced with `HttpClient`, so there may be slight behavioral differences. I did test the updater, though, and it functioned fine.

Some files now have `#pragma` directives to disable cross-platform compatibility warnings, but I limited this specifically to GUI, which is already Windows-only, at least as it stands.

### Notes
~~Starting this as a draft to make sure the warnings no longer show up in the build pipeline.~~ This seems to have worked.


UndertaleModCli/CommandOptions/DumpOptions.cs
```diff
@@ -2,6 +2,9 @@
 
 namespace UndertaleModCli;
 
+#nullable enable
```
Miepee, line 5: Please don't. Either actually fix the warnings, or keep them as is. Silently repressing them just hides that there's an issue present.

UndertaleModCli/CommandOptions/InfoOptions.cs
```diff
@@ -2,6 +2,9 @@
 
 namespace UndertaleModCli;
 
+#nullable enable
```
Miepee, line 5: .

UndertaleModCli/CommandOptions/LoadOptions.cs
```diff
@@ -2,6 +2,9 @@
 
 namespace UndertaleModCli;
 
+#nullable enable
```
Miepee, line 5: .

UndertaleModCli/CommandOptions/NewOptions.cs
```diff
@@ -2,6 +2,9 @@
 
 namespace UndertaleModCli;
 
+#nullable enable
```
Miepee, line 5: .

UndertaleModCli/CommandOptions/ReplaceOptions.cs
```diff
@@ -2,6 +2,9 @@
 
 namespace UndertaleModCli;
 
+#nullable enable
```
Miepee, line 5: .

UndertaleModCli/Program.cs
```diff
@@ -731,8 +731,8 @@ private void RunCSharpFile(string path)
     /// Evaluates and executes given C# code.
     /// </summary>
     /// <param name="code">The C# string to execute</param>
-    /// <param name="scriptFile">The path to the script file where <see cref="code"/> was executed from.
-    /// Leave as null, if it wasn't executed from a script file</param>
+    /// <param name="scriptFile">The path to the script file where the code was loaded from.
```
Miepee, line 734: I'd much rather use something like paramref instead of getting rid of it entirely.
(at least iirc paramref should work?)

UndertaleModLib/Scripting/IScriptInterface.cs
```diff
@@ -313,9 +313,9 @@ bool AreFilesIdentical(string file1, string file2)
     Task ClickableSearchOutput(string title, string query, int resultsCount, IDictionary<string, List<(int lineNum, string codeLine)>> resultsDict, bool showInDecompiledView, IEnumerable<string> failedList = null);
 
     /// <summary>
-    /// Sets <see cref="isFinishedMessageEnabled"/>.
+    /// Sets whether or not the script's "finished message" should be enabled, upon the current script's completion.
```
Miepee, line 316:
```suggestion
    /// Sets whether or not a script should show a message indicating that it completed.
```

UndertaleModTool/Windows/FindReferencesTypesDialog/UndertaleResourceReferenceMethodsMap.cs
```diff
@@ -1442,8 +1442,6 @@ public static async Task<Dictionary<string, List<object>>> GetUnreferencedObject
         {
             UndertaleResourceReferenceMethodsMap.data = data;
 
-            var ver = (data.GeneralInfo.Major, data.GeneralInfo.Minor, data.GeneralInfo.Release);
```
Miepee, line 1445: unused variable?

UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -3084,6 +3075,8 @@ public async void UpdateApp(SettingsWindow window)
             string tempFolder = Path.Combine(Path.GetTempPath(), "UndertaleModTool");
             Directory.CreateDirectory(tempFolder); // We're about to download, so make sure the download dir actually exists
 
+            string downloadOutput = Path.Combine(tempFolder, "Update.zip.zip");
```
Miepee, line 3078: Please move the webclient seperation to its own MR.

Miepee (changes requested):

UndertaleModCli/CommandOptions/DumpOptions.cs
colinator27: There's already a ReSharper disable for this, which specifies that the properties are applied via reflection. I'm not sure what the alternative to this is.

luizzeroxis: That may be true for some of these but not all of them. Would it not be better to enable <Nullable> in the project file and add ? to the types when necessary, since that's the default in .NET 8 anyway?

UndertaleModTool/Windows/FindReferencesTypesDialog/UndertaleResourceReferenceMethodsMap.cs
colinator27: Yep.

UndertaleModCli/CommandOptions/DumpOptions.cs
colinator27: I'm considering (for now) just disabling and removing nullables from the CLI project altogether. Fully adding nullables is a bit of a task, especially since it interfaces with UndertaleModLib, which does not have them enabled (although it should still work). Going through that (and possibly also on the library/GUI) can be for later PRs. Any thoughts?

luizzeroxis: So you would remove them now to add them later? Bit of double work but that's ok if doing all the null checks is too much for now.

colinator27: I'm not going to commit myself (or anyone else) to adding them later, at least for now. But yes, I'm considering removing them now so that, if we really want to commit to nullables in this project (the CLI one specifically, or UTMT as a whole), we can do so in later PRs.

colinator27: Removed the nullables for now. On a second look, it seems like nullables aren't being taken advantage of properly in the first place, anyway, so this seems like the best choice. Future PRs can add nullables project-wide if we want to do that.

UndertaleModCli/Program.cs
```diff
@@ -732,8 +732,8 @@ private void RunCSharpFile(string path)
     /// Evaluates and executes given C# code.
     /// </summary>
     /// <param name="code">The C# string to execute</param>
-    /// <param name="scriptFile">The path to the script file where <see cref="code"/> was executed from.
-    /// Leave as null, if it wasn't executed from a script file</param>
+    /// <param name="scriptFile">The path to the script file where <paramref name="code"/> was loaded from.
+    /// Leave as null, if it wasn't executed from a script file.</param>
     private void RunCSharpCode(string code, string scriptFile = null)
```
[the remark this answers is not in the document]
colinator27: Nullables are disabled for the CLI project, at least for now, yeah.

UndertaleModTests/GameScriptTests.cs
```diff
@@ -319,6 +319,20 @@ public string DummyString()
...
+        }
+
+        public Task ClickableSearchOutput(string title, string query, int resultsCount, IDictionary<string, List<(int lineNum, string codeLine)>> resultsDict, bool showInDecompiledView, IEnumerable<string> failedList = null)
+        {
+            throw new NotImplementedException();
+        }
+
+        public void ChangeSelection(object newSelection, bool inNewTab = false)
+        {
+        }
```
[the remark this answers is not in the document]
colinator27: Those are stubs for `IScriptInterface`; they were previously causing compile errors by not existing.

UndertaleModTool/Editors/UndertaleRoomEditor.xaml.cs
```diff
@@ -1,3 +1,5 @@
+#pragma warning disable CA1416 // Validate platform compatibility
```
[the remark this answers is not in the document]
colinator27: I think I would prefer it to only be used when absolutely necessary, rather than giving us the opportunity to accidentally write non-cross-platform code where we don't expect it.

Issue #1898, opened 2024-08-30
Null Reference Exception When Adding New Texture Page Item or Embedded Texture


Garethp:
### Describe the bug

When attempting to add a new Texture Page Item or Embedded Texture to the `data.win` file of Fields of Mistria, you receive an NRE. Specifically, adding a new Texture Page Item results in the following error

```
System.NullReferenceException: Object reference not set to an instance of an object.
   at UndertaleModTool.UndertaleTexturePageItemEditor.UpdateImages(UndertaleTexturePageItem item) in D:\SteamLibrary\steamapps\common\RimWorld\Mods\UndertaleModTool\UndertaleModTool\Editors\UndertaleTexturePageItemEditor.xaml.cs:line 43
   at UndertaleModTool.UndertaleTexturePageItemEditor.SwitchDataContext(Object sender, DependencyPropertyChangedEventArgs e) in D:\SteamLibrary\steamapps\common\RimWorld\Mods\UndertaleModTool\UndertaleModTool\Editors\UndertaleTexturePageItemEditor.xaml.cs:line 56
   at System.Windows.FrameworkElement.OnDataContextChanged(DependencyObject d, DependencyPropertyChangedEventArgs e)
   at System.Windows.FrameworkElement.OnPropertyChanged(DependencyPropertyChangedEventArgs e)
   at UndertaleModTool.DataUserControl.OnPropertyChanged(DependencyPropertyChangedEventArgs e) in D:\SteamLibrary\steamapps\common\RimWorld\Mods\UndertaleModTool\UndertaleModTool\Controls\DataUserControl.cs:line 17
   at System.Windows.DependencyObject.NotifyPropertyChange(DependencyPropertyChangedEventArgs args)
   at System.Windows.DependencyObject.UpdateEffectiveValue(EntryIndex entryIndex, DependencyProperty dp, PropertyMetadata metadata, EffectiveValueEntry oldEntry, EffectiveValueEntry& newEntry, Boolean coerceWithDeferredReference, Boolean coerceWithCurrentValue, OperationType operationType)
[...]

PR #1899, opened 2024-08-30
Issue #1898: Prevent NREs when adding new Texture Page Items or Embedded textures

Garethp:
Prevent NREs when adding new Texture Page Items or Embedded textures by adding more null checks. Closes issue #1898

## Description
As described in issue #1898, attempting to add new Texture Page Items or Embedded Textures results in Null Reference Exceptions being thrown. This MR attempts to mitigate those by adding more null value checks.

### Caveats
I didn't immediately see any tests that covered these areas. If I'm missing tests that I should have added, please let me know

### Notes
I'm not sure if there are any other NRE's that might be lying about. These only seem to fix the ones that I came across personally

### DIFF
--- a/UndertaleModLib/Models/UndertaleEmbeddedTexture.cs
+++ b/UndertaleModLib/Models/UndertaleEmbeddedTexture.cs
@@ -316,12 +316,12 @@ public class UndertaleEmbeddedTexture : UndertaleNamedResource, IDisposable
         /// <summary>
         /// The width of the texture.
         /// </summary>
-        public int Width => _image.Width;
+        public int Width => _image?.Width ?? -1;
 
         /// <summary>
         /// The height of the texture.
         /// </summary>
-        public int Height => _image.Height;
+        public int Height => _image?.Height ?? -1;
 
         /// <summary>
         /// Whether this texture uses the QOI format.
--- a/UndertaleModLib/Models/UndertaleEmbeddedTexture.cs
+++ b/UndertaleModLib/Models/UndertaleEmbeddedTexture.cs
@@ -316,21 +316,33 @@ public class UndertaleEmbeddedTexture : UndertaleNamedResource, IDisposable
         /// <summary>
         /// The width of the texture.
         /// </summary>
+        /// <remarks>
+        /// In the case of an invalid or missing image, this will always be <c>-1</c>.
+        /// </remarks>
         public int Width => _image?.Width ?? -1;
 
         /// <summary>
         /// The height of the texture.
         /// </summary>
+        /// <remarks>
+        /// In the case of an invalid or missing image, this will always be <c>-1</c>.
+        /// </remarks>
         public int Height => _image?.Height ?? -1;
 
         /// <summary>
         /// Whether this texture uses the QOI format.
         /// </summary>
+        /// <remarks>
+        /// In the case of an invalid or missing image, this will always be <c>false</c>.
+        /// </remarks>
         public bool FormatQOI => _image.Format is GMImage.ImageFormat.Qoi or GMImage.ImageFormat.Bz2Qoi;
 
         /// <summary>
         /// Whether this texture uses the BZ2 format. (Always used in combination with QOI.)
         /// </summary>
+        /// <remarks>
+        /// In the case of an invalid or missing image, this will always be <c>false</c>.
+        /// </remarks>
         public bool FormatBZ2 => _image.Format is GMImage.ImageFormat.Bz2Qoi;
 
         /// <summary>
--- a/UndertaleModLib/Models/UndertaleEmbeddedTexture.cs
+++ b/UndertaleModLib/Models/UndertaleEmbeddedTexture.cs
@@ -333,7 +333,7 @@ public class UndertaleEmbeddedTexture : UndertaleNamedResource, IDisposable
         /// Whether this texture uses the QOI format.
         /// </summary>
         /// <remarks>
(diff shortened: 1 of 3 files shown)


UndertaleModLib/Models/UndertaleEmbeddedTexture.cs
```diff
@@ -316,12 +316,12 @@ public GMImage Image
         /// <summary>
         /// The width of the texture.
         /// </summary>
```
colinator27, line 318: We probably want to document that this value can be -1 again, by adding:
```cs
<remarks>
In the case of an invalid or missing image, this will always be <c>-1</c>.
</remarks>
```
Same for `Height` as well.

colinator27: We should also document `FormatQOI` and `FormatBZ2` to always be `false`. I think pattern matching does the null checks automatically.

UndertaleModTool/Editors/UndertaleEmbeddedTextureEditor.xaml.cs
```diff
@@ -69,6 +69,9 @@ public UndertaleEmbeddedTextureEditor()
 
         private void UpdateImage(UndertaleEmbeddedTexture texture)
         {
+            if (texture.TextureData?.Image is null)
```
colinator27, line 72: If the `Image` property itself is set to null, we may want to actually clear the image sources in the UI, so they disappear.

UndertaleModTool/Editors/UndertaleTexturePageItemEditor.xaml.cs
```diff
@@ -40,6 +40,9 @@ public UndertaleTexturePageItemEditor()
 
         private void UpdateImages(UndertaleTexturePageItem item)
         {
+            if (item.TexturePage?.TextureData?.Image is null)
```
colinator27, line 43: Same as above, copying for clarity:
If the `Image` property itself is set to null, we may want to actually clear the image sources in the UI, so they disappear.

colinator27 (changes requested):
Looks good, I just have a few nitpicks we'll probably want to address to finalize it.

UndertaleModLib/Models/UndertaleEmbeddedTexture.cs
Garethp: I've added the comments, let me know if those look alright to you

UndertaleModTool/Editors/UndertaleEmbeddedTextureEditor.xaml.cs
Garethp: I think I added what you were after. Is that the way you wanted it done?

UndertaleModTool/Editors/UndertaleTexturePageItemEditor.xaml.cs
Garethp: As above: I think I added what you were after. Is that the way you wanted it done?

UndertaleModLib/Models/UndertaleEmbeddedTexture.cs
colinator27: Looks good to me!

UndertaleModTool/Editors/UndertaleEmbeddedTextureEditor.xaml.cs
colinator27: Looks good to me!

colinator27 (approved):

UndertaleModLib/Models/UndertaleEmbeddedTexture.cs
```diff
@@ -316,21 +316,33 @@ public GMImage Image
...
+        /// <remarks>
+        /// In the case of an invalid or missing image, this will always be <c>-1</c>.
+        /// </remarks>
+        public int Height => _image?.Height ?? -1;
 
         /// <summary>
         /// Whether this texture uses the QOI format.
         /// </summary>
+        /// <remarks>
+        /// In the case of an invalid or missing image, this will always be <c>false</c>.
```
Miepee, line 336: use `see langword` please instead of `c` for the false

UndertaleModLib/Models/UndertaleEmbeddedTexture.cs
```diff
@@ -316,21 +316,33 @@ public GMImage Image
...
+        /// </remarks>
         public bool FormatQOI => _image.Format is GMImage.ImageFormat.Qoi or GMImage.ImageFormat.Bz2Qoi;
 
         /// <summary>
         /// Whether this texture uses the BZ2 format. (Always used in combination with QOI.)
         /// </summary>
+        /// <remarks>
+        /// In the case of an invalid or missing image, this will always be <c>false</c>.
```
Miepee, line 344: .

Miepee (changes requested):

UndertaleModLib/Models/UndertaleEmbeddedTexture.cs
Garethp: Fixed

UndertaleModLib/Models/UndertaleEmbeddedTexture.cs
Garethp: Fixed

Miepee (approved):
Thanks for the fix!

PR #1412, opened 2023-06-16
Creation Codes Adding by AwfulNasty

XDOneDude:
## Description

This Code in UndertaleObjectReference.xaml.cs can add creation codes without any errors

PR #1901, opened 2024-09-04
Added way to add creation code in rooms and instances

### LINKED ISSUE #1900: Doesnt allow to add creation codes
### Describe the bug

изображение
"adding to non-objects is unsupported" while im trying to add creation code to an object

### Reproducing steps

1. select an object in the room
2. try to create a creation code by using the plus button

### Setup Details

1. 0.6.1.0
2. win 11
3. Pizza Tower

luizzeroxis:
## Description
Made so the little + icon on creation code in rooms and instances can actually create the script instead of giving an error. Fixes #1900. Based on code from #1412.

### Caveats
I tried to make it in a way that it could be expanded with other stuff. Maybe you could create derived classes from UndertaleObjectReference instead, so it doesn't clutter it up with a bunch of properties. Also the part that actually creates the code should probably be in the room model, and the part that creates arguments be separated and merged with the one that does that for object events already.

### Notes
None


UndertaleModTool/Controls/UndertaleObjectReference.xaml.cs
```diff
@@ -149,22 +182,31 @@ private void Details_Click(object sender, RoutedEventArgs e)
...
+                    else
+                    {
+                        if (!IsPreCreate)
+                        {
+                            ObjectReference = CreationCode(mainWindow.Data, "gml_RoomCC_" + Room.Name.Content + "_" + RoomGameObject.InstanceID.ToString() + "_Create");
+                        }
+                        else
+                        {
+                            ObjectReference = CreationCode(mainWindow.Data, "gml_RoomCC_" + Room.Name.Content + "_" + RoomGameObject.InstanceID.ToString() + "_PreCreate");
+                        }
```
Miepee, line 204: please condense these

UndertaleModTool/Controls/UndertaleObjectReference.xaml.cs
```diff
@@ -173,6 +215,33 @@ private void Details_Click(object sender, RoutedEventArgs e)
             }
         }
 
+        // TODO move this to the models
+        UndertaleCode CreationCode(UndertaleData data, string name)
```
Miepee, line 219: function name feels too generic.

This is adding a code entry, not creation code.
And I'm fairly certain we already have a function in the lib to do exactly that

Miepee (changes requested):

luizzeroxis:
Note that this will need to be changed if #1937 is merged

colinator27:
Made some changes to ensure duplicate code entries aren't accidentally created, and also added a null check for code locals while I was at it. Will be merging this as long as the builds all work, but I just tested this locally and everything was fine.

PR #1902, opened 2024-09-04
Add manifest to enable better styled dialogs

luizzeroxis:
## Description
A manifest file allows you to enable 'visual styles' which makes the buttons a bit better and styled to the OS.

Old:
old

New:
new

(screenshots taken on Windows 10)

### Caveats
None

### Notes
The other stuff is just the default. Better keep it there just in case

### DIFF
--- a/UndertaleModTool/app.manifest
+++ b/UndertaleModTool/app.manifest
@@ -1,0 +1,77 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<assembly manifestVersion="1.0" xmlns="urn:schemas-microsoft-com:asm.v1">
+  <assemblyIdentity version="1.0.0.0" name="MyApplication.app"/>
+  <trustInfo xmlns="urn:schemas-microsoft-com:asm.v2">
+    <security>
+      <requestedPrivileges xmlns="urn:schemas-microsoft-com:asm.v3">
+        <!-- UAC Manifest Options
+             If you want to change the Windows User Account Control level replace the 
+             requestedExecutionLevel node with one of the following.
+
+        <requestedExecutionLevel  level="asInvoker" uiAccess="false" />
+        <requestedExecutionLevel  level="requireAdministrator" uiAccess="false" />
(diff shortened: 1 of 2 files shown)


UndertaleModTool/app.manifest
Miepee, line 1: What exactly in this file is enabling that visual identity? All I see here are a bunch of comments which should be removed.

Dobby233Liu: https://github.com/UnderminersTeam/UndertaleModTool/pull/1902/files#diff-ef283036f5a7a52ff44df5a95240986ad29f640dadf267dc3d149feb32ec1902R63-R75

luizzeroxis: If you absolutely want to, the comments can be deleted, but they come in by default so maybe just keep them for reference.

VladiStep: I also would like to keep them, but exclude the following Windows versions: Vista, 7, 8, 8.1, since UTMT is .NET 8 now.

VladiStep: And I just did that myself

VladiStep (approved):

UndertaleModTool/app.manifest
```diff
@@ -0,0 +1,64 @@
...
+  </trustInfo>
+
+  <compatibility xmlns="urn:schemas-microsoft-com:compatibility.v1">
+    <application>
+      <!-- A list of the Windows versions that this application has been tested on
+           and is designed to work with. Uncomment the appropriate elements
+           and Windows will automatically select the most compatible environment. -->
+
+      <!-- Windows 10 -->
+      <!--<supportedOS Id="{8e0f7a12-bfb3-4fe8-b9a5-48fd50a15a9a}" />-->
```
VladiStep, line 30:
```suggestion
      <supportedOS Id="{8e0f7a12-bfb3-4fe8-b9a5-48fd50a15a9a}" />
```

Issue #1814, opened 2024-06-25
Search.csx popups are annoying


Couchy:
### Which component should be improved?

UI/UX

### Describe your feature suggestion in more detail

Search.csx creates two Yes/No popups for "Case sensitive?" and "Regex search?" before the text input popup. This gets tedious when doing multiple searches. My suggestion is to keep only the text input and have it search case insensitive by default, but also accept regex format to specify case sensitivity, etc, ex:

seatchtext -> Non-Regex, case-insensitive
/searchtext/ -> Regex, case-sensitive
/searchtext/i ->Regex, case-insensitive

Miepee:
Related to #545

colinator27:
Addressed by https://github.com/UnderminersTeam/UndertaleModTool/pull/1905

Issue #942, opened 2022-05-27
Add a keybind for quickly searching across the code


Grossley:
CTRL+Shift+F?

Miepee:
Probably better to be done after #545

colinator27:
Addressed by https://github.com/UnderminersTeam/UndertaleModTool/pull/1905

Issue #545, opened 2021-09-18
Feature request: make auto-backup and code search built in


Grossley:
> another thing : i always think why don't make auto backup and search a built-in thing instead of a script?

colinator27:
Partially addressed by https://github.com/UnderminersTeam/UndertaleModTool/pull/1905

PR #1905, opened 2024-09-06
Add built-in search in code function

luizzeroxis:
## Description
Added a search in code option (Ctrl+Shift+F or in the Search menu) that's meant to replace the Search.csx script. It's faster to open, has that shortcut, and has all the options in one screen. Related issues: #545, #942, #1814

test3

### Caveats
There's definitely improvements to be made

### Notes
Mostly copied from Search.csx ~~and ClickableTextOutput~~, so it's a little messy. Using LoaderDialog sucks, not sure if that should be improved or there should be a new loading bar inside the window itself. ~~That rich text box should probably be a table or list or something.~~ Searches should be cancellable. Also I'm not sure how to deal with all those threads being used, might not be 100% thread-safe.


VladiStep:
1) Have you checked the RAM usage for large search results?
2) Is it possible to save (export) the search results?
3) What are the advantages of using the table layout instead of text?

luizzeroxis:
@VladiStep 

> 1. Have you checked the RAM usage for large search results?

Did some quick tests in Visual Studio right now:

Snapshots № 1 is after loading DELTARUNE Chapter 1 & 2 DEMO's data.win, № 2 is after opening the search and getting to a text box, and № 3 is after showing the search results.

- Searching "test" (419 results)
  - Before
crop_old_test
  - After
crop_new_test

- Searching "a" (160805 results)
  - Before
crop_old_a
  - After
crop_new_a

> 2. Is it possible to save (export) the search results?

Yes, you can right click and copy all, or select the ones you want to copy. You can paste it in a spreadsheet if you want

> 3. What are the advantages of using the table layout instead of text?

The main reason is performance, the text box lagged massively when displaying large amounts of results. Other than that, most software uses a list of some kind when doing this sort of search. Sure, maybe in a grid it repeats too much the code name, but we could add a sort of grouping by that. The full line isn't visible (although you can hover over it to see it in full), but it's probably possible to make each line individually taller and with word wrap, but perhaps that's too cluttered in a grid format. BUT, uh, I just think it looks neater.

VladiStep:
Oh, wow, nice!

Miepee (commented):
no full review yet.
1) more gui screenshots?
2) This feels like something that should be mostly implemented in the Lib, not the GUI. Mostly for resusability.

colinator27 (approved):
Everything looks good to me. Seems like there's no major feature regressions, and the overall experience I would say is improved in most cases.

PR #1908, opened 2024-09-08
Replace WebClient with HttpClient

colinator27:
## Description
Originally was part of https://github.com/UnderminersTeam/UndertaleModTool/pull/1894, now is its own PR. Replaces `WebClient` with `HttpClient` in the updater code, as the former is obsolete (and was producing a warning).

### Caveats
There may be slight behavioral differences (e.g. download progress may update at a different interval). I did test the updater myself, though, and it functioned fine.

### Notes
N/A

### DIFF
--- a/UndertaleModTool/MainWindow.xaml.cs
+++ b/UndertaleModTool/MainWindow.xaml.cs
@@ -3084,6 +3084,8 @@ namespace UndertaleModTool
             string tempFolder = Path.Combine(Path.GetTempPath(), "UndertaleModTool");
             Directory.CreateDirectory(tempFolder); // We're about to download, so make sure the download dir actually exists
 
+            string downloadOutput = Path.Combine(tempFolder, "Update.zip.zip");
+
             // It's time to download; let's use a cool progress bar
             scriptDialog = new("Downloading", "Downloading new version...")
             {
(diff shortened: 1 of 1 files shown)


UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -3093,135 +3095,156 @@ public async void UpdateApp(SettingsWindow window)
...
-                    {
-                        // wait until progress bar updater loop is finished
-                        while (!ended)
-                            Thread.Sleep(100);
-
-                        scriptDialog = null;
-                    });
-
-                    if (e.Error is not null)
+                    using (HttpClient httpClient = new() { Timeout = TimeSpan.FromHours(1) })
```
Miepee, line 3102: can you change this to half an hour or 20 min instead?
1H is a waaay to big timeout IMO. Even if you have a worst case of our builds being 200MB and someone only downloading 200kb per second, thatd still be done in ~15min.

Or is this just for the download request? in which case it should definitely not be longer than 5min

UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -3093,135 +3095,156 @@ public async void UpdateApp(SettingsWindow window)
...
-                                        OpenBrowser("https://www.microsoft.com/en-us/download/details.aspx?id=44622");
-
-                                    window.UpdateButtonEnabled = true;
-                                    return;
+                                    bytesToProgressCounter -= bytesToUpdateProgress;
+                                    try
+                                    {
+                                        UpdateProgressStatus($"Downloaded MB: {(totalBytesDownloaded / bytesToMB).ToString("F2", CultureInfo.InvariantCulture)}");
+                                    }
+                                    catch { }
```
Miepee, line 3137: ?

UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -3093,135 +3095,156 @@ public async void UpdateApp(SettingsWindow window)
...
 
+                    // Extract ZIP
                     string updaterFolderTemp = Path.Combine(tempFolder, "Updater");
+                    bool extractedSuccessfully = false;
                     try
                     {
-                        if (Directory.Exists(updaterFolderTemp))
-                            Directory.Delete(updaterFolderTemp, true);
+                        // Unzip double-zipped update
+                        ZipFile.ExtractToDirectory(Path.Combine(tempFolder, "Update.zip.zip"), tempFolder, true);
```
Miepee, line 3155: use the downloadOutput variable here.

UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -3093,135 +3095,156 @@ public async void UpdateApp(SettingsWindow window)
...
+                        File.Move(Path.Combine(tempFolder, $"{patchName}.zip"), Path.Combine(tempFolder, "Update.zip"), true);
+                        File.Delete(Path.Combine(tempFolder, "Update.zip.zip"));
```
Miepee, line 3157: variable

UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -3093,135 +3095,156 @@ public async void UpdateApp(SettingsWindow window)
...
-                        }
-                        catch {}
+                            if (this.ShowQuestion(errMsg + win7upd, MessageBoxImage.Error) == MessageBoxResult.Yes)
+                                OpenBrowser("https://www.microsoft.com/en-us/download/details.aspx?id=44622");
 
-                        Thread.Sleep(100);
+                            window.UpdateButtonEnabled = true;
+                            return;
+                        }
                     }
```
Miepee, line 3237: can we just yoink this whole case?
We dont support win7, and if you have an SSL error then youll see that anyway.

Miepee (changes requested):

UndertaleModTool/MainWindow.xaml.cs
colinator27: Honestly this was here in the code before and I'm afraid to remove it now. Guessing there's some rare chance WPF throws something here, but not sure what that is. Do you think I should remove it, though?

UndertaleModTool/MainWindow.xaml.cs
colinator27: Yoink!

UndertaleModTool/MainWindow.xaml.cs
Miepee: Yeah id remove it. By looking at it, i cant ses any case where it would throw
And if it does, then thatll be useful for figuring it out when we dont silently catch!

colinator27: Alright, went ahead and removed it.

Miepee (approved):

PR #1929, opened 2024-09-27
Add support for games that use DDS textures

### LINKED ISSUE #1907: Error opening Home World Run's data.win file.
### Describe the bug

image

### Reproducing steps

1. Get Home World Run's SAGE 2024 Demo.
2. Open its data.win file with UMT
3. error

### Setup Details

1. UndertaleModTool Bleeding Edge (git: 9f4f13a)
2. Windows 11 
3. Game: https://sonicfangameshq.com/forums/showcase/home-world-run.2124/

luizzeroxis:
## Description
Fix #1907

Some games use the official GM-GPUTextureCompression extension to use GPU compressed textures. One of the formats supported is BCN compression, which uses the DDS format, at least on Windows. This PR aims to allow games using that to be modded.

### Caveats
Although you can load it, there's no support for saving these textures in their original format, so it just saves in PNG format and gives a warning to the user. This works, but obviously will have a performance impact. This compression format is lossy anyway.

### Notes
There's other formats supported by this extension, and in other operating systems, hopefully this can be addressed in future PRs.

### DIFF
--- a/UndertaleModLib/Util/GMImage.cs
+++ b/UndertaleModLib/Util/GMImage.cs
@@ -1,8 +1,8 @@
-﻿using ICSharpCode.SharpZipLib.BZip2;
-using ImageMagick;
-using System;
+﻿using System;
 using System.Buffers.Binary;
 using System.IO;
+using ICSharpCode.SharpZipLib.BZip2;
+using ImageMagick;
 
 namespace UndertaleModLib.Util;
 
@@ -34,7 +34,12 @@ public class GMImage
         /// <summary>
         /// BZip2 compression applied on top of GameMaker's custom variant of the QOI image file format.
         /// </summary>
-        Bz2Qoi
+        Bz2Qoi,
+
+        /// <summary>
+        /// DDS file format.
+        /// </summary>
+        Dds,
     }
 
     /// <summary>
@@ -76,6 +81,11 @@ public class GMImage
     /// Magic value found near the end of a BZip2 stream (square root of pi). 
     /// </summary>
     private static ReadOnlySpan<byte> MagicBz2Footer => new byte[] { 0x17, 0x72, 0x45, 0x38, 0x50, 0x90 };
+
+    /// <summary>
+    /// DDS file format magic.
+    /// </summary>
+    private static ReadOnlySpan<byte> MagicDds => "DDS "u8;
 
     /// <summary>
     /// Backing data for the image, whether compressed or not.
@@ -335,6 +345,14 @@ public class GMImage
             return FromQoi(reader.ReadBytes(12 + (int)compressedLength));
         }
 
+        // DDS
+        if (header.StartsWith(MagicDds))
+        {
+            // Read entire image
+            reader.Position = startAddress;
+            return FromDds(reader.ReadBytes((int)(maxEndOfStreamPosition - startAddress)));
+        }
+
         throw new IOException("Failed to recognize any known image header");
     }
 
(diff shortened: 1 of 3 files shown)


UndertaleModLib/Util/GMImage.cs
```diff
@@ -335,6 +345,14 @@ public static GMImage FromBinaryReader(IBinaryReader reader, long maxEndOfStream
             return FromQoi(reader.ReadBytes(12 + (int)compressedLength));
         }
 
+        // DDS
+        if (header.StartsWith(MagicDds))
+        {
+            // Read entire image
+            reader.Position = startAddress;
+            return FromDds(reader.ReadBytes((int)(maxEndOfStreamPosition - startAddress)));
```
colinator27, line 353: If possible, we should be finding the length of the image data using the image contents, rather than using `maxEndOfStreamPosition`, which can result in unnecessary padding when re-saving with modifications.

UndertaleModLib/Util/GMImage.cs
```diff
@@ -452,6 +470,31 @@ public static GMImage FromQoi(byte[] data)
...
+
+        // Get height and width
+        int height = (int)BinaryPrimitives.ReadUInt32LittleEndian(span[12..16]);
+        int width = (int)BinaryPrimitives.ReadUInt32LittleEndian(span[16..20]);
+
+        // Create wrapper image
+        return new GMImage(ImageFormat.Dds, width, height, data);
+    }
+
+    private void AddMagickToPngSettings(MagickReadSettings settings)
```
colinator27, line 491: Unsure if this is going to stick around in the long run, but this can probably be marked `static`.

colinator27 (changes requested):
Small nitpicks, but looks good on the whole. Ideally before merging completely, we should investigate the feasibility of encoding the BCN format (assuming ImageMagick doesn't fully support it?).

UndertaleModLib/Util/GMImage.cs
luizzeroxis: Didn't think about that! Sadly there's no easy way to find the length in this format. I can get the size of file without the extra bdata2 field, because getting that is even harder.

UndertaleModLib/Util/GMImage.cs
luizzeroxis: Totally

sonic3XE:
There is another file extension that was never mentioned before. (e.g. `.PVR`)

luizzeroxis:
@sonic3XE Could you send a game that uses it, or tell which custom option and OS creates such format, so it can be tested?

sonic3XE:
image

It looks like Android has ASTC available.

sonic3XE:
@luizzeroxis
[game.droid](https://cdn.glitch.global/b7ab8e8a-7603-4d11-bd6f-f10b4508902c/game.droid)
This file contains a PVR file, UMT may fail at loading that texture.

sonic3XE:
GPU-compressed textures are referred as https://github.com/UnderminersTeam/UndertaleModTool/issues/1932 including DDS, ASTC, and GNF.

sigmagamma:
The artifacts for this have been deleted... Is there any way to get them? @luizzeroxis 
Thanks!

luizzeroxis:
@sigmagamma I updated it to the newest version, it should be downloadable now.

sigmagamma:
Thanks @luizzeroxis ! it worked!

PR #1939, opened 2024-10-05
Add ParentEntry to code editor

Skirlez:
## Description
This PR makes it so you can see the ParentEntry property in the code editor.
Often when searching for a specific function you'll find the code entry referencing it, but then have to manually search for the parent entry to view/edit its contents. This change would make it easier to go there, you just need to double click it.

The property won't be displayed when null (so it will only show up when relevant), and can't be edited or set to null.

Screenshot from Void Stranger's `gml_Script_scr_steam_achievement_get`:
image

### DIFF
--- a/UndertaleModTool/Controls/UndertaleObjectReference.xaml.cs
+++ b/UndertaleModTool/Controls/UndertaleObjectReference.xaml.cs
@@ -67,8 +67,8 @@ namespace UndertaleModTool
                 new FrameworkPropertyMetadata(true,
                     FrameworkPropertyMetadataOptions.BindsTwoWayByDefault));
 					
-        public static DependencyProperty CanDragIntoProperty =
-            DependencyProperty.Register("CanDragInto", typeof(bool),
+        public static DependencyProperty CanChangeProperty =
+            DependencyProperty.Register("CanChange", typeof(bool),
                 typeof(UndertaleObjectReference),
                 new FrameworkPropertyMetadata(true,
                     FrameworkPropertyMetadataOptions.BindsTwoWayByDefault));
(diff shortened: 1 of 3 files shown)


zivmaor:
Seems like a good addition.
One caveat with this is that you can change the ParantEntry by dragging a different one in, which makes no sense and doesn't update that entry's ChildEntries list, but it doesn't seem to cause any bugs I could find and it doesn't affect the saved data.win file so it might be fine to leave as is.

colinator27:
> Seems like a good addition to me. One caveat with this is that you can change the ParantEntry by dragging a different one in, which makes no sense and doesn't update that entry's ChildEntries list, but it doesn't seem to cause any bugs I could find and it doesn't affect the saved data.win file so it might be fine to leave as is.

I think this can be addressed by adding a boolean to `UndertaleModTool/Controls/UndertaleObjectReference.xaml.cs`, specifying whether any changes (by drag and drop) are allowed. That or with some fancy bindings, but I'm not too sure how WPF handles that.

Miepee:
@Skirlez can you send a new screenshot for GUI? Am interested in seeing a specific thing there.

UndertaleModTool/Controls/UndertaleObjectReference.xaml.cs
```diff
@@ -94,8 +100,14 @@ public Type ObjectType
 
         public bool CanRemove
         {
-            get { return (bool)GetValue(ObjectTypeProperty); }
-            set { SetValue(ObjectTypeProperty, value); }
+            get { return (bool)GetValue(CanRemoveProperty); }
+            set { SetValue(CanRemoveProperty, value); }
+        }
+
+        public bool CanDragInto
```
Miepee, line 107: This attribute feels like it should be "can be changed" instead.

Miepee (approved):
Hm, could maybe do something else UX wise to indicate more that its a read only field, but I can't think of anything right now.
Thanks for the PR!

PR #1942, opened 2024-10-06
Fix UTMT crash when searching for a lot of unreferenced assets

maxiterr1:
For example, in Undertale, if you tried to find unreferenced assets in all groups UTMT would crash. This fixes that by using just a try catch.

## Description
Old error:
old error

PR #1947, opened 2024-10-08
Find references crash fix

### LINKED ISSUE #1946: UTMT crashes when using "Find unreferenced assets" on Undertale version 1.001
### Describe the bug

image
Seems to be because this data.win contains no LANG chunk.

### Reproducing steps

1. Open Undertale version 1.001 in UTMT
2. Try to use "Find unreferenced assets"
   * Make sure strings are selected.
4. The tool crashes.

### Setup Details

Tested on latest commit
data.win is Undertale version 1.001 (Windows, Steam), MD5: cd48b89b6ac6b2d3977f2f82726e5f12

zivmaor:
## Description

Fixes #1946 by only checking language for bytecode 16 and above games.
Based on https://github.com/UnderminersTeam/UndertaleModTool/wiki/Bytecode-version-differences.

Also adds error handling to all functions which use GetReferencesOfObject so if something similar happens again it won't cause the tool to crash.
### Notes

Replaces #1942.
Thanks to @maxiterr1 for bringing this to my attention.

### DIFF
--- a/UndertaleModTool/Windows/FindReferencesTypesDialog/FindReferencesTypesDialog.xaml.cs
+++ b/UndertaleModTool/Windows/FindReferencesTypesDialog/FindReferencesTypesDialog.xaml.cs
@@ -22,6 +22,8 @@ namespace UndertaleModTool.Windows
     /// </summary>
     public partial class FindReferencesTypesDialog : Window
     {
+        private static readonly MainWindow mainWindow = Application.Current.MainWindow as MainWindow;
+
         private readonly UndertaleResource sourceObj;
         private readonly UndertaleData data;
         private readonly bool dontShowWindow = false;
@@ -140,9 +142,21 @@ namespace UndertaleModTool.Windows
                     return;
                 }
 
-                var results = UndertaleResourceReferenceMethodsMap.GetReferencesOfObject(sourceObj, data, typesList);
-                FindReferencesResults dialog = new(sourceObj, data, results);
-                dialog.Show();
+                FindReferencesResults dialog = null;
+                try
+                {
+                    var results = UndertaleResourceReferenceMethodsMap.GetReferencesOfObject(sourceObj, data, typesList);
+                    dialog = new(sourceObj, data, results);
+                    dialog.Show();
+                }
+                catch (Exception ex)
+                {
+                    mainWindow.ShowError("An error occured in the object references related window.\n" +
+                                         $"Please report this on GitHub.\n\n{ex}");
+                    dialog?.Close();
+
+                }
+
             }
             else
             {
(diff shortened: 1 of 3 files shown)


maxiterr1 (approved):

UndertaleModTool/Windows/FindReferencesTypesDialog/UndertaleResourceReferenceMethodsMap.cs
```diff
@@ -1577,6 +1590,9 @@ await Task.Run(() =>
                 funcReferences = null;
                 variReferences = null;
 
+                await mainWindow.StopProgressBarUpdater();
```
colinator27, line 1593: Super small nitpick, but we could probably use a `finally` block for this instead of duplicating the code. Otherwise, this looks good to me.

colinator27 (changes requested):

UndertaleModTool/Windows/FindReferencesTypesDialog/UndertaleResourceReferenceMethodsMap.cs
zivmaor: fixed

colinator27 (approved):

UndertaleModTool/Windows/FindReferencesTypesDialog/FindReferencesTypesDialog.xaml.cs
```diff
@@ -173,9 +187,19 @@ private async void SearchButton_Click(object sender, RoutedEventArgs e)
...
-                FindReferencesResults dialog = new(data, results);
-                dialog.Show();
+                FindReferencesResults dialog = null;
+                try
+                {
+                    var results = await UndertaleResourceReferenceMethodsMap.GetUnreferencedObjects(data, typesDict);
+                    dialog = new(data, results);
+                    dialog.Show();
+                }
+                catch (Exception ex)
```
Miepee, line 197: can we somehow trickle the exception type further instead of just having a catchall?

Miepee (changes requested):
blocking accidental merge, core dev can dismiss later if my above comment ends up being void

UndertaleModTool/Windows/FindReferencesTypesDialog/FindReferencesTypesDialog.xaml.cs
zivmaor: As far as I understand we can't. This function is called directly from a button in XAML, so if we don't catch it here the application will crash.
This isn't a silent catchall and it should never be reached unless there is a bug. It is only here to stop the application from crashing if there is a bug, which is exactly what happens in #1946 and is way worse than the search just failing and showing an error message.

colinator27: Yeah, if it's just to prevent the entire application stability from being compromised, I'm OK with a catchall like this. Just need confirmation from @Miepee too

Miepee: well i meant more having a specicif exception type
so instead of `catch (Exception ex)` have `catch (InvalidAccessException ex)` or whatever exception type is used here.

colinator27: So in this case the crash was caused by a null pointer dereference (which also became an AggregateException), but since the point of it is to protect the user in the future, regardless of whatever bugs get introduced in the search process, I think plain Exception is probably the best. It's not like there's documented behavior for a bug we're not sure exists yet.

Miepee: eh, sure

Miepee (approved):

PR #1986, opened 2024-12-01
Better compiler accuracy for functions named differently/the same as the containing script name

CST1229:
## Description
This is a port of a very tiny compiler assembly accuracy improvement from UTMTCE.

You know that one pushi.e instruction for function definitions that is -1 sometimes and -6 other times? That instructions's value is -1 if the function's name is the same as the script it's in, and -6 if it isn't.
Not sure what effect this has on the runtime (possibly none), but it does make a difference in the assembly.

### Caveats
This is done by comparing to the code entry name, I'm not sure if it should be done like that. (Iterating through all script entries might be cleaner but that would also probably be slower.)

### DIFF
--- a/UndertaleModLib/Compiler/AssemblyWriter.cs
+++ b/UndertaleModLib/Compiler/AssemblyWriter.cs
@@ -1376,9 +1376,18 @@ namespace UndertaleModLib.Compiler
                             cw.typeStack.Push(DataType.Variable);
                             cw.Emit(Opcode.Dup, DataType.Variable).Extra = 0;
                             if (isStructDef)
+                            {
                                 cw.Emit(Opcode.PushI, DataType.Int16).Value = (short)-16;
+                            }
+                            else if (cw.compileContext.OriginalCode?.Name?.Content == ("gml_GlobalScript_" + funcDefName.Text))
+                            {
+                                // -1 when the function is named the same as the script name
+                                cw.Emit(Opcode.PushI, DataType.Int16).Value = (short)-1;
+                            }
                             else
-                                cw.Emit(Opcode.PushI, DataType.Int16).Value = (short)-1; // todo: -6 sometimes?
+                            {
+                                cw.Emit(Opcode.PushI, DataType.Int16).Value = (short)-6; 
+                            }
                         }
                         break;
                     case Parser.Statement.StatementKind.ExprBinaryOp:


colinator27 (approved):

colinator27 (commented):
As a small note, I think I did this already in the `underanalyzer` branch as part of the small compiler upgrades. I think in GM 2024 and above they changed this behavior as well (so it only uses -1), but I can fix the merge conflict for that once this gets merged in.

UndertaleModLib/Compiler/AssemblyWriter.cs
```diff
@@ -1376,9 +1376,18 @@ private static void AssembleExpression(CodeWriter cw, Parser.Statement e, Parser
...
+                            }
+                            else if (cw.compileContext.OriginalCode?.Name?.Content == ("gml_GlobalScript_" + funcDefName.Text))
+                            {
+                                // -1 when the function is named the same as the script name
+                                cw.Emit(Opcode.PushI, DataType.Int16).Value = (short)-1;
+                            }
                             else
-                                cw.Emit(Opcode.PushI, DataType.Int16).Value = (short)-1; // todo: -6 sometimes?
+                            {
+                                cw.Emit(Opcode.PushI, DataType.Int16).Value = (short)-6;
```
Miepee, line 1389: Can you add a comment explaining tbe -6, similar to above? Probably cleaner to move the comment outside the else if  and have it explain both cases.

UndertaleModLib/Compiler/AssemblyWriter.cs
```diff
@@ -1376,9 +1376,18 @@ private static void AssembleExpression(CodeWriter cw, Parser.Statement e, Parser
                             cw.typeStack.Push(DataType.Variable);
                             cw.Emit(Opcode.Dup, DataType.Variable).Extra = 0;
                             if (isStructDef)
+                            {
                                 cw.Emit(Opcode.PushI, DataType.Int16).Value = (short)-16;
+                            }
+                            else if (cw.compileContext.OriginalCode?.Name?.Content == ("gml_GlobalScript_" + funcDefName.Text))
```
Miepee, line 1382: Seems kinda messy but also dont know any other cleaner way.
I guess linq maybe, but dunno if we have access to the functions here?

CST1229:
> As a small note, I think I did this already in the `underanalyzer` branch as part of the small compiler upgrades.

oh i didn't know

PR #1992, opened 2024-12-10
Better dark mode

luizzeroxis:
## Description
Added [WPFDarkTheme](https://github.com/AngryCarrot789/WPFDarkTheme) to the project, to replace the current dark mode, which doesn't use the proper theme system and doesn't work in many cases.

### Caveats
I couldn't manage to make it switch between the new dark mode and the regular non-themed mode, there's glitches I don't fully understand. Because of that LightTheme.xaml is being used as the light mode, which looks different but looks acceptable to me.

There's also some small changes to be made, because some colors were hard coded in, should be easy as long as they are spotted.

Also messages boxes still aren't dark...

### Notes
Lots of file changes since I had to both replace all the "Dark" variants of controls with the regular ones and make sure all Style tags have the BasedOn property, so they keep the theme going. I recommend looking at each commit to have a better idea of what's going on, if you want to.

Screenshot of this PR:

image

Screenshot of bleeding edge, for reference:

image


CST1229:
i don't like how light mode looks, mainly how every UI element is darkened by way too much (imo) instead of having an outline like before. the disabled menu bar items especialy look jarring (and context menus are also always dark which i don't think should happen)
image
image

(also this is a bit of a nitpick but i like the arrows better instead of the +/- for the resource tree items)

luizzeroxis:
@CST1229 I agree, these colors really aren't the best in light mode. Thankfully it should be easy to change the colors in LightTheme.xaml so it's more similar to the default theme that was there before. I'm not a designer so I won't risk making something terrible, unless no one steps up for that. Probably the accent color should be blue instead of orange.

One question though... what color is Papyrus's scarf? Just curious.

CST1229:
> One question though... what color is Papyrus's scarf? Just curious.

idk i'm not an undertale guy

PR #1993, opened 2024-12-12
Add flags editor

### LINKED ISSUE #75: A better editor for the flag fields
https://github.com/krzys-h/UndertaleModTool/commit/ce572216a3514c5ef7a222de37dd2b29f994f309#commitcomment-31303588

There seems to be a lot more flags in Options now, a better editor may be a good idea

luizzeroxis:
## Description
Added a FlagsBox that display all possible flag values of a field as checkboxes. They are hidden in an Expander initially to not create clutter. The TextBox, as it existed before, is still there, so the old method of changing flags still works. If there's unrecognized flags, it'll display the number, and the checkboxes won't override it, simply adding or removing flags normally.

The FlagsBox was added to general info, the sound editor and the room editor.

Example:

image

### Caveats
Alright, the Expander is a bit a ugly. But what isn't ugly in this, huh? I tried a ComboBox design, but modifying these built in controls is complicated.

### Notes
Fixes #75.


colinator27 (approved):
Tested locally and looks good!

PR #2005, opened 2024-12-31
UnknownAlways2 -> TileOffset

fixdude:
## Description
Changes UnknownAlways2 in the background editor to TileOffset (because that's what it is)

### Caveats
I don't know

### Notes
I don't know


colinator27:
Want to make sure this is correct before merging it in - how did you find out about this?

BenjaminUrquhart:
From what I can tell, `2` is always written here regardless of anything selected in the IDE.

colinator27:
Did my own digging into the runtime to figure out the actual truth behind the matter, and it turns out this is actually a version number! When the number is > 1, it tells the runtime to load the GMS2 fields that come after it. So this is incorrect, unfortunately.

PR #2037, opened 2025-02-10
Underanalyzer: Fix incorrect data type in tiles

crystallizedsparkle:
## Description
changes the data type of SourceX and SourceY inside of the UndertaleRoom.Tile class to be integers instead of unsigned integers to allow for negative numbers.

### Caveats
Theres a chance that I missed a point in which the variable is used, which should be ironed out with more testing.

### Notes
This was mostly a quick fix for an issue a friend found in Undertale Yellow.


colinator27:
Going to try retargeting this to master, but we'll see if things go horribly wrong or not.

colinator27:
Nope, that was disaster.

PR #2057, opened 2025-03-10
Update FontEditor.csx

GFOXSH:
A gap of 1 may cause issues rendering Undertale fonts such as fnt_maintext at non-standard resolutions.


colinator27 (approved):
Took a look and this does seem to produce more accurate results to the official font packer, so this seems good to me!

PR #2058, opened 2025-03-11
Another TouchControlsEnabler fix(sorry)

NC-devC:
## Description
Why i did this PR? I found that i was made an issue, while was trying to make this script work more stable, and im doing this to fix it. What kind of issue? I forgot to change scaling offset for the vertical part of controlers. So, visual interface was worked fine, but controls wasnt on their right spots technicaly. I fixed that now. Also, i deleted part, that was checked if user's device is mobile, because its useless, and could be dissapointing. So, now script should hopefully work more stable.

### Caveats
image
For some reason, on NXTALE version of Undertale, controls doesnt takes the right scale. But, i think it should work fine on mobile devices, because screen size would be different.

### Notes
Now touchcontrols should work more stable. But i think if game would change the gui size, controllers wouldn't take their correct spots technicaly. I dont know any sort of examples now.

### Tests:
Tests was made on destkop version of the UT Yellow(gms 2023) and Undertale(gms 1.0.6)
image
image


colinator27:
Hmm... Has this actually been tested on some varying mobile device/monitor resolutions? Want to make sure that this doesn't negatively affect anything accidentally before merging it in.

NC-devC:
I would test it now, thx for answer

NC-devC:
I tested it on the Deltarune(used booker's port as a base for test), and it works, but in deltarune for some reason(maybe other gui scale) text in the settings wasn't scaled right. But it donts influes on the game.

colinator27 (approved):
Looks good, then. Thank you! Feel free to make further PRs if any other issues arise.

Issue #2055, opened 2025-03-08
you cannot zoom in far enough


Tablemen:
### Which component should be improved?

UI/UX

### Describe your feature suggestion in more detail

you barely can zoom in while in the room view panel. i wanted to zoom in further, however i could not

colinator27:
Should be addressed by #2060.

PR #2060, opened 2025-03-13
Zoom changes

luizzeroxis:
## Description
The main thing this PR does is increase the max and min zoom levels and make them align with powers of 2, so it's possible to get to 200% or 50%. It also changes the embedded texture editor so it starts at 100% instead of at the max width. And the font glyph editor initial scale rounds to a power of 2 so you can have the alignment said before.

Helps with #2055.

### Caveats
None

### Notes
None


colinator27 (approved):
Tried it out, and it seems like a good improvement to me!

PR #2063, opened 2025-03-19
UndertaleModCli: Saving fix when replacing larger older file

Skirlez:
## Description
Previously, when saving and replacing an older data file with UndertaleModCli, if the older data file was bigger, UndertaleModCli would not truncate it down to the new size, causing it to have extra data at the end. Both GameMaker and UndertaleModTool seem reject loading data files with leftover data.

This PR modifies how UndertaleModCli saves to be more in line with UndertaleModTool, using a temp file which replaces the original older file if saving was successful.


colinator27 (approved):
Looks good to me - I'll make some small changes to formatting (and to how the file is moved, even in the existing GUI implementation) after merging, though.

PR #2064, opened 2025-03-20
More room editor tweaks (more CE backports)

CST1229:
## Description
This mostly backports UTMTCE's other changes to the room editor (first 2 commits), some of being improvements to AwfulNasty's original changes.

- Ctrl+V paste now snaps to the grid, mainly for GMS1 tile editing (if you don't want to use Alt+Click). Ctrl+Shift+V was added for if you don't want the snapping.
- Mostly fix middle-click scrolling going way slower (or way faster depending on the zoom) if it needs to expand the scroll viewer. It still has a very minor desync for some reason, but it's better than before.

Other changes:
- Fix automatic instance layer creation not working when a non-instance layer is selected. The instance layer creation feature itself was [actually originally created for UTMTCE](https://github.com/XDOneDude/UndertaleModToolCE/commit/22d9b3105cb6733f8925c2d3f5f9dfd3367bfa9a), and I guess AwfulNasty added it to his PR. (I was gonna backport it anyways though, so it doesn't matter). Anyways, this fix itself though, was not in UTMTCE.
- Point the user to the Auto button when trying to edit a 0-sized tilemap. The dialog for when the vertical size is 0 also no longer mistakenly says "horizontal size" instead of "vertical size".

### Caveats
The middle-click scrolling fix:
> It still has a very minor desync for some reason, but it's better than before.

### DIFF
--- a/UndertaleModTool/Editors/UndertaleRoomEditor/UndertaleRoomEditor.xaml
+++ b/UndertaleModTool/Editors/UndertaleRoomEditor/UndertaleRoomEditor.xaml
@@ -58,11 +58,13 @@
     <UserControl.CommandBindings>
         <CommandBinding Command="Copy" Executed="Command_Copy" />
         <CommandBinding Command="Paste" Executed="Command_Paste" />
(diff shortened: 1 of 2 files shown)


colinator27 (approved):
Looks good, gave it a test and it works well! Thanks!

PR #2072, opened 2025-03-24
Refactor list structures, change list base class to a generic observable list

colinator27:
## Description
Refactors all of the list classes inheriting from `UndertaleListBase<T>`, which itself is no longer an abstract class, and has been renamed to `UndertaleObservableList<T>`. This allows data models to take advantage of observable collections for lists that need to unserialize using `InternalAdd()` (for performance purposes), when those lists have an irregular format in the WAD that the other lists can't handle. This functionality is currently unused in this PR, but will be taken advantage of in at least `UndertaleGameObject` in a future project system PR, for the list of physics vertices.

In general, I took to cleaning up and commenting all of the list code, as well as moving try/catch statements outside of loops to reduce their overhead. I have no idea if this has any noticeable impact on performance, but if it does, it's probably marginally faster now.

### Caveats
There's probably an extremely small chance something has been broken with these changes, but I ran through my whole game test suite and everything saved/loaded identically byte-for-byte, as usual (and the tool operates as normal).

### Notes
It's a little bit hacky how the base observable list has to use reflection to get and modify an internal member of `Collection<T>`, but it was already that way previously, and I'm not sure if we have any better alternatives. Perhaps a future PR could address that, if there is another way.

### DIFF
--- a/UndertaleModLib/UndertaleLists.cs
+++ b/UndertaleModLib/UndertaleLists.cs
@@ -7,99 +7,124 @@ using UndertaleModLib.Models;
 
 namespace UndertaleModLib
 {
-    public abstract class UndertaleListBase<T> : ObservableCollection<T>, UndertaleObject
+    /// <summary>
+    /// Basic observable list type, usable for data models to be presented in interfaces.
+    /// </summary>
+    public class UndertaleObservableList<T> : ObservableCollection<T>
     {
+        // Reference to the private internal list inside Collection<T>
         private readonly List<T> internalList;
 
-        public UndertaleListBase()
-        {
-            try
-            {
-                FieldInfo itemsField = typeof(Collection<T>)
-                                       .GetField("items", BindingFlags.NonPublic | BindingFlags.Instance);
-                internalList = (List<T>)itemsField.GetValue(this);
+        // Internal list field (name is guaranteed not to change due to the underlying type being serializable)
+        private static readonly FieldInfo itemsField = typeof(Collection<T>).GetField("items", BindingFlags.NonPublic | BindingFlags.Instance);
 
-            }
-            catch (Exception e)
-            {
-                throw new UndertaleSerializationException($"{e.Message}\nwhile trying to initialize \"UndertalePointerList<{typeof(T).FullName}>\".");
-            }
+        /// <summary>
+        /// Creates a new observable list, which is empty and has default capacity.
+        /// </summary>
+        public UndertaleObservableList()
+        {
+            // Get ahold of internal list
+            internalList = (List<T>)itemsField.GetValue(this);
         }
 
-        /// <inheritdoc />
-        public abstract void Serialize(UndertaleWriter writer);
+        /// <summary>
+        /// Creates a new observable list, which is empty and has the specified initial capacity.
+        /// </summary>
+        public UndertaleObservableList(int capacity)
+        {
+            // Get ahold of internal list
+            internalList = (List<T>)itemsField.GetValue(this);
 
-        /// <inheritdoc />
-        public abstract void Unserialize(UndertaleReader reader);
+            // Set capacity directly
+            internalList.Capacity = capacity;
+        }
 
+        /// <summary>
+        /// Sets the capacity of the internal list of this collection.
+        /// </summary>
         public void SetCapacity(int capacity)
         {
-            try
-            {
-                internalList.Capacity = capacity;
-            }
-            catch (Exception e)
-            {
-                throw new UndertaleSerializationException($"{e.Message}\nwhile trying to \"SetCapacity()\" of \"UndertalePointerList<{typeof(T).FullName}>\".");
-            }
+            internalList.Capacity = capacity;
         }
+
+        /// <summary>
+        /// Sets the capacity of the internal list of this collection.
+        /// </summary>
         public void SetCapacity(uint capacity) => SetCapacity((int)capacity);
 
+        /// <summary>
+        /// Adds an item to the internal list of this collection (that is, without notifying any observers).
+        /// </summary>
+        /// <param name="item">Item to add.</param>
         public void InternalAdd(T item)
         {
             internalList.Add(item);
         }
     }
 
-    public class UndertaleSimpleList<T> : UndertaleListBase<T> where T : UndertaleObject, new()
+    /// <summary>
+    /// Simple list object, serialized as a 32-bit count followed immediately by all objects in the list.
+    /// </summary>
+    public class UndertaleSimpleList<T> : UndertaleObservableList<T>, UndertaleObject where T : UndertaleObject, new()
     {
         /// <inheritdoc />
-        public override void Serialize(UndertaleWriter writer)
+        public void Serialize(UndertaleWriter writer)
         {
-            writer.Write((uint)Count);
-            for (int i = 0; i < Count; i++)
+            // Write list count
+            int count = Count;
+            writer.Write((uint)count);
+
+            // Write actual objects
+            int i = 0;
+            try
             {
-                try
+                for (; i < count; i++)
                 {
-                    writer.WriteUndertaleObject<T>(this[i]);
+                    writer.WriteUndertaleObject(this[i]);
                 }
-                catch (UndertaleSerializationException e)
-                {
-                    throw new UndertaleSerializationException(e.Message + "\nwhile writing item " + (i + 1) + " of " + Count + " in a list of " + typeof(T).FullName, e);
-                }
+            }
+            catch (UndertaleSerializationException e)
+            {
+                throw new UndertaleSerializationException($"{e.Message}\nwhile writing item {i + 1} of {count} in a list of {typeof(T).FullName}", e);
             }
         }
 
         /// <inheritdoc />
-        public override void Unserialize(UndertaleReader reader)
+        public void Unserialize(UndertaleReader reader)
         {
+            // Read count and initialize list with that capacity
             uint count = reader.ReadUInt32();
             Clear();
             SetCapacity(count);
-            for (uint i = 0; i < count; i++)
+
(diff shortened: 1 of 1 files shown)


UndertaleModLib/UndertaleLists.cs
```diff
@@ -7,171 +7,214 @@
...
         private readonly List<T> internalList;
 
-        public UndertaleListBase()
-        {
-            try
-            {
-                FieldInfo itemsField = typeof(Collection<T>)
-                                       .GetField("items", BindingFlags.NonPublic | BindingFlags.Instance);
-                internalList = (List<T>)itemsField.GetValue(this);
+        // Internal list field (name is guaranteed not to change due to the underlying type being serializable)
```
Miepee, line 18: i dont quite get what the stuff within parantheses is supposed to mean.

UndertaleModLib/UndertaleLists.cs
```diff
@@ -7,171 +7,214 @@
...
+            {
+                throw new UndertaleSerializationException($"{e.Message}\nwhile writing item {i + 1} of {count} in a list of {typeof(T).FullName}", e);
+            }
         }
 
         /// <inheritdoc />
-        public override void Unserialize(UndertaleReader reader)
+        public void Unserialize(UndertaleReader reader)
         {
+            // Read count and initialize list with that capacity
```
Miepee, line 95: wording nit, but i'd prefer something like "set list with that capacity" instead.

UndertaleModLib/UndertaleLists.cs
```diff
@@ -7,171 +7,214 @@
...
         }
 
         /// <inheritdoc cref="UndertaleObject.UnserializeChildObjectCount(UndertaleReader)"/>
         public static uint UnserializeChildObjectCount(UndertaleReader reader)
         {
+            // Read base object count
             uint count = reader.ReadUInt32();
             if (count == 0)
+            {
+                // Short-circuit if there's no objects
```
Miepee, line 122: can you put the comment above the if statement please?

UndertaleModLib/UndertaleLists.cs
```diff
@@ -7,171 +7,214 @@
...
                 return 0;
+            }
 
-            uint totalCount = 0;
-
-            Type t = typeof(T);
-            if (t.IsAssignableTo(typeof(UndertaleResourceRef)))
+            // If the objects are resource refs, the size is known as 4 bytes
+            if (typeof(T).IsAssignableTo(typeof(UndertaleResourceRef)))
```
Miepee, line 127: why remove the variable t here? makes things easier to read imo especially since its usded 5 times

UndertaleModLib/UndertaleLists.cs
```diff
@@ -7,171 +7,214 @@
...
+                throw new UndertaleSerializationException($"{e.Message}\nwhile reading child object count of item {i + 1} of {count} in a list of {typeof(T).FullName}", e);
             }
 
             return totalCount;
         }
     }
 
-    public class UndertaleSimpleListString : UndertaleListBase<UndertaleString>
+    /// <summary>
+    /// Simple list object, specifically for strings, serialized as a count followed immediately by all strings in the list.
```
Miepee, line 172: see comment for utsimplelist

UndertaleModLib/UndertaleLists.cs
```diff
@@ -184,7 +227,10 @@ public static uint UnserializeChildObjectCount(UndertaleReader reader)
         }
     }
 
-    public class UndertaleSimpleListShort<T> : UndertaleListBase<T> where T : UndertaleObject, new()
+    /// <summary>
+    /// Simple list object, serialized as a 16-bit count followed immediately by all objects in the list.
+    /// </summary>
+    public class UndertaleSimpleListShort<T> : UndertaleObservableList<T>, UndertaleObject where T : UndertaleObject, new()
```
Miepee, line 233: see above

UndertaleModLib/UndertaleLists.cs
```diff
@@ -200,231 +246,314 @@ private void EnsureShortCount()
...
+            {
+                throw new UndertaleSerializationException($"{e.Message}\nwhile reading child object count of item {i + 1} of {count} in a list of {typeof(T).FullName}", e);
             }
 
             return totalCount;
         }
     }
 
-    public class UndertalePointerList<T> : UndertaleListBase<T> where T : UndertaleObject, new()
+    public class UndertalePointerList<T> : UndertaleObservableList<T>, UndertaleObject where T : UndertaleObject, new()
```
Miepee, line 340: docstring?

UndertaleModLib/UndertaleLists.cs
```diff
@@ -7,171 +7,214 @@
 
 namespace UndertaleModLib
 {
-    public abstract class UndertaleListBase<T> : ObservableCollection<T>, UndertaleObject
+    /// <summary>
+    /// Basic observable list type, usable for data models to be presented in interfaces.
+    /// </summary>
+    public class UndertaleObservableList<T> : ObservableCollection<T>
```
Miepee, line 13: for your future data models, are those intended to have their own serialize and unserialize functions as well?

UndertaleModLib/UndertaleLists.cs
```diff
@@ -7,171 +7,214 @@
...
+            {
+                throw new UndertaleSerializationException($"{e.Message}\nwhile writing item {i + 1} of {count} in a string-list", e);
             }
         }
 
         /// <inheritdoc />
-        public override void Unserialize(UndertaleReader reader)
+        public void Unserialize(UndertaleReader reader)
         {
+            // Read count and initialize list with that capacity
```
Miepee, line 201: as above

UndertaleModLib/UndertaleLists.cs
```diff
@@ -200,231 +246,314 @@ private void EnsureShortCount()
...
+            {
+                throw new UndertaleSerializationException($"{e.Message}\nwhile writing item {i + 1} of {count} in a list of {typeof(T).FullName}", e);
+            }
         }
 
         /// <inheritdoc />
-        public override void Unserialize(UndertaleReader reader)
+        public void Unserialize(UndertaleReader reader)
         {
+            // Read count and initialize list with that capacity
```
Miepee, line 273: ditto

UndertaleModLib/UndertaleLists.cs
```diff
@@ -200,231 +246,314 @@ private void EnsureShortCount()
...
         }
 
         /// <inheritdoc cref="UndertaleObject.UnserializeChildObjectCount(UndertaleReader)"/>
         public static uint UnserializeChildObjectCount(UndertaleReader reader)
         {
+            // Read base object count
             ushort count = reader.ReadUInt16();
             if (count == 0)
+            {
+                // Short-circuit if there's no objects
```
Miepee, line 300: ditto

UndertaleModLib/UndertaleLists.cs
```diff
@@ -200,231 +246,314 @@ private void EnsureShortCount()
...
                 return 0;
+            }
 
-            uint totalCount = 0;
-
-            Type t = typeof(T);
-            if (t.IsAssignableTo(typeof(IStaticChildObjectsSize)))
+            // If objects have a static child object size/count, simply multiply to determine the number of total objects
+            if (typeof(T).IsAssignableTo(typeof(IStaticChildObjectsSize)))
```
Miepee, line 305: ditto

UndertaleModLib/UndertaleLists.cs
```diff
@@ -200,231 +246,314 @@ private void EnsureShortCount()
...
+
+            // Write all pointers
             foreach (T obj in this)
+            {
                 writer.WriteUndertaleObjectPointer<T>(obj);
-            for (int i = 0; i < Count; i++)
+            }
+
+            // Write blobs, if necessary for the given type
+            if (typeof(T).IsAssignableTo(typeof(UndertaleObjectWithBlobs)))
```
Miepee, line 356: less necessary here but still preferrable

UndertaleModLib/UndertaleLists.cs
```diff
@@ -200,231 +246,314 @@ private void EnsureShortCount()
...
+                // Short-circuit if there's no objects
                 return 0;
+            }
 
-            uint totalCount = 0;
-
-            Type t = typeof(T);
-            if (t.IsAssignableTo(typeof(IStaticChildObjectsSize)))
+            // If objects have a static child object size/count, simply multiply to determine the number of total objects
+            if (typeof(T).IsAssignableTo(typeof(IStaticChildObjectsSize)))
```
Miepee, line 492: as above

UndertaleModLib/UndertaleLists.cs
colinator27: Reworded it now to be a bit more clear.

UndertaleModLib/UndertaleLists.cs
colinator27: Put the variable back in place.

UndertaleModLib/UndertaleLists.cs
Miepee: Was more thinking on whether we want to enforce a serialization and deserialization method, a la interface
But it sounds like for that, its not the physicsvertices that would be able to implement it, but rather the (de)serialization of undertalegameobject, so it doesnt work nicely?

colinator27: Yeah, this is deliberately *not* inheriting from `UndertaleObject` (the interface with `Serialize()`, etc.). Just meant for a list that can be added to efficiently during deserialization, without having any purview over how it's (de)serialized. (And then to be observable for the GUI as well.)

Miepee (approved):

PR #2091, opened 2025-04-03
feat: export backgrounds to png script

a2937:
## Description

I made a Community Script that generates pictures of the background for every room in the game. This script heavily used logic that existed in the "Export Room to PNG" script, but does so without any objects in the generated picture. 

### Caveats

It removes all of the objects in the rooms in order to generate the picture, but then puts the data all back. 

### Notes

This is something I will use hands down. I think the rest of the Community can probably use it as well.


a2937:
I need to stretch it 3 times to make it easier to see.

PR #2093, opened 2025-04-06
More builtin functions

CST1229:
## Description
Adds a lot of builtin functions to the builtins list (specifically, everything from 2023.11 fnames that wasn't in the list before. That was just the latest runtime I had on hand. Also I did a lot of complex regex stuff and manual editing to parse the file to turn it into DefineFunction calls)

### Caveats
Nothing post-2023.11, obviously. Also, I did not find the exact versions of each function (only ones I am very sure were added in 2023, and then everything else is just 2.3+).


colinator27 (approved):
I think it should be fine if the versions aren't exactly correct, as it'll just result in a load error for earlier versions, and we could correct them if need be. Looks good to me!

PR #2095, opened 2025-04-09
feat: try and fail to use ffmpeg

a2937:
## Description

I wanted to have the "Externalize all oggs" script return every audio file as an ogg. 

### Caveats

The externalize script behavior has changed and a new Nuget package has been added to the codebase. I think it has other requirements to convert audio. 

### Notes

I'm having trouble with the script itself. Some sort of weird curly braces problem.


VladiStep:
I guess the FFMPEG package will add at least 50 MB to the program package?
Then it should require to download FFMPEG by users and use it through command, so it's optional.

PR #2107, opened 2025-04-19
Settings window refactor

luizzeroxis:
## Description
Updated settings window so it's not based on a giant Grid, but instead StackPanels. This should make it easier to add or edit these settings in the future. As a bonus, you can now click the labels of checkboxes to check them

### Caveats
None

### Notes
None


colinator27 (approved):

colinator27:
Looks good, thanks! I'll probably follow up with a few tiny tweaks, but nothing too specific to this PR.

PR #2114, opened 2025-04-29
Add "Restore to defaults" button to (de)compiler settings window

### LINKED ISSUE #2112: Restore to default action button for compiler/decompiler settings

UI/UX

Restore to default action button for GML compiler/decompiler settings

CST1229:
Resolves #2112

## Description
exactly what it says on the tin
image

### Caveats
There's no confirmation. IDK whether there should be one or not (it should be easy to add). Also, the way it resets the settings might be a bit hacky (because I don't feel like PropertyChanged-ing), but It seems to work.

### DIFF
--- a/UndertaleModTool/Settings.cs
+++ b/UndertaleModTool/Settings.cs
@@ -64,7 +64,8 @@ namespace UndertaleModTool
         public bool EnableDarkMode { get; set; } = false;
         public bool ShowDebuggerOption { get; set; } = false;
         public DecompilerSettings DecompilerSettings { get; set; }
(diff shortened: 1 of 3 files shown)


colinator27:
Maybe the settings could be rewritten at some point, but at least right now, the `Settings` constructor actually changes the global instance to be itself. That meant that *all* settings were accidentally being reverted when clicking the button, so I've fixed the code to not use that constructor.

colinator27:
I also just rewrote the logic so that the decompiler settings instance is reused, rather than created from scratch. Otherwise, the updated settings may not go into effect until the next time a data file is loaded.

colinator27 (approved):

PR #2123, opened 2025-05-13
Clean up UndertaleSequence

colinator27:
## Description
Cleans up `UndertaleSequence`, changing out some technically undefined behavior of `Dictionary` to use regular lists, and also adding some basic documentation (though still incomplete for certain fields that I don't fully understand).

### Caveats
Potential incompatibility with existing code relying on sequence structures, although these changes should be easy to adapt to. This has been tested on my regular set of games for save/load regressions.

### Notes
I also modified the `Align` methods in `UndertaleReader`/`UndertaleWriter` to no longer accept a padding byte other than 0, because it was unused and had some incorrect logic associated.

Support was also added for `GMAudioEffectTrack`, which acts just like a track with real keyframes for our purposes.


colinator27:
Also forgot to mention, but this also resolves https://github.com/UnderminersTeam/UndertaleModTool/issues/2079.

colinator27:
Going to merge this in now as it seems low-risk (tested save/load on my suite of test games), and so that dependent PRs won't have conflict issues down the line.

PR #2125, opened 2025-05-17
feat: Making some scripts crossplatforms

Eliandro4:
Makes some scripts crossplatfrom

NewTextureRepacker, ExportAllStringsBetter, ImportAllStringsBetter, ImportASound and FindNullsInOffsetMap by removing WinForms use.

ImportMask by using ImageMagick instead of System.Drawing.

UndertaleDialogSimulator, FindObjectsinRoom, CopySound, GameOjectCopy, GameOjectCopyInternal, ImportSoundsBulk, SpriteOriginCopy and ExportSpritesAsGIF by removing unecessary "using windows_only_api" statements.

**NOTE**
I also added a method to the IScriptInterface to save files. Called "PromptSaveFile"

### DIFF
--- a/UndertaleModCli/Program.UMTLibInherited.cs
+++ b/UndertaleModCli/Program.UMTLibInherited.cs
@@ -389,6 +389,28 @@ public partial class Program : IScriptInterface
         return path;
     }
 
+    public string PromptSaveFile(string defaultExt, string filter)
+    {
+        string path;
+        do
+        {
+            Console.WriteLine("Please type a path (or drag and drop) to save the file:");
+            Console.Write("Path: ");
+            path = RemoveQuotes(Console.ReadLine());
+
+            if (Directory.Exists(path))
+            {
+                Console.WriteLine("You selected a directory");
+            }
+
+            if (string.IsNullOrWhiteSpace(Path.GetExtension(path)))
+            {
+                path += defaultExt.StartsWith(".") ? defaultExt : "." + defaultExt;
+            }
+        } while (string.IsNullOrWhiteSpace(path));
+        return path;
+    }
+
     /// <inheritdoc/>
     public string GetDecompiledText(string codeName, GlobalDecompileContext context = null, IDecompileSettings settings = null)
     {
--- a/UndertaleModLib/Scripting/IScriptInterface.cs
+++ b/UndertaleModLib/Scripting/IScriptInterface.cs
@@ -432,4 +432,12 @@ public interface IScriptInterface
     /// <param name="filter">The filters used for the file select.</param>
     /// <returns>The file selected by the user.</returns>
     string PromptLoadFile(string defaultExt, string filter);
+
+    /// <summary>
+    /// Used to prompt the user to select a location and name to save a file..
+    /// </summary>
+    /// <param name="defaultExt">The default extension that should be used when saving the file.</param>
+    /// <param name="filter">The filters used to specify the file types in the dialog.</param>
+    /// <returns>The file path specified by the user for saving.</returns>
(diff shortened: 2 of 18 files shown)


UndertaleModCli/Program.UMTLibInherited.cs
```diff
@@ -389,6 +389,24 @@ public string PromptLoadFile(string defaultExt, string filter)
...
+        string path;
+        do
+        {
+            Console.WriteLine("Please type a path (or drag and drop) to save the file:");
+            Console.Write("Path: ");
+            path = RemoveQuotes(Console.ReadLine());
+
+            if (Directory.Exists(path))
+            {
+                Console.WriteLine("You selected a directory");
```
colinator27, line 404: We should probably make this an error message (and add a `continue;` so that it doesn't let you use a directory at all).

UndertaleModTool/Scripts/Resource Repackers/ImportFontData.csx
```diff
@@ -257,13 +269,11 @@ public class Packer
         {
             string atlasName = $"{prefix}{atlasCount:000}.png";
             //1: Save images
-            Image img = CreateAtlasImage(atlas);
+            MagickImage img = CreateAtlasImage(atlas);
             //DPI fix start
-            Bitmap ResolutionFix = new Bitmap(img);
-            ResolutionFix.SetResolution(96.0F, 96.0F);
-            Image img2 = ResolutionFix;
+            img.Density = new Density(96, 96);
```
colinator27, line 274: Do we need this DPI fix anymore? Pretty sure this was due to some odd behavior of System.Windows.Drawing specifically. Has this new version of the script been tested?

UndertaleModTool/Scripts/Resource Repackers/ImportFontData.csx
```diff
@@ -433,42 +443,53 @@ public class Packer
...
+
                 if (DebugMode)
                 {
-                    string label = n.Bounds.Width.ToString() + "x" + n.Bounds.Height.ToString();
-                    SizeF labelBox = g.MeasureString(label, SystemFonts.MenuFont, new SizeF(n.Bounds.Size));
-                    RectangleF rectBounds = new Rectangle(n.Bounds.Location, new Size((int)labelBox.Width, (int)labelBox.Height));
-                    g.FillRectangle(Brushes.Black, rectBounds);
-                    g.DrawString(label, SystemFonts.MenuFont, Brushes.White, rectBounds);
+                    // Desenhar retângulo e texto com Drawables
+                    var drawables = new Drawables()
```
colinator27, line 462: Do we need this debug mode drawing anymore? Pretty sure it's not used normally, but if we want to keep it, these comments should probably be in English.

colinator27 (changes requested):
Overall, this looks really good! I just have a few questions and small things that will need corrections before being merged. Once those are done, I'll also make some formatting changes of my own, for consistency with the rest of the codebase. Thanks!

UndertaleModTool/Scripts/Resource Repackers/ImportFontData.csx
Eliandro4: Actually, no.

UndertaleModTool/Scripts/Resource Repackers/ImportFontData.csx
Eliandro4: No, it is no longer necessary.
Yes, it has been tested and works well.

UndertaleModCli/Program.UMTLibInherited.cs
Eliandro4: ok

PR #2139, opened 2025-06-08
ExportSpritesAsGIF.csx fix for frameless sprites

bariobro1:
This fixes the ExportSpritesAsGIF.csx script to not crash when it tries to export a sprite with no frames.

Note: This is my first pull request, and my first time using C#. It's not much, but it works.


colinator27 (approved):
Looks good, thanks!

PR #2140, opened 2025-06-08
Add Japanese translations

Cheesecake2960:
Hello! I've been working on the Japanese translation of UTMT over the past few days. Although it's still a work in progress, I'm opening this PR as a Draft to invite collaboration from others.
If time allows, I'd also like to work on translating UTMTUpdater and the CLI as well.

## TODO List
I'm sure there are many more, but here is the TODO list that we are aware of at this time.
- UTMT
  - [x] Menubar
  - [x] Data tree
  - [x] Save and load window
  - [ ] Editors
    - [x] Audio group editor
    - [x] Sound editor
    - [ ] Other editors
  - [ ] Message boxes (information and error dialogs)

> **Note to Japanese speakers:** 日本語を話せる方がいたら、できれば日本語でコメントお願いします！


colinator27:
Nice work so far! Will need to hear thoughts from others, but I think something we'll need to figure out if this is to be merged is how it'll be maintained. Realistically, I think we'll need to have a process where we can add/change English strings and have no corresponding strings in other languages (due to being untranslated). Then we can have translation PRs come in to address those missing strings, I guess. I'm sure there are better workflows, though, and I'm all ears. (But I'd rather not increase the general tool maintenance burden.)

Also just to check, is this all being manually translated?

Cheesecake2960:
> Nice work so far! Will need to hear thoughts from others, but I think something we'll need to figure out if this is to be merged is how it'll be maintained. Realistically, I think we'll need to have a process where we can add/change English strings and have no corresponding strings in other languages (due to being untranslated). Then we can have translation PRs come in to address those missing strings, I guess. I'm sure there are better workflows, though, and I'm all ears. (But I'd rather not increase the general tool maintenance burden.)
> 
> Also just to check, is this all being manually translated?

Yes, I used machine translation as a reference for these texts, but I did most of the translation myself.
(I am Japanese!)

Cheesecake2960:
Regarding maintenance when English strings are added or edited:
I think it's fine to temporarily use the English text as-is until someone can provide a translation.
This helps avoid confusing users with blank strings in other languages.

Also, when adding new English strings, it would be helpful if they could be placed in Resource files rather than being hardcoded in code-behind or XAML. That way, it makes translation easier later on.

CST1229:
it might be a better idea to do it on a separate branch on your fork too

PR #2141, opened 2025-06-09
Add zoom function for code editor

### LINKED ISSUE #1557: Make the code look larger with CTRL and +

UI/UX

A chrome zoom-like feature to make the decompiled code and the UI look larger

Cheesecake2960:
Resolves #1557

I have added a zoom feature to the code editor.
You can zoom with the Ctrl key and the mouse wheel or the + - key.


colinator27:
I haven't tested this yet (I'll do a proper review later), but does this behave well with multiple code tabs? Not sure what the expected behavior should actually be for those; maybe it would be expected to be the same between all tabs, always?

Cheesecake2960:
The test confirmed that the zoom level is applied to all tabs.

PR #2146, opened 2025-06-17
v1 of an updated DeltaHATE that works with the new chapters

hexadexx:
## Description
I've attempted to make something like DeltaHATE, for the newer versions of the game. I've tested it and it works as well as I can get it to, though as it is a randomizer behaviour may be inconsistent.

### Caveats
Due to some sprites breaking undertale when shuffled (culprits not yet found) there is a whole ton of softlocks. If anyone knows the sprites, please feel free to add them to the exclusion list.

### Notes
This is one of my first scripts, though I've been working on it for a while. Hope this is useful to at least someone.

### DIFF
--- a/UndertaleModTool/Scripts/UTDR Scripts/UTDRShuffler.csx
+++ b/UndertaleModTool/Scripts/UTDR Scripts/UTDRShuffler.csx
@@ -1,0 +1,1049 @@
+using System.Linq;
+using System.IO;
+
+EnsureDataLoaded();
+
+static Random rng;
+
+ScriptMessage("Thank you to redspah for creating HATE, krzys-h for creating DeltaHATE, and anyone else involved with them for being a massive help.");
+
+string seedInput = "0";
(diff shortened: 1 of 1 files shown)


sennaako:
What's DeltaHATE and why did it get closed? No reason was given.

Vào CN, 14 thg 9, 2025 lúc 12:25 Deleted user ***@***.***> đã
viết:

Jacky720:
HATE, aka Undertale Corruptions, is an old Undertale mod that randomizes sprites, dialogue, etc with the goal of producing funny moments in exchange for sometimes breaking the game. DeltaHATE is the same thing for Deltarune, which has been prototyped several times over the last seven years (including at least one that worked pretty well on chapters 1&2 but whose author requested it not be hosted with UTMT).

This pull request was closed automatically because the user who was hosting it (not the same as the user who opened the PR) apparently had their GitHub account deleted, and you can't pull changes from an account that doesn't exist.

PR #2154, opened 2025-06-19
Use the selected code for code search.

### LINKED ISSUE #2143: When using "Find in code" (`Ctrl+Shift+F`), use the selection from the code editor.

UI/UX

It would be neat if when you use "Find in code" feature, it would insert the currently selected code from opened code editor into the search query (e.g. like when you use `Ctrl+F` in a browser).
Though, I think it only should do that if the selection is not too long?

VladiStep:
## Description
Closes #2143.


UndertaleModTool/Windows/SearchInCodeWindow.xaml.cs
```diff
@@ -53,9 +53,20 @@ public partial class SearchInCodeWindow : Window
...
+            if (query is not null)
+            {
+                if (query.Length > 256 || query.Count(x => x == '\n') > 16)
+                    return; // Ignore if the query is longer than 256 characters or 16 lines.
+
+                SearchTextBox.Text = query;
+                SearchTextBox.SelectAll();
+            }
+
+            InAssemblyCheckBox.IsChecked = inAssembly;
```
colinator27, line 69: Very small logic thing, but I think the `return` from above will accidentally skip this if you're selecting a lot of text in diassembly.

VladiStep: Hm, but if the length is too big, then it shouldn't check the checkbox as it counts as you have selected nothing in disassembly?

colinator27: I would think the user *intended* to search in the disassembly if that's the standard selection behavior... but not sure. Your call, I guess. I don't care too much.

VladiStep:
> 0 conflicts

Ummm, okay
image

colinator27 (approved):
Did a quick test on my end, and it seems to work well!

PR #2176, opened 2025-06-26
Add the missing "Select fonts" functionality in "ExportAllFonts.csx" script.

VladiStep:
## Description
When there was the "ExportFontData.csx" script, there was a window that allowed you to select only few fonts for exporting.
This PR reimplements the selection window, but it's WPF instead of WinForms.
image

### Notes
@C-arth asked me to make this, so thanks to him.


VladiStep:
Oh, hm, fair point.
But for now, it can stay like that? It's not difficult to revert the changes when it will become cross-platform

PR #2177, opened 2025-06-27
Fix an obscure crash on right-clicking some blank spots in `DataGrid`.

VladiStep:
## Description
https://github.com/user-attachments/assets/aa300845-2256-4b29-b086-8e2deea0d242


colinator27 (approved):
Thanks for looking into this!

PR #2195, opened 2025-07-12
Use MaxBy instead of OrderBy when resolving Room Layer Instance IDs for performance

BioTomateDE:
## Description
Small Performance and code readability change in `UndertaleModLib/Models/UndertaleRoom.cs:395`.

### Caveats
Should not have any caveats.

### Notes
my first pr


colinator27 (approved):
Looks good, thanks!

PR #2204, opened 2025-07-24
Add scale transform to tiles in room

luizzeroxis:
## Description
Legacy tiles didn't have a scale transform at all, so I added one. This should fix tile rotation problems in Undertale Yellow.

### Caveats
None

### Notes
None

### DIFF
--- a/UndertaleModTool/Controls/UndertaleRoomRenderer.xaml
+++ b/UndertaleModTool/Controls/UndertaleRoomRenderer.xaml
@@ -235,7 +235,11 @@
                     <Rectangle Width="{Binding Width, Mode=OneTime}" Height="{Binding Height, Mode=OneTime}"
                                Opacity="{Binding Color, Mode=OneTime, Converter={StaticResource ColorToOpacityConverter}}">
                         <Rectangle.RenderTransform>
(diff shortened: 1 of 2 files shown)


UndertaleModTool/Controls/UndertaleRoomRenderer.xaml
```diff
@@ -235,7 +235,11 @@
                     <Rectangle Width="{Binding Width, Mode=OneTime}" Height="{Binding Height, Mode=OneTime}"
                                Opacity="{Binding Color, Mode=OneTime, Converter={StaticResource ColorToOpacityConverter}}">
                         <Rectangle.RenderTransform>
-                            <TranslateTransform x:Name="transform2_0" X="{Binding X, Mode=OneTime}" Y="{Binding Y, Mode=OneTime}"/>
+                            <TransformGroup>
+                                <TranslateTransform x:Name="transform2_0" X="{Binding X, Mode=OneWay}" Y="{Binding Y, Mode=OneWay}"/>
```
colinator27, line 239: Is there a reason this was changed from `OneTime` to `OneWay`? I'm guessing that was done that way for a performance reason, just not totally sure.

PR #2209, opened 2025-08-01
Fix room editor crashes with grid size of one

BenjaminUrquhart:
## Description
The room editor would crash if your grid size was set to 1 and you held control while dragging an object. Now it shouldn't do that.

### Caveats
None that I know

### Notes
I misnamed the commit and don't feel like amending it now that it's pushed


colinator27 (approved):
Looks good, thanks!

PR #2270, opened 2025-11-09
Add multiline regex option for code searching.

### LINKED ISSUE #2144: Add "Multiline" option for code searching.

UI/UX

When you search for code (either "Find in code" or using `Ctrl+F` in the code editor), there is no option to make regular expression multiline (so `^` and `$` matching line start and end).
This is supported by C# - by adding the `RegexOptions.Multiline` flag to the `Regex` constructor.

zivmaor:
Implements #2144. Adds new "Multiline regex" checkbox to Search in code UI. When the box is checked while doing a regex search, the search is treated as a multiline regular expression (^ and $ match line start and end instead of string start and end). The option is only visible when doing a regex search.

## Description

While the Regex search box is unticked, the UI remains the same as before this change:
image
When Regex search is selected, the new Multiline regex box appears. It's default value is false, which was the behaviour of the regex search before the change:
image

As an example I search for the third line of "gml_Script_attention_hackerz_no_2" from Undertale to show that it only appears while using a multiline regex:
image
image
### Caveats

There should be no caveats with the changes in this PR. I have tested multiline regex and it works correctly with and without Case sensitive. Any user who doesn't use regex search is completely uneffected, and those who do use regex search are uneffected unless they tick the new box.
### Notes

Issue #2144 Requested multiline regexes in both Search in code (`Ctrl+Shift+F`) and the code editor (`Ctrl+F`), but this PR only adds them to Search in code. I looked at how searching in the code editor works and it is part of avalonEdit and not UTMT. It is already in multiline mode (see https://github.com/icsharpcode/AvalonEdit/blob/862415d51eddc9eac93f462dbc522ffbf929cd52/ICSharpCode.AvalonEdit/Search/SearchStrategyFactory.cs#L37) with no easy way to turn off multiline mode.
Closes #2144.

### DIFF
--- a/UndertaleModCli/Program.cs
+++ b/UndertaleModCli/Program.cs
@@ -4,10 +4,6 @@ using Newtonsoft.Json;
 using System;
 using System.Collections.Generic;
 using System.CommandLine;
-using System.CommandLine.Builder;
-using System.CommandLine.Invocation;
-using System.CommandLine.NamingConventionBinder;
-using System.CommandLine.Parsing;
 using System.IO;
 using System.Linq;
 using System.Reflection;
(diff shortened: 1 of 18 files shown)


colinator27 (approved):
Looks good to me, thanks!

PR #2277, opened 2025-11-23
Add remaining GM 2024.14 format changes

Dobby233Liu:
Completely changes how vector shapes are stored in UndertaleSprites

### Caveats
* No proper version detection according to the new features is present. Currently there is a hack to assume everything is built with GM 2024.14.1
* No idea how to deal with particle system instance IDs


colinator27:
Made a bunch of fixes (and resolved TODOs/FIXMEs) and tested on what I could so far, I think this is ready to merge now!

PR #2281, opened 2025-11-29
Add external define constants for call and argument types and add FunctionArguments for external_define

Genouka:
## Description

Make codes to show correctly like this:
image


CST1229:
i thought external_define did not exist in GM:Studio and up (and utmt does not support GM8)?

PR #2288, opened 2025-12-13
Label individual size, margin, and origin fields

Jacky720:
## Description
Minor change to the sprite editor to clarify the width, height, and each margin side.
image

### Caveats
Looks worse than the current solution at extremely wide (too much empty space) and extremely narrow (cropped text, too-small fields) tool scales.
image
image

### Notes
Suggested by HalfBreadChaos in the Underminers server.


colinator27:
Looks good to me. I think it's probably acceptable for the worse look when it's shrunk horizontally, because you can still tell what it's supposed to mean from just a few letters. Not sure if there's a way to get breakpoints set up properly...

Besides that, I wonder if we also want to rename "Margin" to something like "BBox" (for bounding box), since that's the terminology GameMaker itself uses.

Issue #2286, opened 2025-12-12
`Resource Importers/ImportSounds.csx` InvalidOperationException (WPF)


BioTomateDE:
### Describe the bug

The UndertaleModTool GUI crashes with a InvalidOperationException when trying to use the ImportSounds script: "Added item does not appear at given index '304'.". This error occurs right after the script prints its success message.
Stacktrace (image): https://media.discordapp.net/attachments/586954521880690690/1449141796360687786/image.png

**Original conversation:**
Link: https://discord.com/channels/566861759210586112/586954521880690690/1449141796620992757
Guild: Underminers
Channel: #help
Date: 2025-12-12 20:52 UTC

### Reproducing steps

1. Open Undertale in UndertaleModTool GUI
2. Use ImportSounds.csx
3. Select folder with WAV files
4. This script imports in bulk. Are you sure? > Yes
5. Replace sound properties? > No
6. Specify properties automatically? > Yes
7. Import OGG internal > No
8. If a sound exists it will be replaced > Ok
9. "Sounds successfully added"
10. Program successfully crashes

### Setup Details

Tool Version: 0.8.2.0
OS: probably windows 10 or 11

idk, ask halfbreadchaos. im just the janitor.

BioTomateDE:
the culprit might be line 179:
```cs
soundData = new UndertaleEmbeddedAudio() { Data = File.ReadAllBytes(file) };
Data.EmbeddedAudio.Add(soundData);
if (existingSound is not null)
{
    Data.EmbeddedAudio.Remove(existingSound.AudioFile); // HERE
}
```

idk how remove is handled, but if this shifts audio ids, its gonna be fucked.
my fix™ (without testing) would be to just replace the EmbeddedAudio sounddata instead adding then removing

PR #2292, opened 2025-12-20
Reworked ImportSingleSound.csx

BioTomateDE:
## Description
Refactored code of `UndertaleModTool/Scripts/Resource Importers/ImportSingleSound.csx`.
The semantics should remain the same, except for one (potential) bugfix regarding #2286.

### Caveats
I can't get the stupid WPF app to work with Wine so I couldn't test my changes (sorry).
***please test before merging***

### Notes
if LangVersion is upgraded to 11, I can use `"""` for multiline strings (cleaner probably)

### DIFF
--- a/.github/workflows/pr_artifacts.yml
+++ b/.github/workflows/pr_artifacts.yml
@@ -13,17 +13,17 @@ jobs:
     if: github.event.workflow_run.event == 'pull_request' && github.event.workflow_run.conclusion == 'success'
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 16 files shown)


colinator27:
I'm not sure the bugfix regarding #2286 actually fixes what it's trying to... It does get rid of old embedded audio, though, which is maybe something desirable? Maybe not? There were just some other changes merged for `ImportSounds.csx` itself (still see no reason why they would affect #2286 though), which probably should get updated similarly if this script gets updated, behavior-wise.

PR #2315, opened 2026-01-31
Add hint message for zooming in the code editor

### LINKED ISSUE #2224: Add a hint about Ctrl + mouse button feature in the code editor.

UI/UX

There should be a hint (probably at the bottom of the code editor tab) that you can use Ctrl + mouse buttons (or wheel mouse button) to interact with code objects like functions and numbers (asset references) in the code editor.

Cheesecake2960:
## Description
Resolves #2224
I added a hint at the bottom of the code editor indicating that you can zoom using Ctrl and + or - or the mouse wheel.


colinator27:
Hmm... considering these are very common keybinds across text editors in general, I'm not too sure this is something that should be so prominent and take up vertical real estate. That said, I do think we should document them *somewhere*, along with other things like the super-obscure Ctrl+K keybind to recompile code, and other things along those lines.

PR #2321, opened 2026-02-10
Improve BuiltinList.cs

BioTomateDE:
`UndertaleModLib/Compiler/BuiltinList.cs` was a mess, so I fixed it up.

## Variable Definition
Every variable definition has its name written twice.
Example: `GlobalNotArray["current_second"] = new VariableInfo("current_second", true, false);`
(Name is needed in the dict key and in the VariableInfo.)

I introduced some helper methods to create variables instead.
In this example: `DefineGlobal("current_second", true);`

More generally:
```cs
private void DefineGlobal(string name, bool canSet)
{
    Globals[name] = new VariableInfo(name, canSet);
}

private void DefineGlobalArray(string name, bool canSet)
{
    GlobalArrays[name] = new VariableInfo(name, canSet);
}

private void DefineGlobalAutoArray(string name, bool canSet)
{
    GlobalArrays[name] = new VariableInfo(name, canSet, true, true);
}

private void DefineInstanceVar(string name, bool canSet)
{
    InstanceVars[name] = new VariableInfo(name, canSet, false);
}
```

## Versioning
Lots of if statements like this: `if (data?.GeneralInfo?.Major >= 2)`.
I instead defined a `gms2` (and `gms2_3`) variable to use.
(This is not really a big deal, but I changed it while I was at it anyway.)

I just really hate redundancy so i made this to reduce verbosity later
```cs
var gen8 = data?.GeneralInfo;
uint major = gen8?.Major ?? 0;
uint minor = gen8?.Minor ?? 0;
uint release = gen8?.Release ?? 0;
uint build = gen8?.Build ?? 0;

bool gms2 = major >= 2;
bool gms2_3 = major > 2 || (major == 2 && minor >= 3);
```

## Small bugfixes
There was a version check like this:
`if (data?.GeneralInfo?.Major >= 2 || data?.GeneralInfo?.Minor >= 3)`.
This *might* work but is probably not intended. I assume it meant to check `2.3+`.
The comment on that line said `since 1.3` (which probably was a typo for 2.3)
i changed it to `gms2_3` now.

## Renaming (breaking API!)

The name `GlobalNotArray` is kind of stupid in my opinion, so I changed it to just `Globals`.
I also changed `GlobalArray` to `GlobalArrays` and `Instance` to `InstanceVars`.

Since these fields are public, they might break dependents. I can revert these renames if you want.


UndertaleModLib/Compiler/BuiltinList.cs
```diff
@@ -2223,7 +2304,10 @@ private void InitializeMain(UndertaleData data)
         DefineFunction("buffer_async_group_begin", 1);
         DefineFunction("buffer_async_group_end", 0);
         DefineFunction("buffer_async_group_option", 2);
-        DefineFunction("buffer_get_surface", ((data?.IsVersionAtLeast(2, 3, 1) ?? false) ? 3 : 5)); // be more robust here
+        if (gms2_3 && release >= 1)
```
colinator27: I'm pretty sure the check for `release` here is bugged, since it doesn't account for when the major version is > 2, or the minor version is > 3 (although the latter never happens).

UndertaleModLib/Compiler/BuiltinList.cs
```diff
@@ -55,14 +86,14 @@ public FunctionInfo(string name, int numArguments, FunctionClassification classi
...
 public class BuiltinList : IBuiltins
 {
+    public Dictionary<string, FunctionInfo> Functions { get; private set; } = null;
     public Dictionary<string, double> Constants { get; private set; } = null;
-    public Dictionary<string, VariableInfo> GlobalNotArray { get; private set; } = null;
-    public Dictionary<string, VariableInfo> GlobalArray { get; private set; } = null;
-    public Dictionary<string, VariableInfo> Instance { get; private set; } = null;
+    public Dictionary<string, VariableInfo> Globals { get; private set; } = null;
+    public Dictionary<string, VariableInfo> GlobalArrays { get; private set; } = null;
+    public Dictionary<string, VariableInfo> InstanceVars { get; private set; } = null;
```
colinator27, line 93: I'm a little curious why only this variable had "Vars" appended to the end of it... since some of the surrounding dictionaries are also variables. The naming is a bit strange either way, but I feel like we should probably be consistent.

UndertaleModLib/Compiler/BuiltinList.cs
```diff
@@ -73,6 +104,49 @@ public BuiltinList(UndertaleData data)
...
+        {
+            string name = function?.Name?.Content;
+            if (name is null)
+                continue;
+            if (name.StartsWith("gml_Script_", StringComparison.Ordinal))
+                continue;
+            if (scriptLookup.Contains(name))
+                continue;
+            if (Functions.ContainsKey(name))
+            {
```
colinator27: Small thing, but I guess we should probably lose the `{` here.

UndertaleModLib/Compiler/BuiltinList.cs
```diff
@@ -206,7 +288,7 @@ private void InitializeMain(UndertaleData data)
         DefineFunction("matrix_stack_clear", 0);
         DefineFunction("matrix_stack_top", 0);
         DefineFunction("matrix_stack_is_empty", 0);
-        if (data?.GeneralInfo?.Major < 2)
+        if (gms2)
```
colinator27: This check is inverted.

UndertaleModLib/Compiler/BuiltinList.cs
```diff
@@ -907,7 +989,7 @@ private void InitializeMain(UndertaleData data)
             DefineFunction("texturegroup_get_fonts", 1);
             DefineFunction("texturegroup_get_tilesets", 1);
         }
-        if (data?.GeneralInfo?.Major >= 2 || data?.GeneralInfo?.Minor >= 3) // Since 1.3?
+        if (gms2_3)
```
colinator27: This check should be reverted to check for 1.3 and above, it was actually correct before based on some quick research.

UndertaleModLib/Compiler/BuiltinList.cs
```diff
@@ -3293,10 +3377,9 @@ private void InitializeMain(UndertaleData data)
         Constants["c_white"] = 16777215.0;
         Constants["c_yellow"] = 65535.0;
         Constants["c_orange"] = 4235519.0;
-        if (data?.GeneralInfo?.Major < 2)
-        {
+        if (!gms2)
             Constants["bm_complex"] = -1.0;
-        }
+
```
colinator27: Really tiny thing, but I guess we probably don't want an empty line here, or just keep the braces for consistency. (Same for a few others of these.)

UndertaleModLib/Compiler/BuiltinList.cs
```diff
@@ -3758,12 +3843,9 @@ private void InitializeMain(UndertaleData data)
         Constants["asset_sprite"] = 1.0;
         Constants["asset_sound"] = 2.0;
         Constants["asset_room"] = 3.0;
-        if (data?.GeneralInfo?.Major < 2)
+        if (!gms2)
         {
             Constants["asset_background"] = 4.0;
-        }
-        else
```
colinator27, line 3765: This `else` needs to be re-added. Backgrounds were renamed to tilesets in GMS2.

colinator27 (changes requested):
Looks mostly good, aside from a few errors I noticed so far. I'll do a more thorough review once those are fixed. Thanks!

UndertaleModLib/Compiler/BuiltinList.cs
BioTomateDE: i like conciseness but thought `Instances` would be a bit weird.
but you're right, it's better to make it consistent.
should i rename `InstanceVars` to `Instances` or rename all others to `*Vars`?

UndertaleModLib/Compiler/BuiltinList.cs
BioTomateDE: oh shit sorry lol

UndertaleModLib/Compiler/BuiltinList.cs
BioTomateDE: ah ok, i thought this was a typo since 2.3 is common and i've never seen 1.3 or similar

UndertaleModLib/Compiler/BuiltinList.cs
colinator27: I guess it's your choice, really. Though I suppose adding 
`Vars` to the end of all of them would make the most sense for anyone reading the code in the future, and I guess there won't be much noise created by doing that.

BioTomateDE: alright theyre called `*Vars` now :+1:

colinator27 (approved):
Did some minor additional cleanup and made some corrections (for some reason, two variables were configured incorrectly after the refactor, probably a regex error - caught it through my own double-checking). I think this is good now, thanks!

PR #2322, opened 2026-02-13
Cleanup enums, rename variables, and improve class documentation

setupwitch:
## Description
- Change UndertaleBackground.GMS2UnknownAlways2 to UndertaleBackground.GMS2TilesetVersion
- Document a bit of UndertaleParticleSystem and add the DrawOrderEnum enum
- Delete UndertaleExtensionKind.Unknown0, its never used as far as I know.
- Add all values for Mode in AudioKeyframe

### Caveats
All of these changes are untested, though I don't believe there will be any issues.

### Notes
I know that GMS2UnknownAlways2 is the version due to this snippet of code in the compiler.
```
      bool flag = false;
      if (Program.MajorVersion >= 2)
        flag = true;
      if (!flag)
        return;
      __s.WriteInteger(2);
```


colinator27 (approved):
Looks good! Did some additional cleanup, and I think this is good to merge now.

PR #2335, opened 2026-03-13
use Extent instead of Composite for padding

setupwitch:
## Description
An optimization to TextureWorker which improves the speed of adding padding to images

### Caveats
This is not intensively tested, though from my (minimal) testing, it looks alright.

### Notes
Everywhere I look, Extent is said to be faster.
https://stackoverflow.com/questions/54270807/how-to-add-padding-to-an-image-using-magickimage-in-net-core

### DIFF
--- a/UndertaleModLib/Util/TextureWorker.cs
+++ b/UndertaleModLib/Util/TextureWorker.cs
@@ -95,6 +95,7 @@ namespace UndertaleModLib.Util
             if (includePadding)
             {
                 // Based on a benchmark, Extent is faster than creating a new image and using Composite
+                image.Compose = CompositeOperator.Copy;
                 image.Extent(new MagickGeometry(-texPageItem.TargetX, -texPageItem.TargetY, (uint)exportWidth, (uint)exportHeight), MagickColors.Transparent);
             }
 


colinator27:
Aside from a few minor style things which we can fix up pretty quick, I am a bit concerned about regressions with this kind of thing. Particularly, does anything bad happen if the target X/Y are out of bounds? Do images look exactly identical? Probably will need to confirm those.

As for performance, it probably is faster to use one API over two separate ones (in general), although it would be nice to get data on it I guess.

setupwitch:
Just ran a benchmark between main branch and this PR, Extent is ~10000ms faster on Pizza Tower, and using a tool like odiff showed no differences in either image.

composite:
```
Time passed: 68455 ms
Peak Memory Usage: 5397.63 MB
```

extent:
```
Time passed: 58435 ms
Peak Memory Usage: 5395.11 MB
```

Untitled

colinator27:
After some testing, it seems that Extent has a behavioral difference after all: it loses RGB values on pixel that are fully-transparent. Not sure if there's any way to remedy this.

colinator27:
Actually, that was a quick fix! Seems to be good now.

PR #2336, opened 2026-03-15
Updated README.md

YarTom:
Added "Quick start" section, and information about releases to different platforms has been fixed and updated. I also added my own repo with auto-installer for Linux and macOS.


colinator27 (approved):
Simplified things a bit to condense things down, and I think it's good now. Thanks!

If you need any changes to your link in the future, feel free to PR again.

PR #2339, opened 2026-03-23
Update Magick.NET package version to 14.10.3

y114git:
## Description
Updated Magick.NET-Q8-AnyCPU from 14.10.2 to 14.10.3.
This update addresses a lot of high-severity bugs.
### Caveats
No breaking changes to the public API were identified in this minor version bump, so backward compatibility is maintained.

### Notes
image

There just were 300+ warnings when i tried to build UndertaleModLib
All unit tests passed successfully on the local environment after the update.


colinator27:
Ended up using https://github.com/UnderminersTeam/UndertaleModTool/pull/2337 instead, there seem to be no breaking changes in the API in general, aside from additions.

PR #2340, opened 2026-03-25
Refactor ExportAllSprites.csx and fix memory usage

setupwitch:
Refactor sprite export process to use concurrent collections and improve performance with parallel processing.

## Description
A refactoring of ExportAllSprites.csx that exports sprites texturepage by texturepage instead of sprite by sprite.

### Caveats
The progress bar now displays the amount of texture pages left to dump instead of sprites.

### Notes
I ran a small benchmark, while on my  here are the results on Pizza Tower

This PR:
```
Time passed: 59003 ms
Peak Memory Usage: 1120.95 MB
```

PR #2335:
```
Time passed: 58435 ms
Peak Memory Usage: 5395.11 MB
```

Slightly slower, but I believe the memory usage speaks for itself.


colinator27 (approved):
While computers with more cores are also more likely to have more RAM, I do wonder if there should be an additional hard limit on the parallelism to bound the memory usage as well. But seeing as this seems to be an improvement either way, I think this is good to merge, thanks!

PR #2355, opened 2026-04-28
Add project build option

luizzeroxis:
## Description
Add command to build a project file, so you can do it without the GUI.

`UndertaleModCli project build "project/project.json" -s "source/data.win" -d "destination/data.win"`

### Caveats
I'm not sure if you actually need to call `RecompileAllCodeSources` here, since everything should already be compiled when importing. Also not sure if the options need to be required always.

### Notes
I made it a subcommand so other options could be added there later, such as `new`.


colinator27 (approved):
Looks good to me, thanks!

PR #2376, opened 2026-05-16
Remove empty object events to prevent segfault

BioTomateHP:
## Description
It is possible to create a new event in the UndertaleModTool GUI that has no actions.
Events with zero actions will lead to a segfault in the runner when loading chunk OBJT, because it assumes there is at least one action per event.

I encountered this bug on Undertale 1.01 Linux, maybe modern runners are not affected.

How to reproduce:
* Open Undertale in UndertaleModTool
* Go to any game object
* Create a new event but do not add any actions (code entries) to it.
* Save data file
* Run game
* `Process Chunk: OBJT   730320`
* `fish: Job 1, '~/.sg/Undertale/runner' terminated by signal SIGSEGV (Address boundary error)`
* Unprofit

I added a small method to remove those empty events; fixing this bug.

### Caveats
This code removes the empty events when building the data file.
If it's not desired to change anything about the UndertaleData while serializing, then this new methods needs to be called _before_ writing.

### DIFF
--- a/UndertaleModLib/Models/UndertaleGameObject.cs
+++ b/UndertaleModLib/Models/UndertaleGameObject.cs
@@ -209,6 +209,7 @@ public class UndertaleGameObject : UndertaleNamedResource, IProjectAsset, INotif
         {
             v.Serialize(writer);
         }
+        FixEvents();
         writer.WriteUndertaleObject(Events);
     }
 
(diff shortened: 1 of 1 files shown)


colinator27:
This is definitely a nice thing to have, although part of me is wondering if it should be a save error instead, somehow... I think there's a bunch of places where things can be null, e.g. due to scripts malfunctioning, and those cases may not be as simple to automatically repair when saving. Though I did recently make a 2024.13+ change which creates an empty room instance list if it happens to be null, when saving... Not too sure. If there *is* an error, I guess it would probably need to be obvious to fix, but not sure how often this situation comes up.

BioTomateDE:
> I think there's a bunch of places where things can be null

If an element is null, then it writes 4 null bytes, which happens to be equivalent to an empty list.
With this new code, it *would* actually throw an error if the event group / event pointer list is null, unlike before (where it would essentially write an empty list). That's what you wanted, right?

I think this "empty event" situation mostly occurs when a UTMT GUI user either:
* accidentally creates a new event and doesn't add any actions to it
* removes all actions from an event without removing the event itself

Right now, this will serialize without any error or warning but lead to an immediate crash when loading the game, which is very frustrating. Most users probably forgot that they even had an empty event somewhere or didn't expect it to cause problems, so, in my opinion, this *should* get automatically removed, rather than letting it pass or throwing some cryptic error message.

(Afaik the writer doesn't have a warning handler, so throwing a warning here is out of question anyway...)

so yes, i also like explicitness instead of silent weird behavior, but in the "severe" cases you mentioned, with a malformed script output, it already throws a NRE lol

colinator27 (approved):
I think I'm okay with how it is now, so I'll merge it as-is. Thanks!

(I also tested this on a YYC game just to make sure that it still worked in that case. I believe YYC generates actions with null code references, which shouldn't be touched by this.)

PR #2380, opened 2026-05-20
CLI: Support dumping sprites

ephoton:
## Description
Adds a CLI option to dump all sprites

### Caveats
N/A

### Notes
I'll make later PRs for dumping other parts

### DIFF
--- a/UndertaleModCli/Program.cs
+++ b/UndertaleModCli/Program.cs
@@ -2,6 +2,7 @@
 using Microsoft.CodeAnalysis.Scripting;
 using Newtonsoft.Json;
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.CommandLine;
 using System.IO;
@@ -20,6 +21,15 @@ using UndertaleModLib.Util;
 using static UndertaleModLib.UndertaleReader;
 
 namespace UndertaleModCli;
+
+class TextureToExport
+{
+    public UndertaleTexturePageItem PageItem { get; set; }
+    public UndertaleEmbeddedTexture Page => PageItem.TexturePage;
+    public string FileExportLocation { get; set; }
+
+    public TextureToExport(UndertaleTexturePageItem pageItem, string fileExportLocation) => (PageItem, FileExportLocation) = (pageItem, fileExportLocation);
+}
 
 /// <summary>
 /// Main CLI Program
@@ -518,7 +528,9 @@ public partial class Program : IScriptInterface
             program.DumpAllTextures();
 
         if (options.Sprites)
-            program.DumpAllSprites();
+        {
+            int _wait = program.DumpAllSprites().Result;
+        }
 
         return EXIT_SUCCESS;
     }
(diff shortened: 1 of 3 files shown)


UndertaleModCli/Program.cs
```diff
@@ -888,6 +894,37 @@ private void DumpAllTextures()
...
+        foreach (UndertaleSprite sprite in Data.Sprites)
+        {
+            if (Verbose)
+                Console.WriteLine($"Dumping {sprite.Name}");
+
+            for (int i = 0; i < sprite.Textures.Count; i++)
+            {
+                if (sprite.Textures[i]?.Texture is not null)
+                {
+                    string dest = Paths.JoinVerifyWithinDirectory(directory, $"{sprite.Name.Content}_{i}.png");
```
colinator27: I *think* this is supposed to be TryJoinVerifyWithinDirectory, because currently this throws upon failure.

UndertaleModCli/Program.cs
```diff
@@ -888,6 +894,37 @@ private void DumpAllTextures()
...
+                    if (dest is null)
+                    {
+                        Console.Error.WriteLine($"Failed to export sprite with name {sprite.Name.Content}");
+                        return;
+                    }
+                    using FileStream fs = new(dest, FileMode.Create);
+                    sprite.Textures[i]?.Texture.TexturePage.TextureData.Image.SavePng(fs);
```
colinator27: I think this is currently exporting the entire texture page that each sprite frame belongs to... You could take a look at what `ExportAllSprites.csx` does for this, using `TextureWorker`. Similarly, we could copy some of the parallelism potentially... but probably not a huge concern for now.

UndertaleModCli/Program.cs
```diff
@@ -888,6 +894,37 @@ private void DumpAllTextures()
...
+    /// </summary>
+    private void DumpAllSprites()
+    {
+        string directory = Path.Join(Output.FullName, "Sprites");
+
+        Directory.CreateDirectory(directory);
+
+        foreach (UndertaleSprite sprite in Data.Sprites)
+        {
+            if (Verbose)
```
colinator27: I think we need a null check on `sprite` here, which can happen with unused asset removal.

colinator27 (changes requested):
Overall structure so far seems good, will just need to fix some of the details I think. Thanks!

UndertaleModCli/Program.cs
ephoton: sure
the better thing to do in general is to allow the CLI to run scripts too

ephoton: with some noop polyfill for all the gui stuff

colinator27: > sure the better thing to do in general is to allow the CLI to run scripts too

It can already do that...

ephoton:
I left a placeholder for showing progress, I think it is probably a good idea to increment an atomic int and show progress in CLI output like `Fetching X/Y` then `Exporting X/Y`

UndertaleModCli/Program.cs
ephoton: with gui polyfill?

colinator27 (approved):
Made some small tweaks, tested and it seems to work great! Thanks!

PR #2387, opened 2026-05-31
Update Resource Exporters scripts

YarTom:
Replace "ExportAll..." and "ExportSpecific..." scripts with unified export scripts that let the user filter resources via a new options dialog.

img


colinator27:
I really like this idea, although I think we could possibly take this the final mile and convert most of the resource import/export scripts to be built-in to the library itself. Not sure how much that conflicts with this design just yet... (will need to take an in-depth look)

Although, if it can be naturally extended from the new design in this PR, I think it could still be worthwhile to merge this as-is. Curious for thoughts!

PR #2389, opened 2026-06-05
Optimize UnescapeText method using StringBuilder

y114git:
Refactor UnescapeText method to use StringBuilder for optimization.

## Description
This refactors `UndertaleString.UnescapeText` to parse escape sequences in a single pass with `StringBuilder` instead of applying multiple chained `string.Replace` calls.

The previous implementation could re-process characters produced by an earlier replacement. For example, an escaped literal backslash followed by `n` (`"\\n"` in assembler text) could be transformed into a real newline because `Replace("\\n", "\n")` ran before `Replace("\\\\", "\\")`.

The new implementation handles only the escape sequences emitted by `UndertaleString.ToString(true)`:

- `\r` -> carriage return
- `\n` -> newline
- `\"` -> double quote
- `\\` -> literal backslash

Unknown escape-like sequences are preserved as-is instead of dropping the backslash. This keeps strings such as Windows paths (`C:\Users\Name`) or unsupported escape-like text (`\t`, `\x`, `\0`, `\U`) from being silently changed.

## Caveats
This intentionally does not add support for additional GML/C-style escape sequences such as `\t`, `\xNN`, octal escapes, or Unicode escapes. The method is used by the VM assembler text path and is kept as the inverse of `UndertaleString.ToString(true)`, rather than a full GML string literal parser.

## Notes
Needs tests ig

### DIFF
--- a/UndertaleModLib/Models/UndertaleString.cs
+++ b/UndertaleModLib/Models/UndertaleString.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 
 namespace UndertaleModLib.Models;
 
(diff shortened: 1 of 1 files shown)


colinator27 (approved):
Looks good to me, thanks! (Made some very tiny refactors.)

PR #2398, opened 2026-06-20
perf(scripts): improve sound export throughput

cmdr-chara:
## Summary

* Improves `ExportAllSounds.csx` throughput by caching audio group lookups, avoiding repeated directory creation, and writing/copying export jobs with bounded parallelism.
* Preserves output filenames and contents for the tested Undertale export case.

## Testing

* Compared upstream script vs this change on `UNDERTALE/data.win`.
* Output matched: 443 files, 128,007,271 bytes, 0 missing/extra/changed files by SHA-256.
* 3-run median: 4412 ms -> 4082 ms (~7.5% faster).

## Risks

* Tested with Undertale `data.win`; other GameMaker titles may exercise different external audio layouts.


cmdr-chara:
@colinator27 Could you please check the code? This script was improved in my WinUI fork, but I'd like for it to be included here as well.

Issue #2399, opened 2026-06-21
failed to run a data.win file


Goldetm:
### Describe the bug

when i try to open a data.win file from a game called monster girl assult and it crashed like this 

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/0c941c2c-4f92-4df7-a4af-aed592d1df40" />

### Reproducing steps

1 i open the app and open the file and it just show error

### Setup Details

i use the newest version

setupwitch:
I dont seem to be able to replicate it, have you extracted the undertalemodtool zip file?

Dobby233Liu:
> I dont seem to be able to replicate it

Which version of the game did you test? OP is apparently trying to open v0.3.6.1a

> have you extracted the undertalemodtool zip file?

Well you can see all the files in the background...

PR #2400, opened 2026-06-21
Hack to make Monster Girl Assault [NSFW] Spine sprites load

Dobby233Liu:
## Description
As the title says. Per #2399, and to be clear I'm only doing this for the bug report

### Caveats
Idk what version exactly brought the change, or whether the 2023.1 check is just broken, so no version gate for now. Also I still don't know how to write version detection code.

### DIFF
--- a/UndertaleModLib/Models/UndertaleSprite.cs
+++ b/UndertaleModLib/Models/UndertaleSprite.cs
@@ -428,7 +428,7 @@ public class UndertaleSprite : UndertaleNamedResource, IProjectAsset, PrePaddedO
                     byte[] encodedJson = EncodeSpineBlob(Encoding.UTF8.GetBytes(SpineJSON));
                     byte[] encodedAtlas = EncodeSpineBlob(Encoding.UTF8.GetBytes(SpineAtlas));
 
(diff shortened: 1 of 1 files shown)


Dobby233Liu:
I can confirm however that UndertaleSpineTextureEntry is indeed written like that in 2026.0.

Goldetm:
so is there any way to fix this

Dobby233Liu:
> so is there any way to fix this

This is not a high-quality fix but should make your file load

Goldetm:
so how can i use it

Dobby233Liu:
> so how can i use it

Get the "GUI-windows-latest-isBundled-false-isSingleFile-false" file from the bot reply and install it like normal UMT. Note that with this fix (in the state it is in for now) applied you'll not be able to open some other games

Dobby233Liu:
Colin thinks this is ok to be out of draft

colinator27:
Yeah, things line up very consistently with the bug just being a LTS version branch problem, and this PR works well with every game I've tested myself. If things break somehow, we can always revert this, though.

colinator27 (approved):

PR #2401, opened 2026-06-21
Port over DogScepter's collision mask detection

setupwitch:
## Description
Implements DogScepter's collision mask detection for diamonds and ellipses into the library.

### Caveats
Detection is on a per-mask basis, false positives are almost definite under certain conditions.

### Notes
Since each mask is being handled separately, the IDE's "Precise Per Frame" option is unnecessary.

### DIFF
--- a/UndertaleModLib/Models/UndertaleSprite.cs
+++ b/UndertaleModLib/Models/UndertaleSprite.cs
@@ -81,8 +81,7 @@ public class UndertaleSprite : UndertaleNamedResource, IProjectAsset, PrePaddedO
     /// The separation mask type this sprite has.
     /// </summary>
     public SepMaskType SepMasks { get; set; }
-
-
+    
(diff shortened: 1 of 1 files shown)


UndertaleModLib/Models/UndertaleSprite.cs
```diff
@@ -340,7 +467,11 @@ public void Dispose()
     public class MaskEntry : IDisposable
     {
         public byte[] Data { get; set; }
-
+        
+        /// <summary>
+        /// The type of collision mask used, UTMT only.
+        /// </summary>
+        public ColMaskType? AssumedColMaskType { get; set; }
```
colinator27, line 474: I'm wondering if this is the way that we want people to get this information... maybe the method that calculates the types should just return it as an array or list instead? Not sure.

PR #2403, opened 2026-06-22
Panicking ByName/IndexOf/IndexByName, separate Try- methods for null/-1

BioTomateDE:
The UndertaleData methods and extensions `ByName()`, `IndexOf` and `IndexOfByName` (now called `IndexByName`) now throw an exception when the element could not be found. The old behavior of returning `null`/`-1` is still available via `TryByName`, `TryIndexOf` and `TryIndexByName`.

### Caveats
This is a breaking change.

### Notes
i fixed all the csx scripts (hopefully)

### DIFF
--- a/UndertaleModCli/Program.UMTLibInherited.cs
+++ b/UndertaleModCli/Program.UMTLibInherited.cs
@@ -495,7 +495,7 @@ public partial class Program : IScriptInterface
     }
 
     /// <inheritdoc/>
-    public string GetDisassemblyText(string codeName  GetDisassemblyText(Data.Code.ByName(codeName));
+    public string GetDisassemblyText(string codeName) => GetDisassemblyText(Data.Code.ByName(codeName));
 
     /// <inheritdoc/>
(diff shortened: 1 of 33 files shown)


colinator27:
Thinking about wrapping up the design for this, and honestly I'm super hesitant about breaking so many non-builtin scripts that rely on the existing methods...

BioTomateDE:
@colinator27 i understand. i still think this would be useful to have in utml in some way tho. maybe the old methods keep their names and the new failable ones could be named `Force-` or something. or just keep the existing methods and make a `bool panic = false` argument. what do ~~YOU TWO~~ think?

colinator27:
> @colinator27 i understand. i still think this would be useful to have in utml in some way tho. maybe the old methods keep their names and the new failable ones could be named `Force-` or something. or just keep the existing methods and make a `bool panic = false` argument. what do ~YOU TWO~ think?

Yeah, that does seem reasonable. I think we would want to make new methods, but I feel like "force" isn't quite the right word... maybe something more like "require"? Not sure...

PR #2406, opened 2026-06-24
New IDEVersion struct in GeneralInfo for convenience

BioTomateDE:
## Description
Made a new struct called IDEVersion that contains major, minor, release, build and LTS branch. UndertaleGeneralInfo now stores this in a new field called Version. You can still use GeneralInfo.Major to access GeneralInfo.Version.Major etc using a properties. This ensures this change does not break existing code (and is more convenient anyway).

The reason I implemented this is because I think this could be useful when handling GameMaker versions without needing to create a whole UndertaleGeneralInfo or making your own struct/tuple for the 5 different version parts.

### Caveats
no

### Notes
if you don't like the name IDEVersion I can change it to something else lol

### DIFF
--- a/UndertaleModLib/Models/UndertaleGeneralInfo.cs
+++ b/UndertaleModLib/Models/UndertaleGeneralInfo.cs
@@ -182,7 +182,89 @@ public class UndertaleGeneralInfo : UndertaleObject, IDisposable
         Steam = 0x2000000000000000UL,
         UNUSED5 = 2310346608841064448UL,
         Shaders = 0x4000000000000000UL,
-        VertexBuffers = 9223372036854775808UL
+        VertexBuffers = 0x8000000000000000UL
+    }
+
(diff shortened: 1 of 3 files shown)


luizzeroxis:
Interesting, although you'd have to be careful that it doesn't break XAML bindings and such (Fody would probably be doing a lot the heavy lifting here). I think it would be better to decouple the version as in the data file and the detected version by UTMT. You might as well combine both the version number and the branch into a single object at that point.

BioTomateDE:
we can have an additional OriginalVersion field that denotes the read gen8 version. thats probably better right?

colinator27 (approved):
Did some refactoring and made some corrections (mostly to the comments), and I think this is good now. Thanks!

(Of note: I got rid of `UndertaleObject` on the version struct, because that adds unneeded overhead, even if it's minimal.)

PR #2409, opened 2026-06-28
Fix debug crashes in Chapter 5

break-core:
## Description
Patches some code in chapter 5 that causes crashes when trying to save the game in debug mode, or loading into the game from the debug start menu

### Caveats
As far as I know, there are no caveats. However, the data returned from `scr_get_room_list` will return 2 different room ids, and both will use `room_blank` as it's room.

### Notes
Maybe removing `rm_blank` from `scr_get_room_list` entirely instead of just changing it to a different existing room would be better? The solution I have works FWIW.


colinator27 (approved):
Simplified the code slightly, then tested, and it works great. Thanks!

PR #2411, opened 2026-06-29
Add function parameter definition for scr_84_add_menu_item

vrad-exe:
## Description
Adds a definition to the deltarune decompiler config to make `scr_84_add_menu_item` decompile the third parameter as a room asset when the second parameter is set to `[room]` or `[roomdark]`. This fixes the [obscure debug menu from the Chapter 1 Switch port](https://tcrf.net/Proto:Deltarune/Chapter_1_Demo_(2018)#8-4_Debug_Menu) decompiling with raw room IDs in its menu options.

### Caveats
They seem to have changed how the 8-4 menu works in chapter 3 onwards which would probably break this detection. This doesn't really matter unless a newer version of the menu manages to crop up in the future though.

### Notes
Before:
image
After:
image

The check for `[room]` or `[roomdark]` ensures that the other menu options that give items or the like are unchanged.


colinator27 (approved):
Looks fine to me, thanks!

PR #2417, opened 2026-07-08
Add arm64 build artifact for macOS

jjezewski:
## Description
Adds a macOS arm64 build artifact for the CLI.

### Caveats
The new arm64 artifact uses an additional `-arm64` suffix. I’m not sure if this is the preferred naming, but I tried to avoid breaking existing tooling that may depend on the current artifact names.

### Notes
I am open to making another commit with more standardized artifact names that point to specific architectures, such as `-x86`.

### DIFF
--- a/.github/workflows/publish_cli.yml
+++ b/.github/workflows/publish_cli.yml
@@ -19,6 +19,8 @@ jobs:
             rid: linux-x64
           - os: macOS-latest
             rid: osx-x64
+          - os: macOS-latest
+            rid: osx-arm64
           - os: windows-latest
             rid: win-x64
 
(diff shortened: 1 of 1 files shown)


colinator27:
Considering the possibility of making ARM64 the default instead, and giving `-x86` the suffix. I would expect not too many people (if any) have automated artifact-downloading workflows in the first place, so this should probably be fine? But if anyone has any objections, make them known.

jjezewski:
I swapped arm64 with x86. Also, I improved this workflow’s readability because it could be simplified.

colinator27 (approved):
Mirrored your changes in the other relevant workflows... going to give this a test and see what happens! (This may need to be reverted and attempted again...)

colinator27:
Amazingly, everything seems to be working great! Thanks!

PR #2428, opened 2026-07-14
More accurate GM/WAD Version ranges for BuiltinList functions

BioTomateDE:


UndertaleModLib/Compiler/BuiltinList.cs
```diff
@@ -1025,9 +1103,6 @@ private void InitializeMain(UndertaleData data)
             DefineFunction("draw_background_general", 15);
             DefineFunction("draw_background_tiled", 3);
             DefineFunction("draw_background_tiled_ext", 7);
-        }
-        if (!gms2)
-        {
```
colinator27, line 1030: In the cases of tiles and backgrounds, I'm wondering whether it's actually a good idea to keep the old checks, even if they're inconsistent with the runners themselves. Reason being, backgrounds were conceptually replaced with tileset assets in GMS2, and allowing these functions to compile regularly might cause confusion. (Do they even function properly in GMS2?)

UndertaleModLib/Compiler/BuiltinList.cs
```diff
@@ -1077,8 +1155,11 @@ private void InitializeMain(UndertaleData data)
         DefineFunction("surface_get_texture", 1);
         DefineFunction("surface_set_target", 1);
         DefineFunction("surface_set_target_ext", 2);
-        DefineFunction("surface_reset_target", 0);
-        DefineFunction("surface_depth_disable", 1);
+        if (gms2)
+        {
+            DefineFunction("surface_reset_target", 0);
```
colinator27: This seems wrong (for `surface_reset_target`). Pretty sure that function has been around since well before GMS2.

BioTomateDE: oh ur right, idk why i put those in a gms2 if statement, sorry

fixed in 63965ce3

UndertaleModLib/Compiler/BuiltinList.cs
BioTomateDE: github is showing the wrong lines for this comment i think. u mean this part right?
```c#
if (!gms2_3)
{
  DefineFunction("draw_background", 3);
  DefineFunction("draw_background_ext", 8);
  DefineFunction("draw_background_stretched", 5);
  // ...
}
```
idk how backgrounds/tilesets work lol, should i just revert this `!gms2_3` to `!gms2`?

colinator27: Yeah, I'm thinking of reverting it to `!gms2`, but also I can probably just do that myself, in a moment.

UndertaleModLib/Compiler/BuiltinList.cs
```diff
@@ -235,48 +235,66 @@ private void DefineInstanceVar(string name, bool canSet)
...
+            DefineFunction("@@Other@@", 0);
+            DefineFunction("@@NewGMLArray@@");
+        }
+        if (gms2_3)
+        {
+            DefineFunction("@@Global@@", 0);
+            DefineFunction("@@GetInstance@@", 1);
+            DefineFunction("@@NullObject@@", 0);
+            DefineFunction("@@NewGMLObject@@");
+            DefineFunction("@@SetStatic@@", 0);  // does this actually exist???
```
colinator27: I'll probably get rid of these comments myself in a moment, but yes the static functions in question here *do* exist. I believe `@@SetStatic@@` is used in modern constructor code gen, and `static_get` is a regular function that also happens to be used in code gen when dereferencing a static function by name (like `ExampleFunction.example_static_variable`)

UndertaleModLib/Compiler/BuiltinList.cs
```diff
@@ -3024,11 +3310,27 @@ private void InitializeMain(UndertaleData data)
             DefineFunction("dbg_same_line", 0);
             DefineFunction("dbg_add_font_glyphs");
         }
+        if (gm2022_1) {
+            DefineFunction("event_perform_async", 2);
```
colinator27: I found changelogs suggesting `event_perform_async` existed a bit before 2022.1, so I'm going to move that one back to where it was before, just in case.

colinator27 (approved):

UndertaleModLib/Compiler/BuiltinList.cs
BioTomateDE: yeah i got confused because that function didn't appear in any of my lists. but that was just because my latest tested version was 2022 lts which didn't have this function yet lol (i should've included a modern probe as well)

PR #2433, opened 2026-07-16
Unified Border Enabler script for Undertale and Deltarune

TerraFail451:
## Description
Merges the UndertaleBorderEnabler and UndertaleBorderEnablerv1_11 scripts, and adds support for the full and (most) demo versions of Deltarune.

## Notes
- Deltarune SURVEY_PROGRAM demo is not supported, as it lacks any border code

## TO-DO
- attempt to make Chapter 5 Flowery intro widescreen
- thoroughly test the script

## Screenshots (you can't really tell that they were taken on PC but whatever)

ezgif com-video-to-gif-converter (1)


setupwitch:
I think this would fit https://github.com/UnderminersTeam/awesome-utmt better

TerraFail451:
maybe? but UTMT already has Border Enabler scripts for Undertale already built in, might as well make them one built in script that's also compatible with Deltarune (like the also built in Debug script) right? i don't see why Undertale should get a border enabler script built in but not Deltarune

setupwitch:
My bad, I didn't know that the Undertale border sprites were already in the repository.

YarTom:
Isn't this the same thing that is made in NXRune mod? Here is the repository: https://github.com/IruzzArcana/NXRUNE.git . I tested it with friends about month ago, and we saw some bugs in chapter 5, but there have been a lot of fixes since that time

PR #2438, opened 2026-07-21
add Name.Regex option to GameSpecificResolver

InvoxiPlayGames:
## Description
This adds a `Name.Regex` condition to GameSpecificResolver to allow for GameSpecificData Underanalyzer fixes to be applied to games with a null or otherwise unhelpful DisplayName (e.g. certain console versions of DELTARUNE Chapter 4)

### Caveats
None...?


colinator27 (approved):
Unsure if this makes the existing checks redundant on already-supported platforms, but this seems fine to me. Thanks!

PR #2439, opened 2026-07-22
[CLI] Add Overwrite flag to LoadOptions.

IruzzArcana:
## Description
Adds force overwrite flag to the load command.
image


colinator27:
I wonder if this might break some people's workflows, but I guess the migration is super simple and explained clearly, so I'm not opposed to it. If anyone reading has any objections, make them known, I guess!

natroute:
this should probably apply to `project build` too

colinator27 (approved):
Seems fine to go forward with this for now, thanks!

PR #2454, opened 2026-08-09
Enable usage of custom IGlobalFunctions implementations in GlobalDecompileContext

skirlez:
## Description
This PR adds an optional parameter to GlobalDecompileContext of type IGlobalFunctions.

Previously there was no real way to utilize any other implementations of IGlobalFunctions in neither compilation nor decompilation.

This parameter has several useful applications, such as being able to disallow direct function calls to specific functions, or (my use case) making sure function calls from several code sources resolve correctly in an environment where each code source could have functions with the same name.

Additionally, this PR removes "Can be re-built by setting this to null" comment from UndertaleData's GlobalFunctions field. I believe this is a leftover from the old decompiler/compiler, checking older versions some scripts indeed had null checks in order to rebuild the cache, but today there are none left.

### Caveats

Previously the GlobalFunctions property was a getter pointing to the provided UndertaleData's GlobalFunctions. This means that with this implementation, if the provided UndertaleData's GlobalFunctions is set to a different list, which is possible, it won't be reflected within GlobalDecompileContext. I'm not sure if this should be accounted for. (It's only a concern for other projects utilizing UndertaleModLib, and not within the library itself. Within the library, the only cases where GlobalFunctions is set to a different list is inside GlobalDecompileContext itself, in BuildGlobalFunctionCache).

A way around this would be to add a private field for a user given IGlobalFunctions in GlobalDecompileContext, and to return the UndertaleData's GlobalFunctions as a "fallback" for if the user does not give any.


colinator27 (approved):
This seems good to me, thanks!

PR #2458, opened 2026-08-12
Add safeguards when creating temp data files

luizzeroxis:
## Description
Use FileMode.CreateNew to not overwrite existing temp file (checking if it was really created right now before deleting it) and flush the file to disk before moving it into the regular file path, which overwrites it. In theory this makes so that, in case of a system crash or power failure, at least one of those files (data.win or data.wintemp) will still be available on disk.

### Caveats
N/A

### Notes
N/A

### DIFF
--- a/UndertaleModTool/MainWindow.xaml.cs
+++ b/UndertaleModTool/MainWindow.xaml.cs
@@ -1233,15 +1233,20 @@ namespace UndertaleModTool
                 }
 
                 bool saveSucceeded = true;
+                bool tempFileCreated = false;
 
                 try
                 {
-                    using (var stream = new FileStream(filename + "temp", FileMode.Create, FileAccess.Write))
+                    using (var stream = new FileStream(filename + "temp", FileMode.CreateNew, FileAccess.Write))
                     {
+                        tempFileCreated = true;
                         UndertaleIO.Write(stream, Data, message =>
                         {
                             FileMessageEvent?.Invoke(message);
                         });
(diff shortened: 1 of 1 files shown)


BenjaminUrquhart (approved):
Can't test it right now but nothing stands out to me as obviously wrong so lgtm

UndertaleModTool/MainWindow.xaml.cs
```diff
@@ -1233,15 +1233,20 @@ private async Task<bool> SaveFile(string filename, bool suppressDebug = false)
                 }
 
                 bool saveSucceeded = true;
+                bool tempFileCreated = false;
 
                 try
                 {
-                    using (var stream = new FileStream(filename + "temp", FileMode.Create, FileAccess.Write))
+                    using (var stream = new FileStream(filename + "temp", FileMode.CreateNew, FileAccess.Write))
```
BenjaminUrquhart, line 1240: I wonder if there should be a special message in case you already have a temp file, something like "hey you already have a temp file, maybe something went wrong earlier and you should check it out"

There's also the case of having multiple instances open and trying to save simultaneously but I'd expect that to occur more often with cli scripts which this doesn't apply to.

luizzeroxis: Sure that should be easy to add. Shouldn't be a problem saving multiple if they go in different paths, not sure if they even use this method for that

PR #2466, opened 2026-08-19
New ExportFontsAsTTF.csx script

YarTom:
## Description
Added a new script that allows you to export raster GM fonts as TTF. It's useful for games that use monochrome pixel 
fonts, like Undertale. (I tested it in deltarune and undertale)

### Preview of the exported font in FontForge
Font preview

The script is pretty big, but a major part of the code is a TTF binary writer written from scratch.

The algorithm itself is actually quite simple: for every glyph, we calculate its outlines and write them to the TTF file.

## Notes

The script is quite large and currently supports only monochrome fonts, so maybe it should go into "UTDR Scripts" or "Technical Scripts".

Extra ideas I have:
- Import TTF fonts directly into the font editor. This should be much easier than exporting them, since ImageMagick supports drawing text on images.
- Connect this script with "UndertaleTTFFonts.csx" and "DeltaruneTTFFonts.csx". Right now, I don't even understand how these two scripts work. (Why is `font_add()` called without storing its return value?)
- Add support for colored fonts. This shouldn't be too hard, but I don't think there are many non-monochrome fonts.

### DIFF
--- a/UndertaleModTool/MainWindow.xaml.cs
+++ b/UndertaleModTool/MainWindow.xaml.cs
@@ -781,7 +781,8 @@ namespace UndertaleModTool
             {
                 DefaultExt = "win",
                 Filter = DataFileFilter,
-                FileName = FilePath
+                FileName = FilePath,
+                AddExtension = false
             };
             if (dlg.ShowDialog(this) == true)
             {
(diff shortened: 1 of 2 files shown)


setupwitch:
whats stopping you from making this work for all gamemaker games?

YarTom:
The script should work fine with all GM games, but it cannot transfer colors (all fonts will always be black and white after export). This shouldn’t be a big problem, since 99% of fonts are monochrome. 

In addition, fonts can only be transferred pixel by pixel (which makes sense, since we are converting glyphs from a raster image). This works perfectly with pixel fonts, but common fonts require some smoothing, which I find difficult to implement.

letter A

Issue #2319, opened 2026-02-08
"Find Unreferenced Assets" has many false positives


Crystal-Shard:
### Describe the bug

The game Valdis Story: Abyssal City can be opened in UndertaleModTool and appears to be fully compatible, except that the Find Unreferenced Assets menu option doesn't work.

With 'strings' deselected, it reports many false positives among unreferenced assets.
With 'strings' selected, a null pointer exception occurs.

### Reproducing steps

1. use Find Unreferenced Assets from the menu
2. select every type of asset in the menu, except for string
3. I get a list showing pretty much every asset in the game. Apparently UMT detects every asset as unreferenced.

Alternatively:
1. use Find Unreferenced Assets from the menu
2. select every type of asset in the menu, including string
3. I get the error "Value cannot be null. (Parameter 'source')" and a stack trace, shown below.
4. No results are displayed; EMT does not crash and continues functioning normally.

```
System.AggregateException: One or more errors occurred. (Value cannot be null. (Parameter 'source')) (Value cannot be null. (Parameter 'source')) (Value cannot be null. (Parameter 'source')) (Value cannot be null. (Parameter 'source'))
 ---> System.ArgumentNullException: Value cannot be null. (Parameter 'source')
   at System.Linq.ThrowHelper.ThrowArgumentNullException(ExceptionArgument argument)
   at System.Linq.Enumerable.Where[TSource](IEnumerable`1 source, Func`2 predicate)
   at UndertaleModTool.Windows.UndertaleResourceReferenceMethodsMap.<>c.<.cctor>b__9_42(Object objSrc, HashSetTypesOverride types, Boolean checkOne) in D:\a\UndertaleModTool\UndertaleModTool\UndertaleModTool\Windows\FindReferencesTypesDialog\UndertaleResourceReferenceMethodsMap.cs:line 455
   at UndertaleModTool.Windows.UndertaleResourceReferenceMethodsMap.GetReferencesOfObject(Object obj, UndertaleData data, HashSetTypesOverride types, Boolean checkOne) in D:\a\UndertaleModTool\UndertaleModTool\UndertaleModTool\Windows\FindReferencesTypesDialog\UndertaleResourceReferenceMethodsMap.cs:line 1418
[...]

PR #2467, opened 2026-08-23
Fixes and improvements for "Find all references" feature

### LINKED ISSUE #2475: The color strings from the layer effect properties are ignored by "Find all references" if it's LTS 2022
### Describe the bug

Judging by the source code of `UndertaleRoom.Layer`, the effect properties are added in GM 2022.**1**.
But [the recent commit](https://github.com/UnderminersTeam/UndertaleModTool/commit/6f5db1947e288a7c59603388371d59e17e71c3aa#diff-8060900fe8a156a51375f50cd1b2f9bf700dd48c111254c9fddb3d8faa40a107) makes the feature treat the LTS 2022 game version as 2022.**0**, so this results in unreferenced color strings (e.g. `#FFFFFFFF`), even if they are actually referenced by `UndertaleRoom.EffectProperty`.

### Reproducing steps

1. Open UTMT.
2. Load the latest "Deltarune" (a GM LTS 2022 game) version, Chapter 5.
3. Try to find all references of the "#FFFFFFFF" string.

### Setup Details

1. UTMT version - 0.9.2.0

### LINKED ISSUE #2477: The "Find unreferenced assets" feature is broken for YYC games
### Describe the bug

When you try to use the feature, it shows the `NullReferenceException`, because it tries to access `data.Code`, even if it's a YYC game.

### Reproducing steps

1. Load a YYC game (or simply make the `Data.Code` `null`).
2. Open "Find unreferenced assets".
3. Select any type.
4. Click "Search".

### Setup Details

1. UTMT version - latest (0.9.2.0)

### LINKED ISSUE #2478: The asset references ("Find all references" & "Find unreferenced assets") are not quite ordered

UI/UX

Currently, when you search for asset references ("Find all references"), the results are not ordered - the lists themselves seem to have a random order.

The same goes for "Find unreferenced assets" - since it's using multithreading, both the asset group and the each asset list order is not consistent.
The assets should be sorted by asset ID, so in the same order as in the source asset lists, and the groups should be probably sorted by group name.

VladiStep:
## Description
1. It turns out that I missed that the audio groups are introduced in GameMaker versions with bytecode 14+.
So this type showed for such versions in the referenceable asset list, and this caused the null reference exception.
This PR fixes that.
2. Added a `GameVersion` constructor that accepts `UndertaleGeneralInfo` (and it also has the LTS 2022 logic).
3. Improved performance of "Find unreferenced assets" by adding early `return`s/`break`s for tracks (`void ProcessTrack() ...`), so it stops as soon as it finds a one reference inside a track.
4. Changed the `GameVersion` type to `readonly record struct` for better performance.
5. Simplified the bytecode version checking (comparison) in `GameVersion` by adding a proper support for that (instead of doing `(bytecodeVersion, uint.MaxValue, uint.MaxValue)`).
6. Added `PredicateDelegate` for a better readability of `PredicateForVersion`.
7. Now `UndertaleBackground` uses `BeforeVersion` (as it should) instead of using `null` names.
8. Made the version comparison more readable by changing the operands order.
9. Simplified asset type filtering for YYC.
10. Removed now redundant `PredicateForVersion.DisableForLTS2022` property and check.
11. Minor code changes (made early `continue` in a loop).
12. Treat LTS 2022 GM version as 2022.9, as it's actually like that internally; fixes #2475
13. Fixed a (theoretically possible) bug when `GetReferenceableTypes()` can return stale results for games with same GM version, but different YYC types (first loaded game data has code, second - doesn't).
14. Fixed an error for YYC games when you try to find unreferenced assets - it tried to access `data.Code` (fixes #2477).
15. Sort the results by group names and each asset list by asset ID - closes #2478.

### Caveats
None, I guess.

### Notes
Closes (partially) #2319 (the error part)

### DIFF
--- a/UndertaleModTool/Windows/FindReferencesTypesDialog/FindReferencesTypesDialog.xaml.cs
+++ b/UndertaleModTool/Windows/FindReferencesTypesDialog/FindReferencesTypesDialog.xaml.cs
@@ -86,18 +86,12 @@ namespace UndertaleModTool.Windows
                 return;
             }
 
-            GameVersion ver;
-            if (data.GeneralInfo.Branch == UndertaleGeneralInfo.BranchType.LTS2022_0)
-                ver = (2022, 0, 0);
-            else
-                ver = (data.GeneralInfo.Major, data.GeneralInfo.Minor, data.GeneralInfo.Release);
-            var sourceTypes = UndertaleResourceReferenceMap.GetReferenceableTypes(ver);
+            GameVersion version = new(data.GeneralInfo);
+            bool isYYC = data.Code is null; // There is `UndertaleData.IsYYC()`, but it also checks for general info
+            var sourceTypes = UndertaleResourceReferenceMap.GetReferenceableTypes(version, isYYC);
 
             foreach (var typePair in sourceTypes)
             {
-                if (data.Code is null && UndertaleResourceReferenceMap.CodeTypes.Contains(typePair.Key))
-                    continue;
-
                 TypesList.Items.Add(new CheckBox()
                 {
                     DataContext = typePair.Key,
--- a/UndertaleModTool/Windows/FindReferencesTypesDialog/UndertaleResourceReferenceMap.cs
+++ b/UndertaleModTool/Windows/FindReferencesTypesDialog/UndertaleResourceReferenceMap.cs
@@ -7,41 +7,55 @@ using UndertaleModLib.Models;
 
 namespace UndertaleModTool.Windows
 {
-    public record GameVersion(uint Major, uint Minor, uint Release) : IComparable<GameVersion>
+    public readonly record struct GameVersion((uint Major, uint Minor, uint Release)? GameMakerVersion = null,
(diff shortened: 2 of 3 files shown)


VladiStep:
Okay, it seems the `HashSetTypesOverride` custom type needs to be modified so that it returns `true` only if the requested type is supported in the current GameMaker version

VladiStep:
I also need to rewrite `GameVersion` so it properly supports the bytecode version.
At least because `GetReferenceableTypes()` will ignore audio groups in the current implementation

VladiStep:
It looks finished, but I need to test it properly - at least make sure, that the performance is the same

VladiStep:
Well, at least it has the same performance, I didn't notice visible speed and/or memory allocation changes for "Find unreferenced assets" for Deltarune Chapter 5 - it's ~1 m. 20 s. for my PC.

VladiStep:
Okay, I think it's ready for review.

